PR #3, opened 2016-03-15
Add dropdown menu for Categories

gfjardim:
Just fixed the black theme dropdown CSS.

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -940,7 +940,7 @@ $showAdditionalInfo = '';
     var index = $(el)[0].selectedIndex;
     if (index == 0) {
       // Path
-      mode.html("<dt>Mode</dt><dd><select name='Mode'><option value='rw'>Read/Write</option><option value='ro'>Read Only</option></select></dd>");
+      mode.html("<dt>Mode</dt><dd><select name='Mode'><option value='rw'>Read/Write</option><option value='rw,shared'>RW/Shared</option></option><option value='ro'>Read Only</option></select></dd>");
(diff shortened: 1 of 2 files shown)


limetech:
Curious: What does RW/Shared (rw,shared) actually do, and why is this useful?

gfjardim:
It got lost in the middle of this PR. It's only relevant to [this](https://github.com/limetech/dynamix-6.2/issues/4) issue.

limetech:
What I'm asking is, how is it relevant?  That is, if we do the bind-mount trick on /mnt and we invoke docker with "--propagation slave", isn't that enough now for containers to see "new" mounts that occur under /mnt after the docker engine is started?  Now, what purpose does 'rw,shared' serve?  Guess I'm confused.... as usual :)

gfjardim:
Ah, now I understand.

When docker starts, it will clone the parent namespace (let's call PN the parent namespace and CN the docker child namespace ).

Now, at first, PN and CN are equal, but when docker starts to mount volumes in CN, those targets become invisible to PN processes. This occur because, by default, all mounts are `private`. The same occur when we mount targets on PN, they are invisible to CN process.

Now, when PN makes e.g. **/mnt** shared, this directory can be shared with CN processes, but, when we map a volume in docker (CN), it will be, by default, mounted as `private`. So we have to let the daemon know that we want that specific volume to be mounted as shared or slave.

I think it should be set to `rw,slave`, because this prevents undesirable effects of **upward propagation**; in thesis, if `shared` a container could unmount a PN mount target.

gfjardim:
I've tested it, and a `shared mount` could leave the host vulnerable to a container issued unmount, if the container is privileged. With container's volume mounted as `slave`, the unmount only took place inside the container's namespace, and the host mount was unaffected.

PR #7, opened 2016-03-20
Introduce context -sensitive help

bergware:
This allows for help of individual settings by clicking on the specific
setting context.


bergware:
In general this works fine when pages are constructed using *markdown* (most settings pages).

Notorious exceptions are Docker and VM.

In case of Docker the help text doesn't obey the layout, resulting in too long text (this needs to be corrected anyway).

In case of VM, not all settings get context-sensitive help, due to differences in the item structures (would be nice to have them all equal).

PR #11, opened 2016-03-27
Fix single line display of non-hyperthreaded cores on Dashboard page

bergware:


eschultz:
Looks good, thanks @bergware

PR #14, opened 2016-03-30
No not override /config name or visibility.

gfjardim:


plugins/dynamix.docker.manager/include/CreateDocker.php
```diff
@@ -593,8 +593,12 @@ function setXmlVal(&$xml, $value, $el, $attr = null, $pos = 0) {
         foreach ($xml['Config'] as &$arrConfig) {
           if ($arrConfig['Type'] == 'Path' && strtolower($arrConfig['Target']) == '/config') {
             $arrConfig['Default'] = $arrConfig['Value'] = realpath($dockercfg["DOCKER_APP_CONFIG_PATH"]).'/'.$xml["Name"];
-            $arrConfig['Display'] = 'hidden';
-            $arrConfig['Name'] = 'AppData Config Path';
+            if (empty($arrConfig['Display'])) {
+              $arrConfig['Display'] = 'advanced';
+            }
+            if (empty($arrConfig['Name'])) {
```
eschultz: For V1 templates, `$arrConfig['Name']` will always be something like 'Path 1'.  If we check for empty-ness here the Name will never be set to 'AppData Config Path'.

Same for Display, it'll always be 'always'.

I guess the purpose of this was to hide the /config path so it's one less thing for the user to fill out since unRAID should be installing (and managing) it off of their appdata share.

gfjardim:
So what do you think, we use advanced or advanced-hide? **hidden** is deprecated at this point?

eschultz:
Maybe something like this:
```php
if (preg_match("/^Path\s\d/", $arrConfig['Name'])) {
  // V1 /config detected, clear these out so the defaults apply
  $arrConfig['Display'] = '';
  $arrConfig['Name'] = '';
}
if (empty($arrConfig['Display'])) {
  $arrConfig['Display'] = 'advanced-hide';
}
if (empty($arrConfig['Display']) {
  $arrConfig['Name'] = 'AppData Config Path';
}
```

Basically for V1 /config paths we set some nice defaults.  If it's a V2, the author has full control over the label (Name) and Display... but we have defaults for those if they don't specify either.

eschultz:
Closing this pull request, I ended up doing something like this:
```php
if (empty($arrConfig['Display']) || preg_match("/^Path\s\d/", $arrConfig['Name'])) {
  $arrConfig['Display'] = 'advanced-hide';
}
if (empty($arrConfig['Name']) || preg_match("/^Path\s\d/", $arrConfig['Name'])) {
  $arrConfig['Name'] = 'AppData Config Path';
}
```
Thanks for your help!

gfjardim:
Sorry for the delay. That's seems nice. 

Thanks Eric.

PR #17, opened 2016-04-28
Updated diagnostics utility

bergware:


eschultz:
Lots of nice improvements to the diagnostics script 👍

PR #19, opened 2016-05-09
Enhanced network settings

bergware:


eschultz:
Was there a issue with Font Awesome version 4.6.3 that you had to revert?

bergware:
When I copied the new version, I didn't update the references in  the css file. This caused the font to 'disappear'. At first I didn't know the source of the issue and reverted the PR, thinking it had to do with the new version. Meanwhile I updated the css file and all seems to work okay. Will make another PR.

bergware:
List is getting quite long, do you consider a commit?

PR #24, opened 2016-06-05
Auto-hide unused DNS server fields

bergware:


bergware:
Got confused because your merge and my PR crossed each other...

PR above is one change to auto-hide DNS server fields which are not filled-in (unused)

PR #42, opened 2016-06-20
Add Description field back to Create Docker.

gfjardim:


plugins/dynamix.docker.manager/include/CreateDocker.php
```diff
@@ -618,7 +618,7 @@ function getUsedPorts() {
             if (empty($arrConfig['Display']) || preg_match("/^Path\s\d/", $arrConfig['Name'])) {
```
eschultz: should `Path` be `Host Path` here in the preg_match pattern too?

gfjardim: Yep, because of the mods I made on line 309. Ronald added "Host" prefixing all 1.0 converted fields in JS, so I ported them to PHP.

eschultz: Gotcha, so line 618 should really be:
```php
if (empty($arrConfig['Display']) || preg_match("/^Host Path\s\d/", $arrConfig['Name'])) {
```

gfjardim: Correct.

PR #48, opened 2016-06-29
Added 'preset current configuration' with New Config utility [13]

bergware:


bergware:
Changed "preset" to "retain"

PR #51, opened 2016-07-03
Allow multiple disk selection in NewPerms utility

bergware:


plugins/dynamix/scripts/newperms
```diff
@@ -17,34 +17,34 @@
...
+  IFS=$ifs
 }
 
-if [ -n "$1" ] ; then
+if [[ $# -ge 1 ]]; then
   process "$@"
 else
   process /mnt/cache
-  for disk in /mnt/disk* ; do
+  for disk in /mnt/disk[0-9]*; do
```
eschultz: Look like you also fixed a bug preventing `/mnt/disks` (folder used by Unassigned Devices plugin) from being affected by new permissions.  Nice work!

bergware: Yeah, did a similar trick in the file integrity plugin.

limetech:
Ok, I merged this PR, but in testing - what is the point in specifying individual 'disk's?  And instead of the drop down listing every slot, maybe only list /mnt/disk[0-9]* (ie, those disks that are actually mounted).

Actually I think we should re-address this post 6.2 stable release because probably we want it operate on a share basis (if user shares enabled) or device basis (if user shares disabled).

bergware:
Two reasons for the selection:

1. Doing all disks including cache may break permissions for docker/VM applications stored on cache
2. Convenience of having the possibility to update selective disks as needed (not just for v4/v5-6 upgrade).

Are you seeing "open" disks?

PR #57, opened 2016-08-01
Fixed: network counters and errors on dashboard, when VLANs are used

bergware:


plugins/dynamix/include/DashUpdate.php
```diff
@@ -185,8 +185,8 @@ function active_disks($disk) {
       }
     }
   break;
-  case 'port': exec("ifconfig -a -s|awk '/^(bond|eth|lo\s)/{print $3\"#\"$7}'",$ports); break;
-  case 'link': exec("ifconfig -a -s|awk '/^(bond|eth|lo\s)/{print \"Errors: \"$4\"<br>Drops: \"$5\"<br>Overruns: \"$6\"#Errors: \"$8\"<br>Drops: \"$9\"<br>Overruns: \"$10}'",$ports); break;
+  case 'port': exec("ifconfig -a -s|awk '/^(bond|eth|lo)[0-9]*\s/{print $3\"#\"$7}'",$ports); break;
```
eschultz: The updated regex in this commit will break matching the _lo_ interface:
`(bond|eth|lo)[0-9]*\s`

Possible solution might be something like this?:
`((bond|eth)[0-9]*|lo\s)`

(The `\s` was meant to be just appended to _lo_)

bergware: It won't break 'lo', since `[0-9]*` means with or without digit

The `\s` is now used to eliminate sub-interfaces, eg, `eth0.1`

bergware: `# ifconfig -a -s | awk '/^(bond|eth|lo\s)/{print $1}'`
`bond0`
`eth0`
`eth1`
`lo`

bergware: `ifconfig -a -s | awk '/^(bond|eth|lo)[0-9]*\s/{print $1}'`
`bond0`
`eth0`
`eth1`
`lo`

eschultz: Oh, right, sorry about that.  What you have in the commit will work good, as is.  Apparently I need to have another cup of coffee ☕

bergware: No probs, it is still very early at your side of the world, enjoy your coffee :)

PR #59, opened 2016-08-08
Enhanced date and time settings

bergware:


bergware:
The enhanced time zone settings have many more choices than currently available in unRAID,
e.g. I like to set Europe/Amsterdam but only Europe/Berlin is available.

Can the TZ database be expanded or do you want to limit the available choices?

limetech:
The current set of zoneinfo included in unRAID OS requires 374K.  The entire set of zoneinfo from glibc is over 6MB.  Since root file system is in RAM I don't see any need to waste 5.7MB on unused data files.  Therefore we only include the most common set of timezone identifiers (which I think we got from what windows7 uses).  If we want to include a couple more for convenience sake, then no problem, but I don't want to include the whole set.

bergware:
I was just counting.
The current implementation has 90 locations. The new list has 410 locations, that is about 4.5 times current. A simple calculation 374K * 4.5 = 1.3M. I guess, too much to your liking?

limetech:
But why?  If there are subtle differences between say Amsterdam and Berlin, then fine, but if all the timezone stuff is identical, what is the purpose?

bergware:
There isn't a real difference cause most countries in Western Europe are in the same time zone (CEST), it is more a sentimental thing, as I am Dutch and not German...

I made some test with the map and the reduced list as currently present. This results in many "white" spots, in other words if you want to keep the current list the idea of using a map doesn't work.

Did you try to use the map and like the approach? If not suitable I'll revert to the old approach.

bergware:
The latest update uses the existing list of time zones (no need to make changes).
I grouped zones together for selection in the map.
Instead of a "pin" the complete time zone is now highlighted, looks nifty :)

bergware:
Removed some real remote islands from the map :)

bergware:
You guys maybe wondering... so many updates ;)

The original list which I took from www.timezonepicker.com has several errors. I have been going thru the list a dozen of times to make sure the correct time zones are indicated everywhere (as of July 2016). And updated several shapes to get correct highlighting.

For the dropdown list I followed the latest list of Windows 10, while verifying all entries  run across one *missing* entry: **Sakhalin**. This is because Russia decided to split Kamchatka into two time zones. Herewith my nimble request to add **Asia/Sakhalin  -  (UTC+11:00) Sakhalin**

Sorry if any inconvenience!

bergware:
I am afraid two more additions to the list. North-Korea changed their local time to 30 mimutes earlier and there is a tiny area in Australia using a "special" time.

**Asia/Pyongyang|(UTC+08:30) Pyongyang**
**Australia/Eucla|(UTC+08:45) Eucla**

eschultz:
team_america_a_l

limetech:
Holy crap that explains all the email from NK! lol

bergware:
LOL. I accidentally changed my PC to NK time... but everything is alright :)

image

bergware:
Btw, was considering this map...

image

eschultz:
Those 3 extra timezones will increase the space needed by 2KB ⚠️ .  That's pushing it but I'll let it slide this time and include them, LOL!

bergware:
The Russians, Aussies AND North Koreans thank you immensely 👍

bergware:
Seems like you are not super enthusiastic about the world map feature.
If you want I can take it out and just leave the updated/corrected time zones in the update.
I can always introduce the world map feature as a separate plugin.

PR #69, opened 2016-09-25
Backup .plg files on updates

Squidly271:
To support rollbacks if required

### DIFF
--- a/plugins/dynamix.plugin.manager/scripts/plugin
+++ b/plugins/dynamix.plugin.manager/scripts/plugin
@@ -569,6 +569,9 @@ if ($method == "update") {
     exit(1);
   }
   // install was successful, save the updated plugin so it installs again next boot
+  $oldVer = plugin("version","/var/log/plugins/$plugin",$error);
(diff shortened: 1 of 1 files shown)


bergware:
The ability of doing a plugin rollback depends a lot on the plugin itself.

For example all Dynamix plugins maintain a single version. AKA the latest version. There is no way to downgrade once the new version is installed. This is done on purpose for several reasons.

Furthermore some of the plugin updates in the past were irreversible, e.g. convert to newer database format or relocation of files. These kind of actions make a rollback impossible.

In short a rollback possibility sounds interesting, but it will be limited in use.

bergware:
Another strategy could be to let the plugin itself indicate whether a rollback is available (and which one).

This can be achieved by introducing a new attribute (e.g. **rollback="2016.09.22"**, but requires current plugins to be updated to support that.

At the same time wondering if all the effort is worthwhile...

limetech:
Interesting idea but too much work to do it right.  Better to fix issues so that rollback is unnecessary.

PR #71, opened 2016-10-04
Suppress error if temporary file deletion failed in parity settings

bergware:


bergware:
I think I have the answer for an unrestricted (but clean) system reboot/shutdown.

The buttons reboot and shutdown are permanently available and when click they issue the */sbin/command* which in turn invokes rc.local_shutdown and an attempt to stop the running array before reboot/poweroff.

Have a look.

PR #87, opened 2016-12-17
Add BTRFS file system checking and notifications

bergware:


bergware:
[1]
See this feature request: http://lime-technology.com/forum/index.php?topic=54764.0
Notifications added in *monitor* script

[2]
See this discussion: http://lime-technology.com/forum/index.php?topic=54699.0
User can change filesystem settings of a replaced disk, before a data-rebuild is started, which gives unexpected results

limetech:
[1] Yes that's nice.

[2] Ok first time in 10 years this has come up but I see your point.  You have to also check for mdState SWAP_DSBL, something like this:

`disabled = $var['mdState']=="RECON_DISK" || $var['mdState']=="SWAP_DSBL" ? "disabled" : "";`

bergware:
Updated. 10 years ago, there was only ReiserFS in unRAID :)

limetech:
> Updated. 10 years ago, there was only ReiserFS in unRAID :)

True dat!

limetech:
Hi Ronald, Hey please withdraw the "btrfs raid mode as dropdown selection" since there are two issues.  First, there is an internal emhttp operation that will force rebalance back to raid1 when a new device is added to the pool (this has to be fixed).  Second, we do not wish to expose raid5/6 at this time because a) its not necessary and b) we haven't sufficiently tested ourselves to ensure it's worth the support we'd have to provide if a problem comes up.

bergware:
Okay, reverted. Wasn't aware of the implications.

PR #91, opened 2017-02-03
SweetAlert bugfixes

gfjardim:


eschultz:
@gfjardim I haven't tested this code yet but a quick look over the diff reveals a few unicode and newline characters were restored as well?

gfjardim:
Those were introduced by GitHub, I made that commit directly from this site. I'll commit using git instead.

eschultz:
ok, I see, the main fix is nesting the OK button instead of the Cancel button under the `sa-confirm-button-container` div.

gfjardim:
Yes. The other change is the color of animation, green instead of blue.

PR #94, opened 2017-02-04
Fix version checking

Squidly271:
As it stands, any upgrades from a RC to a final will fail


bergware:
Hmm, this will break the version checking of plugins which follow the yyyy.mm.dd notation, in case an additional letter is added, e.g. 20170204a and 20170204b are seen as identical by the version_compare function.

Squidly271:
I don't believe so.  And I tried it by changing a plugin's version that had a suffix with /var/plugins and it found and installed the update no problems

Squidly271:
kk  You are correct...  version_compare does recognize a difference between 2017.02.05 and 2017.02.05a, but its returning the opposite value than you would expect

PR #95, opened 2017-02-17
Better csrf_token handling

gfjardim:
This is a drop-in replacement for the ajaxSend() function. It will inject the csrf_token in every POST request done using ajax, and, if a status 403 is received from the server, it will retrieve the current csrf_token from the server and retry the request once.

This will require emhttp to trow a 403 error when csrf_token is wrong or missing.


limetech:
Interesting idea.  Another idea is to have emhttp send a redirect to the browser upon "wrong csrf_token", causing browser to reload page.  But things will be changing soon as we replace emhttp with nginx.  btw: don't need test for existence of 'var.ini', it's guaranteed to exist.

PR #98, opened 2017-03-24
Disk read/write IO in background daemon

bergware:
The current implementation lets the GUI do retrieval of the disk read/write IO activity, i.e. reading and processing of /proc/diskstats.

I have changed this function to a background daemon, similar to what is done for CPU load monitoring. This has the benefit that the disk IO retrieval runs independent of the GUI, simplifies the GUI processing and is more accurate.


bergware:
Scripts have been moved under /usr/local/emhttp/webGui/scripts and the diskload daemon is started and stopped together with the array.

limetech:
Nice work man!

PR #113, opened 2017-06-03
Theme optimization and alignment

bergware:


limetech:
we don't need no stinkin' semicolons!

bergware:
one thousand changes and counting :)

PR #118, opened 2017-06-09
Themes, dashboard and docker refinements

bergware:


limetech:
Thanks for the refinements!  Hey what happened to the "Page update frequency" and associated "disable page updates while parity is running" Display Setting?

bergware:
The "page update frequency" was a remnant of the old v5 way of working. The screen updating interfered with disk activity due to emhttp reading the SMART parameters with each HTTP request.
In v6 this is not the case anymore (separate activity) and with nginx handling HTTP(S) requests there is no need to disable screen activity while a parity operation is running. Also because we are moving to a websocket approach to goal of real-time updates will have low(er) impact.

PR #123, opened 2017-06-18
Enhanced plugin manager with separate plugin and OS sections

bergware:


bergware:
My first "assignment" :)
This enhances the plugin manager with the following:
- Original plugins page will only show user (custom) plugins
- Upon entering the plugins page an auto-check is performed (no more 'Check for Updates')
- A new page "Update OS" is available under Tools -> About and shows system updates (unRAIDServer and dynamix)
- Also new page has auto-check functionality, so status is always up-to-date
- It is possible to switch between Stable and Next releases
- System updates are made VERY visible to the user (no way to overlook them)
- Addition to plugin script to allow "forced" installation. This is needed when switching between next and stable releases
- No longer separate messages about dynamix plugin updates
- That's about it ...

bergware:
Fun fact: while working on the plugin manager, suddenly it was telling me a new version 'rc4' was available. Proof the auto-check works nice :)

limetech:
One thing that concerns me with "auto" checking for updates is the effect this might have on servers which have no connectivity to the internet.  Will this auto-check cause the page to 'hang' for a long time, waiting for a connection which will never complete?  If not, no worries.

bergware:
There will be a longer waiting time when the github server isn't responding.
A unRAID server without internet will get almost instant no-reply, actually faster than the regular response.
When no communication could be established a "Retry" button is shown and all plugins are set to "no update".

PR #132, opened 2017-06-23
Added GUI access mode

bergware:


limetech:
Something wrong, clicking on GUI Access causes content to fill with

Warning: array_shift() expects parameter 1 to be array, string given in /usr/local/emhttp/plugins/dynamix/include/DefaultPageLayout.php(371) : eval()'d code on line 15

bergware:
Man, don't you sleep???
What is the content of your go file?

PR #133, opened 2017-06-23
Fix go file parsing

bergware:


bergware:
It doesn't work perfect....

Can you have a look too. Once the "restart_gui" script is executed there is no more communication between nginx and emhttpd. Is it required to rebuild the linux socket?

bergware:
okay, we put this in park mode.

PR #145, opened 2017-07-15
Allow referrals from CA to dockerMan to return to CA when pressing Done

Squidly271:
Currently, if a CA user presses Done after installing / editing a container, upon pressing Done they will be taken to a blank Add Template Page.
It is far neater to instead have them taken back to the Apps tab instead.

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -1726,5 +1726,10 @@ $showAdditionalInfo = '';
     }?>
   });
 </script>
-<?END:?>
+<?END:
+if ($_GET['caRefer']):
+?>
+<div id='caRefer' style='display:none'></div>
+<?endif;?>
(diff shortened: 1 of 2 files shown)


limetech:
You tested this against 6.4.0-rc release?

Squidly271:
Yes (only running 6.4).  If a change is made is made on the template, a "Reset" will simply reload the page, but if the button is "Done" then CA is reloaded.  Also takes an update to CA to accomplish, but I've got another one ready to go anyways..

eschultz:
@Squidly271 Even easier, without needing any code changes here, would be to change the urls you're using to add or edit containers:

Before: /AddContainer

After: /Apps/AddContainer

This way clicking 'Done' will "go up one folder" in a sense and navigate to /Apps

Squidly271:
Never thought of that tidbit.  The "Issue" has been there since the original community repositories 3 years ago.  grr.

PR #152, opened 2017-07-29
Update plugins without page refresh

ljm42:
After updating a plugin, display latest plugin information inline, rather than requiring a full page refresh.


ljm42:
I've got some ideas to improve error handling, will update this pull request

ljm42:
Are you able to squash and merge this, or would I need to close the pull request and create another one?

bergware:
ljm42, your idea is nice and to take more advantage of this approach it should be extended to other plugin operations as well, not just a plugin update. This would benefit the overall look-and-feel of the plugin manager.

If you don't mind lets skip this PS but we use your approach in a more full fletched version in a future release.

ljm42:
Thanks for considering it. Updating plugins is the most common maintenance task I perform on unRAID, so speeding up the process is very important to me.

But if you see ways to take a more holistic approach, that is probably better in the long run.

PR #176, opened 2017-09-21
Addition of Error Correction Type information

realies:
It would be great to start integrating ECC/EDAC support in unRAID, one of the first steps might be to display relevant information from the BIOS/UEFI.

Reducing the `dmidecode -t 16` calls to one is preferable, I am not aware if `Maximum Capacity` and `Error Correction Type` can be extracted with `awk` in one go without having to `explode` the row as done with `$cache`, otherwise it would be fine.

[
screen shot 2017-09-22 at 01 32 00
](url)

### DIFF
--- a/plugins/dynamix/include/SystemInformation.php
+++ b/plugins/dynamix/include/SystemInformation.php
@@ -142,8 +142,8 @@ $memory_installed = exec("dmidecode -t 17 | awk -F: '/^\tSize: [0-9]+ MB$/{t+=$2
 // Extract error correction type, if none, do not include additional information in the output
 list($memory_maximum, $ecc_support) = array_map("trim", explode("#", exec("dmidecode -t16|awk -F: '/^\tMaximum Capacity: [0-9]+ GB$/{t+=$2};/^\tError Correction Type:/{e=$2} END{print t\"#\"e}'")));
 // If maximum < installed then roundup maximum to the next power of 2 size of installed. E.g. 6 -> 8 or 12 -> 16
-if ($memory_maximum < $memory_installed) {$memory_maximum = pow(2,ceil(log($memory_installed)/log(2))); $star = "*";}
 $star = "";
+if ($memory_maximum < $memory_installed) {$memory_maximum = pow(2,ceil(log($memory_installed)/log(2))); $star = "*";}
 echo "$memory_installed GB ".($ecc_support == "None" ? "" : "$ecc_support ")."(max. installable capacity $memory_maximum GB)$star";
 ?>
 </div>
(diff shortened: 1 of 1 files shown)


eschultz:
Just a update: I'm enabling EDAC support in the linux kernel

limetech:
thx.  What about the thing chbmb wants added?  Requested via PM, I think you were included.  Any impact for us?

 

Cheers,

Tom

realies:
@eschultz, good to know. Is this expected to be available in the next RC? Looking forward to testing this on an ECC enabled X370 platform.

eschultz:
@realies We jumped up to kernel 4.13.4 and added EDAC so if the new kernel checks out (i'm sure it will) then it'll all be included in the next RC.  Kernel 4.12.x is EOL now.

bergware:
You can combine both searches in a single awk statement. Like this

`list($memory_maximum,$ecc_support) = array_map('trim',explode('#',exec("dmidecode -t16|awk -F: '/^\tMaximum Capacity: [0-9]+ GB$/{t+=$2};/^\tError Correction Type:/{e=$2} END{print t\"#\"e}'")));`

And make a single echo statement. Like this

`echo "$memory_installed GB ".($ecc_support=='None'?"":"$ecc_support ")."(max. installable capacity $memory_maximum GB)$star";`

realies:
Since the output of `dmidecode` is unlikely to change without hardware or BIOS setting changes, it might be optimal to cache `dmidecode` to a processed file within the ramfs and consult it instead.

plugins/dynamix/include/SystemInformation.php
```diff
@@ -139,11 +139,12 @@ function write($number) {
...
 // might have two or more of these.  The trick is to filter out any Flash/Bios types by matching on GB
 // Sum up all the Physical Memory Arrays to get the motherboard's total memory capacity
-$memory_maximum = exec("dmidecode -t 16 | awk -F: '/^\tMaximum Capacity: [0-9]+ GB$/{t+=$2} END{print t}'");
-$star = "";
+// Extract error correction type, if none, do not include additional information in the output
+list($memory_maximum, $ecc_support) = array_map("trim", explode("#", exec("dmidecode -t16|awk -F: '/^\tMaximum Capacity: [0-9]+ GB$/{t+=$2};/^\tError Correction Type:/{e=$2} END{print t\"#\"e}'")));
 // If maximum < installed then roundup maximum to the next power of 2 size of installed. E.g. 6 -> 8 or 12 -> 16
 if ($memory_maximum < $memory_installed) {$memory_maximum = pow(2,ceil(log($memory_installed)/log(2))); $star = "*";}
-echo "$memory_installed GB (max. installable capacity $memory_maximum GB)$star";
+$star = "";
```
ljm42: Are you sure we should clear the value of $star right after setting it?

PR #186, opened 2017-10-10
Support tooltipster CSS on bottom / left / right

Squidly271:


Squidly271:
Easiest way to test is to use CA with a very small window height to force tooltipster to left & right and also hover over CA Manual / Statistics

PR #187, opened 2017-10-11
Array status statistics panel space optimisation

realies:
Using the available screen estate is always better than leaving unused space. Pre-filling the layout with disk columns that are not in use is unnecessary. This patch expands the columns of available system disks to the full width of the panel and is a step forward towards a responsive layout that scales with screen size.

Future iterations of this panel might benefit from rendering data of individual disks per row instead of per column, exponentially growing vertically instead of horizontally.

unraid array status patch

### DIFF
--- a/plugins/dynamix/DashStats.page
+++ b/plugins/dynamix/DashStats.page
@@ -115,17 +115,17 @@ $slots = count($disks) + count($devs);
 ?>
 <?if ($slots<30):?>
 <style>
-table.share_status.fixed td:nth-child(<?=$slots+1?>)~td{visibility:hidden}
+table.share_status.fixed td:nth-child(<?=$slots+1?>)~td{display:none}
 </style>
 <?endif;?>
 <table class='share_status fixed'>
 <thead><tr>
 <?
 $row0 = array_fill(0,30,"<td>-</td>"); $i = 0;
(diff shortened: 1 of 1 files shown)


bergware:
No, it is designed to keep columns narrow and not stretch over the full width of the screen. People with high resolution monitors have very (too) wide columns when a few disks are present.

With a completely full system (30 data + 24 cache) there are two tables created. The current design allows for up to 60 devices in total (if people have unassigned devices as well).

realies:
@bergware, thoughts on the vertical rather than horizontal listing approach? This would maximise layout space usage which will scale with drives without wasting screen estate and having to insert another table.

bergware:
The purpose of the current setup is to see instantly what state each disk is in (e.g. it is very easy to recognize which disks are active and inactive).

How would vertical listing improve real estate? 8 columns stretched over a full screen width of 1920 pixels is going to give a lot of user disapprovals, and likely those with many disks in the array will complain about the scrolling.

realies:
@bergware, 8 (or up to 30) columns over a full screen is a vertical disk representation. Having a horizontal disk representation as done in the Main page scales up the space used up proportionally to the amount of disks and does not leave potential chunks of empty space as it could currently do.

PR #197, opened 2017-11-03
Fix regression error: check device with smartctl

bergware:


bergware:
Regression after last PR with monitor code updates.

PR #209, opened 2017-11-18
Revised SMART temperature reading logic

bergware:
194 = raw value is first choice
190 = raw value is second choice


bergware:
Using the following awk statement:

`awk 'BEGIN{s=t="*"}$1==190{s=$10};$1==194{t=$10;exit};$1=="Temperature:"{t=$2;exit} END{if(t!="*")print t; else print s}'`

This will let attribute 194 get preference if existing, attribute 190 is read as 'raw' value only when 194 is not present.

PR #214, opened 2017-11-26
Fixed allow "permit format" when input file is selected

bergware:
+ Logical rearrangement of code sections


limetech:
Just a curiosity, in this span:

<span id='pass'><input name='luksReformat' type='checkbox' onchange='selectInput(this.form)'>permit reformat</span>

Why did you change id from 'perm' to 'pass'?  I assumed 'perm' meant "the permit reformat element", what does 'pass' mean?

bergware:
I just didn't like the abbreviation "perm" and took a replacement word. Other then that no difference.

PR #221, opened 2017-12-11
Diagnostics updates

Squidly271:
Human readable free memory
Full ps output with children, sorted by CPU usage


Squidly271:
lol Just noticed that at some point top became included in diagnostics.  But, TBH I still prefer the whole path that my changes would include

limetech:
cool, thanks!

PR #229, opened 2017-12-21
Check for connectivity prior to updating docker updates

Squidly271:
At array start, if the network is invalid, it can take quite a while (20 minutes) for the cumulative timeouts on the template checks / updates before the array is fully functional.

Primarily affects those users running pfSense in a VM where no network access happens until the VM is started, and this can take up to 20 minutes in this situation.  Alternatively, if the user simply has their network cable unplugged, any VM's can take up to 20 minutes to start.

Many feature reqs / defect reports about this over the years.

### DIFF
--- a/plugins/dynamix.docker.manager/scripts/dockerupdate.php
+++ b/plugins/dynamix.docker.manager/scripts/dockerupdate.php
@@ -29,6 +29,10 @@ foreach ($argv as $arg) {
 }
 
 if (!isset($check)) {
+  system("ping -c 1 github.com > /dev/null 2>&1",$retValue);
+  if ( $retValue ) {
+    exit(0);
+  }
   echo " Updating templates... ";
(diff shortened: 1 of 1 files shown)


limetech:
Right, actually we just got rid of invoking dockerupdate.php from startup sequence entirely.

Squidly271:
Even better.  That was my original plan of attack but its not in the public code.

limetech:
For future ref: It's alright to send a forum PM or email, or even on public form to say, "Hey man, what you're doing is freakin' stupid, you need to do <this> instead.."  We don't see every post & usually (hopefully) respond pretty well to direct contact.

Squidly271:
lol.  I know.  Its just that the req / defect has existed and been posted ever since 6.0 beta began.  It was simply my "cause of the day"  Only 5 min of work invested into it.

PR #236, opened 2018-01-01
Allow menu buttons to have a destination

realies:


eschultz:
Won't this make the webterminal open and replace the whole web ui by default?  I know you can ctrl/command click, middle click or right-click and open in a new tab/window but it might not be obvious.

Perhaps the use case I'm envisioning for web terminal would be using it in conjunction with the web ui, like I would with the log window...

eschultz:
Actually, ignore my last comment.  The onclick will still fire to provide the openwindow on a normal left click and the return false will cancel the href.  The href helps for the purpose of ctrl/command click, middle click or right-click and open in a new tab.

Brilliant! 🍾

PR #238, opened 2018-01-04
Minor code optimizations

bergware:


limetech:
I don't want to add automatic killing of shells for array Stop.  If user is smart enough to be using shells they should be smart enough to know they must be closed, or 'cd' to some dir that's not on /mnt/*

Auto-kill of shells (and all processes) for shutdown is different - here we have to shutdown, so any unsaved data as a result of some process running in a shell we be lost.

ljm42:
Would it be possible to do a check for shells open in /mnt/* and not even try to stop the array until they are closed? This would avoid getting stuck at "retry unmounting user shares".

bergware:
It would open terminate a shell process which is open in /mnt/... 
I revert it and make it available as an optional plugin.

PR #239, opened 2018-01-07
Use tooltipster for all title attributes

Squidly271:
Doesn't work though on titles for Unassigned Devices as they are generated after the page is loaded.

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -563,7 +563,7 @@ $(function() {
   }
   $('form').append($('<input>').attr({type:'hidden', name:'csrf_token', value:'<?=$var['csrf_token']?>'}));
   watchdog.start();
-	$('[title][title!=""]').tooltipster();
+  $('[title][title!=""]').tooltipster();
 });
 </script>
 </body>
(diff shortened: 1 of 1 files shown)


limetech:
Where's an example of this being used?

Squidly271:
Hover text on the icons on the right.  (log, etc)  Especially when the windows is less that 1920 and only icons display.  Basically just gets rid of any of the built-in browser hover text and switches it to tooltipster.  Up to you.  Doesn't worry me at all.


Andrew Zawadzki
Technical Trainer
Arrow Games / Bazaar & Novelty
Office: 800.387.7621 ext 246
Cell: 289.213.0130
Email: azawadzki@arrowgames.com<mailto:azawadzki@arrowgames.com>



________________________________

bergware:
Personally I find tooltipster a bit "heavy" at least with the current (default?) styling. Is there a way to make it less prominent for the 'standard' tool tips?

Squidly271:
lol.  That's the last time I reply to a github notification from work....  Nothing like having personal info present.

Personally, I hate the standard tooltips.  Admittedly they aren't used too much in stock unRaid.  I'll close the PR.

You can adjust the sizing etc for tooltipster taking the place of title elements, but I'm not going to invest the time in this.

bergware:
Won't put it completely aside. If a "nicer" default style can be defined then tooltipster is still an option.
I'll keep it on my plate.

PR #242, opened 2018-01-19
Add installed plugins & version to diagnostics

Squidly271:


bergware:
I guess we can leave the '.plg' extention out in the list.

basename($plugin) ==> basename($plugin,'.plg')

Squidly271:
I originally had instead of the filename the result of plugin name $plugin, but decided that the two were close enough to each other.

bergware:
Haven't checked, but you're problably right. File name and plugin name are usually the same/similar.

limetech:
Need to merge into 6.4.1 - are we good to go?

Squidly271:
I think we are.  I reverted another commit that shouldn't have been included with this PR (unused icons)

PR #243, opened 2018-01-20
Updated jquery tablesorter to v2.29.4

bergware:
Note: stiil have issue that sometimes sorting gives wrong output


limetech:
> Note: stiil have issue that sometimes sorting gives wrong output

If that is the case, does this update fix anything else?

bergware:
It is a bug fix release. See https://github.com/Mottie/tablesorter/wiki/Changes
Perhaps I need to have a look at the code myself and figure out the issue, but hoping more it is done in the 'regular' process.

PR #250, opened 2018-01-27
Add in Support & Project Links to Docker Context Menus

Squidly271:
Every container installed via CA should already have the Support Link present (if installed post 6.2)
Majority of apps in CA also have a project link present (and passed to dockerMan), but thus far dockerMan hasn't saved that entry on user templates, so the Project link on the context menus will only appear for new containers added after the PR is approved.  (After published, I'll probably post a script to update everyone's user-templates with the appropriate Project URL's)

Additionally, dockerMan template authoring mode will also now support the most common CA xml entries (but nowhere near all) that template authors currently have to add manually to the generated xml file
untitled1


Squidly271:
Sorry about the commit, followed by a revert and a revert of the revert.  Still learning GitHub.

limetech:
This good to go?

Squidly271:
If there's a problem, I can't find it

Squidly271:
OK.  Found that IE doesn't like default parameters on functions.  Fixed

PR #256, opened 2018-01-31
More enhancements in Docker settings

bergware:
Simplification of networks display


bergware:
I removed the headers "system created" and "custom created" and consolidate the networks in a single list, which is easier to comprohend.

PR #259, opened 2018-02-05
 Enhancements to Docker custom network settings

bergware:


bergware:
The network enhancements include:
1. Fixed default gateway with different metrics. Now same gateway with multiple metrics properly set up
2. Fixed route_up and route_down functions. Now adding and removing routes handles IPv6 link local local addresses
3. Routing table shows interface name too with default routes
4. Network settings page is updated thru a websocket (/sub/dhcp), this ensures info is always present
5. Cleanup of network.cfg file. Only necessary settings are stored
6. Some fixes in the network settings page itself
7. IPv6 routes have their own metric value, can be set independently of IPv4
8. Some code cleanup

This PR goes together with updates of "rc.inet1", "rc.inet1.conf" and "create_ini_file" (in separate zip file).

PR #269, opened 2018-02-25
fix apparent typo

ljm42:


bergware:
There is also missing javascript code. I have made a PR which includes your correction + javascript.

PR #273, opened 2018-03-01
Updated tablesorter to v2.29.6 and fixed sorting issue in browser

bergware:


bergware:
The sorting issue seen earlier was caused by a misdetection of the sorting type. tablesorter looks at the first row of a table and determines then the sorting parser. If the first entry happens to be a name with only numbers (in my case the film title '13') it wrongly assumes it has to sort number based. The solution is to set a fixed sorting method for the column was holds the file names.

limetech:
thank you!  been meaning to check into this...

PR #277, opened 2018-03-06
Load docker and vm lists in background to speed up rendering

bergware:
Add rotating animation when starting or stopping a container or vm


bergware:
Should be alright. Same behavior but coded slightly different.

For example webGUI original (your change):

$webGuiUrl = $info['url'];
addslashes(html_entity_decode($webGuiUrl)),
New code
$webGui = html_entity_decode($info['url']);
addslashes($webGui)

Squidly271:
Damn github emails.  I noticed that you moved the command after I posted and deleted my comments

PR #284, opened 2018-03-11
Corrected memory display on Dashboard page

bergware:
follows now the same logic as system information


realies:
Perhaps abstract into a reusable piece of code?

limetech:
> Perhaps abstract into a reusable piece of code?

I'm all for that, how about you take a stab at it?

PR #285, opened 2018-03-11
Don't display select template drop down if referrer is Community Applications

Squidly271:


limetech:
You really want me to merge a PR that touch 371 files with a 'stable' release imminent?

Squidly271:
Woah.  Good catch.  I was trying to fix another issue I've got with GitHub and that wound up here.  I'll close and redo later tonight.

PR #287, opened 2018-03-11
Prevent text overlapping input boxes on Edit Docker Page

Squidly271:
See here: https://lime-technology.com/forums/topic/56392-support-djoss-dupeguru/?page=2&tab=comments#comment-639931


bergware:
Where does this overlapping happen?

Squidly271:
Follow the link

Squidly271:
It actually only seems to happen on the black theme, but I couldn't manage to figure out what was wrong with the css on the theme file, so put the css fix in the page instead which doesn't appear to affect any of the other themes.

bergware:
Haha, somebody blaming Eric and me ;)

I think the problem in the first place is the very loooonnnnggg description, make it shorter and the problem is solved, but mayby your solution is more appropriate!

PR #290, opened 2018-03-12
Update app icons container slidedown speed

realies:
Ain't nobody got time for that

### DIFF
--- a/plugins/dynamix/DashboardApps.page
+++ b/plugins/dynamix/DashboardApps.page
@@ -55,7 +55,7 @@ function loadlist() {
       if ($('.Panel.started').length===0) $('#noapps').show();
     }
     if ($('.Panel').length===0) $('#noapps').show();
-    $('#apps_icons').slideDown('slow');
+    $('#apps_icons').slideDown('fast');
     context.init({preventDoubleContext:false});
   });
 }


bergware:
Do you have a long list of containers + vms?

realies:
Further investigation shows that the `DockerClient.getAllInfo()` method takes more than 2000ms to complete on the 21 container instance I'm running. Further delaying the slideDown speed makes it unpleasant to watch.

realies:
Commenting out https://github.com/limetech/webgui/blob/fa8b4a931b01142ada2c2c02fa101cd52ac0a30e/plugins/dynamix.docker.manager/include/DockerClient.php#L327-L341 produces a consistent loading time reduction for `DockerClient.getAllInfo()` by about 1800ms.

realies:
For 21 container instances `DockerClient.getIcon()` takes about 150ms, `DockerClient.getControlURL()` about 1150ms and populating `$Registry`, `$Support`, `$Project` takes about 400ms. `DockerClient.getTemplateValue()` seems to be going through all template files every time it's being called. Perhaps load that data into memory and retrieve from there instead.

bergware:
My case:
I have 9 containers + 3 VMs. The complete rendering takes 517 ms.
Your hardware plays a role too!

realies:
@bergware, figured out it is actually 32 containers with 21 being on. Wonder what would be a good way to speed this up.

PR #297, opened 2018-03-15
Plug Security Hole In Docker Template PostArgs

Squidly271:
Hole allows arbitrary execution of bash commands on host

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -441,7 +441,9 @@ function xmlToCommand($xml, $create_paths=false) {
 
 function execCommand($command) {
   // $command should have all its args already properly run through 'escapeshellarg'
-
+  $cmdTmp = explode(";",$command);
+  $command = $cmdTmp[0];
+  
   $descriptorspec = [
     0 => ["pipe", "r"],   // stdin is a pipe that the child will read from
     1 => ["pipe", "w"],   // stdout is a pipe that the child will write to
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -422,6 +422,7 @@ function xmlToCommand($xml, $create_paths=false) {
       $Devices[] = escapeshellarg($hostConfig);
     }
   }
+  $postArgs = explode(";",$xml['PostArgs']);
   $cmd = sprintf($docroot.'/plugins/dynamix.docker.manager/scripts/docker create %s %s %s %s %s %s %s %s %s %s %s',
                  $cmdName,
                  $cmdNetwork,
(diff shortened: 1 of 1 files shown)


Squidly271:
I'm actually undecided on whether to plug the hole or not.  CA will NOT allow a template from a maintainer to include any bash command by default in PostArgs (automatic blacklisting of the template if the attempt is made).

But, there are / could be valid use cases for a user themselves to execute a bash command automatically after installation / editing of a container.  Up to you guys

eschultz:
Does this break if someone used a ; in a environment variable for docker?

Squidly271:
Yes it would.  Wasn't sure if the PostArg's couldn't be encapsulated via escapeshellarg or not. 

Escapeshellarg would fix the problem (if the quoting doesn't mess up the argument in the first place -> My initial thought is that it does.)  Put's everything into a catch-22, unless you only explode the postarg.  Further investigation after I posted the PR shows that everything else on a docker create / run command does have escapeshellarg, and what doesn't docker issues an error on:  (IE: I tried running a repository named xxxxx/yyyyy ; ls /boot and docker wouldn't create the file, nor was the ls executed.

limetech:
What's the verdict?  Do we need to merge this?

Squidly271:
Give me a couple hours.  When I'm at home, I'll readjust to only split on the PostArgs

bergware:
IF we want to inhibit arbitrary command execution then it should be performed in the calling function. Doing this in "execCommand" is the wrong place.

I've wrapped post-arguments in "escapeshellarg" in the calling function "xmlToCommand", this would stop arbitrary command execution, but it remains questionable whether we should allow the user to enter manually such commands or not.

Squidly271:
But the question remains, do the quotes around PostArgs affect them?  There's no postArgs on any app in CA to test with.

bergware:
I was trying to find which post arguments were used at the time (the initator of this feature), but forum find isn't very helpful...

ljm42:
This may help? https://lime-technology.com/forums/topic/58700-passing-commandsargs-to-docker-containers-request/

Squidly271:
I think that actually answered the question.  Snip from the docker log with this command:


`docker run -d --name='tshock-us2' --net="bridge" -p:7781:7777/tcp -p 7781:7777/udp -v "/mnt/user/appdata/world2":"/world":rw ryshe/terraria -world /world/Chmskb.wld`

`[ERROR] FATAL UNHANDLED EXCEPTION: System.IO.FileNotFoundException: Could not find file "/world/Chmskb.wld`

Now if I encapsulate the postargs in single quotes as this:


`docker run -d --name='tshock-us3' --net="bridge" -p:7782:7777/tcp -p 7782:7777/udp -v "/mnt/user/appdata/world3":"/world":rw ryshe/terraria '-world /world/Chmskb.wld'`


I get this repeated continually in the docker logs

`Choose World: Terraria Server v1.3.5.3

n       New World
d <number>Delete World`


So it *appears* that the quotes are messing up the post args, since the container no longer is looking for /world/Chmskb.wld

Net result is the escapeshellarg isn't going to work here, and I think ultimately you're going to have to split the postargs on a semicolon and hope for the best.  CA splits the post args on a semicolon to determine if theirs any excess commands in there.

Squidly271:
Moved the split on a semi colon from the command to postArgs only

PR #301, opened 2018-03-17
Revert "Plug Security Hole In Docker Template PostArgs"

limetech:
Reverts limetech/webgui#297

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -422,7 +422,6 @@ function xmlToCommand($xml, $create_paths=false) {
       $Devices[] = escapeshellarg($hostConfig);
     }
   }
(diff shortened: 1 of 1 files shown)


Squidly271:
I confused about you reverting this commit.  Net result of the revert is that you're stopping the command on the first semicolon.  I thought that we were going to not do that due to a possible semicolon being in a environment variable, and the safer approach was to only apply the fixes to the PostArgs (reverting this reversion will do just that)

limetech:
I reverted it because after merging that PR it resulted in merge conflicts with bergware's PR, so I reverted yours and then merged bergware's.  Next I was confused because github closed the reverted PR, don't know wtf happened there.  So I went in made the same change that was in that PR in the code and pushed it.  Please review the code and see if it is correct or  not.

Squidly271:
Reopened #303

PR #304, opened 2018-03-17
Docker Security: Remove <script> and <iframe>

Squidly271:
Recursively remove any <script> or <iframe> from all elements in a template prior to displaying.

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -374,7 +374,25 @@ function xmlToVar($xml) {
       }
     }
   }
+  xmlSecurity($out);
   return $out;
+}
+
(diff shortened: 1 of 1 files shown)


bergware:
Did you encounter situations where people put these "tags" and you are actively removing them or is it more a precaution thing?

Squidly271:
Preventative medicine.  CA's has this in (stronger actually) forever and no one has yet tried.  But with the ability for dockerMan to update the user's template behind the scenes, it needs to be there.  Currently, if you modify a user template, and add in
`[script]$('#test').html('test failed');[/script][span id='test'][/span]`
the script will execute no problems what so ever when displaying the template (edit) in basic mode.  And, because the csrf is already part of the form, the script could potentially do some major damage.  That, and I had nothing else to do when I did the PR  :)

PR #310, opened 2018-03-21
Added "Update all Plugins" button to plugins page

bergware:


bergware:
It was to be expected. The question came "can we get update all plugins" similar to what is done for containers. So I added a button at the top "Update all containers" which only becomes available when 2 or more plugins have a pending update.

bergware:
The PR about _Apply button_, involves all files which have some settings to apply. By making Apply default disabled, the button stays visually disabled for pages which have a longer loading time.
I updated the copyright to 2018, so it was easier for me to track which files are changed.

bergware:
I revised the execution of docker commands to be done with a timeout of 20s. Don't know if that is really necessary (because I can't replicate the issue), but at least it should prevent waiting indefinitely for docker.

PR #312, opened 2018-03-24
Minor GUI adjustments

bergware:


bergware:
The VM 'detach' button fix has some additional changes
1. Add a cookie variable which remembers the "open" VMs in the list. When the list refreshes the opened VMs stay open
2. Do the refresh of the VM inline without reloaded the complete page
3. Corrected the SWAL message not being displayed when an error occurs
4. Added spinner when the detach function is executed

**Outstanding problem**: The detach function fails with the message "Unable to attach disk".
@Eric: can you have a look? I don't know why it fails

bergware:
Still can't reproduce the container list hanging error. In an attempt to minimize dependency (and speed up a little) on docker processing. The individual calls per container to obtain the different docker information are now grouped in a single call to docker.

bergware:
With the latest optimizations load times again halved. Dashboard takes only 125ms to load containers+vms .

eschultz:
@bergware From the libvurt source it looks like you can only 'detach' a disk when the VM is started.  Furthermore, you cannot detach a cdrom (ISO).  Not sure how that feature ever worked in the unRAID web interface but seems kind of useless or rather unsafe (imagine detaching the C:\ drive while Windows is running).

eschultz:
fixed_the_glitch
Ended up removing the whole Actions column in https://github.com/limetech/webgui/commit/dac97c0eeea1dd70db7e3aeaf505972c482d36ae

PR #315, opened 2018-03-25
Docker: Selectable Global Container Stop Time

Squidly271:
Docker defaults to 10 seconds for a graceful stop before forcibly attempting to kill the container.  Some containers may require more time to exit.


bergware:
All containers I am using stop within 2 seconds max. Do you have examples of containers which take a very long time to stop?

PR #318, opened 2018-03-27
Fix Create Docker php warning

### LINKED ISSUE #316: CreateDocker warning
Creating a new container using the simple UI view produces:
`Warning: count(): Parameter must be an array or an object that implements Countable in /usr/local/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php on line 222
`

Squidly271:
Fixes #316

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -194,6 +194,7 @@ function postToXML($post, $setOwnership=false) {
   $xml->addChild("Data");
   $xml->addChild("Environment");
 
+  $post["confName"] = is_array($post["confName"]) ? $post["confName"] : array();
   for ($i = 0; $i < count($post['confName']); $i++) {
     $Type                  = $post['confType'][$i];
     $config                = $xml->addChild('Config', xml_encode($post['confValue'][$i]));


realies:
This makes no sense. You are reassigning the variable to itself if it is an array or creating an empty array if it doesn't exist. In the case of #316, the following statement would cycle through an empty array and do nothing, so you just waste some memory to create an array that's never getting used.

limetech:
The code you are replacing is correct.  If you are objecting to the knat's ass difference in execution speed, then your code is also wrong.  It should be:
`if (isset($post['confName'])) {`
You don't want to test for an array because it's an error for anyone to pass a value that isn't an array in this case and should be warned about it.  You also should have a space between 'if' and the opening parenthesis.

But the objection to your PR does not come from a code nit-pick.  It comes from the using the word "sanity" in your description, which implies that somehow the original coder is not as smart as you.  There is a technical term for a person who frames things like this: dick.  Now I'm not say you are a dick, just that describing changes like this makes it appear that way.

realies:
I have not reviewed the chain of code that populates the `$post['confName']` variable. The reason for testing for an array after making sure the variable is defined is because the next step in the code is to iterate with a `for` statement which uses `count`. If for some reason `$post['confName']` is anything else than array, `count` would return 1, which would produce a single iteration of the loop and cause unhandled exceptions.

As per the space between `if` and `(`, you are giving a great example of why CI checks would be useful.

As per the objection, my PR comes from the lack of insight into the pull request change that got merged and then restyled by contributors. The Oxford English dictionary defines `sanity` as `the ability to think and behave in a normal and rational manner`. I stand behind this statement with the following pull request and am positive that it is saner than what has been merged into master.

To me, it seems that you are overexposing my addition of `+ sanity` to the pull request title with your understanding that this implies greater mental capability in general than the original contributor. In the essence of things, pull requests are done for specific 'problems', for this specific 'problem' there is a saner solution. I would be interested in the logic about how this goes outside that boundary and seemingly goes on a personal, ego level.

For the record, @Squidly271, do you feel offended by this? Has my addition of `+ sanity` to the following pull request made you feel inferior or made me demonstrate greater mental capabilities in your eyes?

limetech:
Yeah, well the Urban Dictionary defines "dick" as "an abrasive man" which I would expand to "an abrasive person".

realies:
Urban Dictionary is not the main historical dictionary of the English language. Definitions outside this can vary to a great extent.

Squidly271:
Offended by the choice of sanity in the PR: No.    LOL: If anything it actually made me feel superior.

For the record, I am not the original author of this file.

You felt that it was entirely incorrect and insane to run a for/foreach loop against an empty array, and prefers an if conditional followed by the loop.  Personally I think that's nuts because there is nothing wrong with running a loop against an empty array (it won't execute), and in this case with the for it wouldn't have executed either.

Personally, I wouldn't have used a for() either had I done the original code, but the thing is that the code in question was already tested, works perfectly and needs no further debugging / testing so why bother changing it (oh and btw, your issue report also didn't bother specifying the circumstances under which the warning appeared, so it was up to me to determine those circumstances, replicate, fix, re-test. - And the warning came up due to an edge case of your install as it has not been reported previously)

Nitpicking about execution speed:  Minuscule gains only realized on a large set, and you didn't complain about the original author's using the count() inside the loop where it had to be re-evaluated with every iteration.

Switching to foreach: Sure.  If the code was being re-written from scratch, I'm sure there's tons of optimizations that can be done across the board, but there's better things to do than test it under all circumstances (which it would have to be), and the minuscule speed increase isn't worth my time (or anyone's).  This is the real world: If it ain't broke, don't fix it.  And if you do, be prepared to test it under all circumstances and not just assume that there's no way it can fail / cause issues.


My using is_array instead of isset or the null coalesce operator (as bonienl's revision utilizes) is more of a preference / style as to how things should fail.  I don't want the user to see a warning in the OS under any circumstance whereas limetech as mentioned prefers that.  (But this choice is also heavily influenced by how I have to handle CA).  Doesn't bother me they revised the conditional.

As you state, 

> for this specific 'problem' there is a saner solution

except that this is entirely your opinion, and is fundamentally based upon your style.  There are always multiple solutions for any given problem, and if they all work, and are bug free they are all equally acceptable.  I certainly didn't bitch and complain over bonienl's revision.  And my fix did make sense, is acceptable, falls within the style already used across the OS.  Just not by you.

> The Oxford English dictionary defines sanity as the ability to think and behave in a normal and rational manner. I stand behind this statement

Now you're just digging yourself a bigger hole in attempting a defence, since the implication is that I am not doing such.  I believe the reverse is true.

> Urban Dictionary is not the main historical dictionary of the English language. Definitions outside this can vary to a great extent.

I think that it's a more sane to go with Limetech's definition (which I would think is also more universally accepted).

realies:
`... so it was up to me to determine those circumstances, replicate, fix, re-test. - And the warning came up due to an edge case of your install as it has not been reported previously`

Glad you are in no way offended and that you are saying this. It puts the focus back to the actual issue.

`actually made me feel superior`

Both of our commits resolve a symptom of a still unknown problem. Since the reported warning is now muted (_which seems to be the main priority?_), it would more difficult to track the origin of its occurrence. Perhaps [the code of conduct](https://help.github.com/articles/adding-a-code-of-conduct-to-your-project/) should prohibit working on issues that can't be reproduced by other developers and the response to my bug report should be refusing to take any actions before there is sufficient information to reproduce it.

Squidly271:
> still unknown problem.... warning is now muted (which seems to be the main priority?).... track the origin of its occurrence

It's not unknown, nor is the warning "muted" at all.  It was simply that the original designer of the module did not anticipate the circumstances you were using the module.  As I stated, I determined those circumstances, replicated, fixed, and retested. 

>  track the origin of its occurrence

Is this the root of your critique of the commits?  If so, that's where you're off-base.  Had the issue not been reproducible under exact circumstances, there would have been no "work-around" commits.  IE: You can't fix what ain't broke.

In this case, you had no <config> elements at all in the template (no path mappings or ports).  The code simply assumed that there was always going to be one.

realies:
The circumstances were to install a Docker container directly from the Docker hub. The use case is to run a container that is not based on a template. Is that an equivalent to `you had no config elements at all in the template (no path mappings or ports)`?

PS: The root of any critiques so far is based on better ways to resolve unexpected system behavior.

Squidly271:
>  Is that an equivalent to you had no config elements at all in the template (no path mappings or ports)

Sure.  If you didn't add any

PR #320, opened 2018-03-28
Diagnostic improvements

Squidly271:
Suggestions from this thread: https://lime-technology.com/forums/topic/70460-how-to-read-diagnostics/
Moved log capture to very end so its as current as possible


bergware:
Tested and looks good. Nice additions.

Squidly271:
And I thought of you and switched an errant tab from a prior PR to be spaces

bergware:
I saw that! Excellent move :)

PR #322, opened 2018-03-28
Fix Create Docker php warning + sanity

realies:
Any reason for the enclosing for() to not be a foreach()? The latter is faster.


limetech:
What do you mean by "sanity"?  Do you mean converting the style of that for-loop?  If so, why didn't you do it?

realies:
@limetech, https://github.com/limetech/webgui/pull/318#issuecomment-376781033

bergware:
Please close this PR, it has been superceeded with [#324](https://github.com/limetech/webgui/pull/324)

PR #331, opened 2018-04-19
Plugin manager: Minor code optimization

bergware:


bergware:
Ps. reverted another "new" feature. This can wait.

PR #338, opened 2018-05-03
Docker: support custom bridge networks without port bindings

bergware:


bergware:
My local git application stopped working and I needed to update it. This caused some confusing at my side, but all should be okay again.

bergware:
Many changes, but I am satisfied now...

PR #339, opened 2018-05-07
Add support for Docker Labels to Docker plugin

shayne:
Docker object labels:
https://docs.docker.com/config/labels-custom-metadata/

More and more containers are taking advantage of container metadata
where environment variables are not necessary.

Traefik is a good example:
https://docs.traefik.io/configuration/backends/docker/#on-containers

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -192,6 +192,7 @@ function postToXML($post, $setOwnership=false) {
   $xml->Networking->addChild("Publish");
   $xml->addChild("Data");
   $xml->addChild("Environment");
+  $xml->addChild("Labels");
 
   $size = is_array($post['confName']) ? count($post['confName']) : 0;
   for ($i = 0; $i < $size; $i++) {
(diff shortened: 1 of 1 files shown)


Squidly271:
Without trying out the code, shouldn't there be another %s on line 449 since you're adding another entry?

Also, for CA purposes, if a maintainer includes labels in a default template does that mean that the container won't run as is on say 6.5.0 stable without moving the labels to instead be extra arguments?

shayne:
**Re: Community Applications (CA)**
Labels today have to be added via Extra Arguments. The template could include additional information in the Overview, similar to the "Directions" sections we see now. Given that Extra Arguments is hidden under Advanced View, populating them placeholders is probably not the best idea.

Looking at this from a practical use case, however. Containers like nignx-proxy provide a reverse proxy to other containers, and the user adds compatible environment variables to existing containers (outside of what those containers templates' provide). Labels tend to follow the same pattern...

That is you'd setup the Traefik container (no labels are setup on the container itself). Then you could go to your, let's say Python web server container and add the labels as specified by Traefik's documentation. Similar to the nginx-proxy, these labels are added as an additional step in order to setup and configure the reverse proxy. Which isn't an assumption on the Python web service container, but rather an expected additional step to connect the two together.

bergware:
Tested your PR and label creation works fine.
Any new labels will become part of the stored template, there is no issue with earlier unRAID versions as these would simply ignore the unknown XML tags.
I think this is a useful addition.

limetech:
Thank you!

PR #347, opened 2018-06-23
Plugin List: Show support thread if attribute exists

Squidly271:
Supports a new attribute on plugins (**support**) which will contain a link to the appropriate support thread in the forum. 
Easiest way to test is to remove then reinstall CA which now has that attribute present.
If this PR does get added to dynamix, then at time of release I would publish a script to update all the .plg's present on a user's system with the appropriate attribute, and also have CA on installation of a .plg automatically add it to the .plg if it doesn't already exist (since I insist that CA's XML file has a support entry)
Only dmacias would have to adjust his README.md files as he is the only one who includes a link to support already contained within it.


Squidly271:
Script and CA is ready to go https://github.com/Squidly271/community.applications/blob/master/source/community.applications/usr/local/emhttp/plugins/community.applications/scripts/updatePluginSupport.php

PR #359, opened 2018-08-01
Use timeout in command execution of diagnostics

bergware:
To prevent hanging of diagnostics, commands are killed after 30s.


Squidly271:
Cool.  I can stop maintaining my autostart manager :)

bergware:
I introduced some regression errors in Docker. Need to fix that :(

limetech:
Looks like this is ready to rock 'n roll!  Merging...

Squidly271:
Just a comment on docker CPU pinning.  I would like to see that if ALL cores are checked off, then dockerMan does NOT add any --cpuset-cpus to the docker run command.  (ie: if all cores are selected, then don't enter anything into the <CPUSet> tag at all.

My feeling is that users would wind up checking off all cores (when they don't have to), and if/when they upgrade their CPU, then they would be surprised that their containers are no longer running on all the possible cores without editing each container.  

Related, I would also prefer to see all cores checked off by default if nothing is entered into <CPUSet>  

Trouble is that this behaviour I'm proposing is opposite to the behavior of the VM manager where you have to select everything.  Up to @bergware

PR #360, opened 2018-08-12
Add help text to docker CPU pinning

Squidly271:


Squidly271:
No point in merging if my suggestions to behaviour change on #359 gets implemented

bergware:
The NO cores warning message is for VMs only (don't get fooled by it).
The current implementation is as you describe. Containers can have no cores assigned.

Squidly271:
Yes, but it's confusing to the user having no checkmarks by default, and if they do check off everything, then on an upgrade of the hardware then the container(s) will not operate on all cores.

bergware:
I guess you need to try first, it is not as unlogical or confusing as you think.

Squidly271:
Of course I tried it first, saw the resulting template for when everything is selected, and when nothing is selected, and made a suggestion.  My point being that as it stands, selecting no cores means that the container runs on all cores.  Selecting all cores means the container runs on all cores.  All I was suggestion in a nutshell was that by default all cores are checked off, and in the event that all cores are selected that no entry into CPUset gets added (since that's the default behaviour of docker)

bergware:
Say this protection is in place. The following will happen
When the user selects all cores and presses update, the page comes back showing all is selected.
Now if the user exits the page and returns he will see none of the cores selected.
At this point he might think something went wrong and does it again.
In short it adds another type of confusion.
The current approach doesn't forcibly change the user selection, but requires him to read the help :)

Squidly271:
No it wouldn't.  If there is no entry at all in CPUset, then have all cores checked off.  If the user makes changes, and winds up selecting every core, then also clear CPUset.  IE:  CPUset only has entries if only **some** of the cores are selected.

> requires him to read the help :)

LMAO.  Yeah, reading help text is something that users are known for on the forum.

It doesn't matter to me personally.  It took me < second to figure out what was going on.  It will solve a lot of confusion and posts on the threads.  

Additionally, my suggestion has the upshot that if a user doesn't make any changes to the pinning, then on an eventual CPU upgrade the container will still execute on all cores instead of being limited to the number of cores they checked off on the original CPU thinking that "I have to check off every core so that it will run on them"

BTW, looking forward to the dedicated CPU pinning page.  :)

PR #361, opened 2018-08-13
VMs: preserve XML custom settings in VM config updates

bergware:


bergware:
This update is about keeping custom XML modifications done by the user while the VM gets updated via the GUI, e.g. cpu  pinning or memory allocation is changed and updated.
The trick to make this work is converting the complete XML settings into an array, apply the changes and convert the array back to XML.
This requires additional conversion routines to go back and forth. I lost time because the general approach is to convert XML to json and then json to an array. Almost every solution points to the json_encode/json_decode functions of PHP, but this doesn't work when a XML node has both attributes and values. Attributes are lost in this case, and it took me some time to get this right. Things became even more complicated because of the usage of the  'unraid' namespace attribute, which eventually led me to a bug in VMedit.php (correction included in this PR).
I don't have much VM customization myself, sofar all is working great for me, but obviously this needs more thorough testing. This PR would make life a lot easiers for those who have (lots of) modifs though!

bergware:
Hmm, I am not done yet. The "XML view" updates don't get thru...
More to come.

eschultz:
Wow, nice work @bergware ! 🎉   Many people who have custom XML will greatly appreciate this PR.  I'll need some time to test this out.

Sorry about the custom 'unraid' namespace, I seem to remember libvirt being picky about this if it wasn't a custom namespace.

bergware:
I didn't do the LibreELEC and openELEC forms. Is it benficial for these forms or these don't require manual tweaking?

bergware:
I need to make support for USB + PCI devices, these are not properly added or removed.
Working on it.

bergware:
Several attempts, but I think I got it right now :)

bergware:
@eschultz you may want to look at the legacy VM support in VMedit.php (line 60)
It doesn't look quite right, but I have no clue what it should be.

bergware:
Have a look at the revamped themes, they became my favorite :)

Squidly271:
> Have a look at the revamped themes, they became my favorite :)

I like the revamped styling of the black/white menu bar (never quite like azure / gray).  Undecided though on the Red/Orange button colors.  Of course this means another CSS update to CA when this gets approved though to fix some minor issues on displays with the revamped theme.

bergware:
@Squidly271 when you start looking into CA. Two observations I have

1. During system startup CA comes with an error message "**cat: write error: Broken pipe**". This doesn't happen in earlier versions of unRAID, some mismatch between 6.6 and CA

2. The font size of the footer changed (smaller) when the CA page is opened. Perhaps CA uses the same ID name as what is used for the footer?

Ps. there is a reason for the new button style, you'll find out soon(tm).

Squidly271:
> During system startup CA comes with an error message "cat: write error: Broken pipe". This doesn't happen in earlier versions of unRAID, some mismatch between 6.6 and CA

Are you sure it's CA _proper_?  There's no cat command in the .plg, the .page, and CA doesn't use any events.  I'll look at my other plgs later on.

> The font size of the footer changed (smaller) when the CA page is opened. Perhaps CA uses the same ID name as what is used for the footer?

If this is the case, I was there first :)  I'll check it out...

bergware:
When I uninstall CA the error message is gone. Not sure why it occurs, I have a look too.

Squidly271:
The CSS (I see that "Array Started") is shrinking when going into Apps.  At work, so can't do a perfectly indepth investigation, but I don't see any conflict or inherited CSS on the element.  Will tear through everything tonight and try and see where if anywhere its changing the style.

bergware:
You can solve the "broken pipe" error by changing this line in the PLG file:

rm -f $(ls /boot/config/plugins/&name;/&name;*.txz 2>/dev/null|grep -v '&version;')
into
rm -f $(ls /boot/config/plugins/&name;/&name;*.txz 2>/dev/null&#124;grep -v '&version;')

In other words replace the | (pipe) character by its html representation &#124;

limetech:
Amazing

Squidly271:
> The font size of the footer changed (smaller)
That was fun.  I missed a &lt;/font&gt; on a form that is set to be display:none.  Added it and your Array Started stays the same size

PR #363, opened 2018-08-19
dockerMan: Avoid filename collisions on FAT32

Squidly271:
Because under FAT32, my-plexmediaserver == my-PlexMediaServer.  Solves an edge case


Squidly271:
Story of my life:  Fix an edge case in unRaid that only I noticed, and the fix under certain circumstances causes an edge case scenario in CA.

limetech:
The bane of programmers everywhere: side effects.  btw it's Unraid now :)

Squidly271:
> btw it's Unraid now :)

Even when it was unRAID, I still typed unRaid.  Muscle memory.

PR #369, opened 2018-08-28
Fixed regression error in parity history

bergware:


bergware:
One day I need to sit down and see if I can simplify and cleanup the styling and get rid of legacy stuff.
Not for now though.

PR #370, opened 2018-09-01
Remove DockerRepositories.page

Squidly271:
Contentious, and not really expecting for the PR to be approved, but with CA being linked to directly on LT's web page, why even include the DockerRepositories page anymore?  And, all of LT's wiki docs since 6.2 have recommended to install CA for any additional apps for unRaid.


eschultz:
screen shot 2018-09-03 at 9 44 23 pm

We discussed this briefly today.  Don't think we're going to make any major changes in 6.6.  If we're going to remove DockerRepositories we might as well add in CA.  That's more of a change for 6.7 👨‍💻

PR #378, opened 2018-09-04
Added CPU isolation

bergware:
Both CPU pinning and CPU isolation is possible thru the GUI


Squidly271:
Without adding this PR, if you isolate a core, will the docker pinning then disallow the isolated cores?   Pinning a container to one or more isolated cores will effectively only allow execution on the first selected core.

bergware:
The current implementation doesn't have restrictions.
I am no expert in this field, but why would you make isolated cores for containers?
Isn't there a conflict situation, because containers run under the host which has these cores excluded?

Squidly271:
> The current implementation doesn't have restrictions.
I am no expert in this field, but why would you make isolated cores for containers?
Isn't there a conflict situation, because containers run under the host which has these cores excluded?

Not a conflict per se.  What isolating CPUs actually does is stop the context switcher from utilizing those cores.  Doesn't actually prevent any executable from manually utilizing them via the appropriate commands (ie: pinning a container to an isolated core)  But, since the context switcher won't operate on those isolated cores, Linux won't switch the threads back and forth between isolated cores automatically.  Net result is that if a container is pinned to multiple isolated cores, it will only actually execute on the first core.

Would just solve some problems on the forum by removing the option to pin a container to an isolated core.  (Plus it makes no sense anyways, because even without knowing the fundamental goings on beneath the surface, by isolating cores you are telling the OS to not use the cores for anything, but to reserve them for VMs)

Of course the complication arises if a user has already pinned cores, and then chooses to isolate them.  But, that would be a whole different story catching that situation at time of docker run and adjusting the pinning accordingly.  Probably something more suited for FCP to catch.

Side related thought:  is there a variable (or can there be) that shows the mode unRaid booted in (GUI / normal)

limetech:
The cephalopod is correct: pinning multiple isolated CPU's to a container only lets container utilize lowest numbered CPU.  This is a kernel "issue" though kernel guys would probably argue it's a kernel "feature".

I think the best approach is to prohibit users from selecting isolated cores for container cpu-pinning.  If user really wants to dedicate a single core to a container (maybe they turned off hyperthreading or CPU does not have hyperthreading), then they can turn off isolcpus, pin their cores the way they want, then turn isolcpus on the way they want.  FCP might gripe but they'll know why it's griping.

re: boot mode: you could check for existence of /root/.mozilla/ or /root/.Xauthority/ which are only present in GUI boot mode.

bergware:
Adding restrictions may lead to confusion?
Perhaps we leave the assignments for the moment unrestricted and wait for user feedback.
I believe people also need to educate themselves when "toying" with this.

limetech:
Yes that's fine.

PR #379, opened 2018-09-06
Docker: Suppress error if no volumes present in template

Squidly271:


bergware:
I am not sure this works:  **$ct['Volumes'] = is_array($ct['Volumes']) ?: [];**
In case the array exists it will be set as "true" which is returned by the "is_array" function.

Squidly271:
Yeah you're right.  Weird thing though is that it worked lol  Fixed proper.  TY

eschultz:
Thanks!

bergware:
Just for fun (and own interest) I made a small test
```
$test = [1,2,3,4];
$test = is_array($test) ?: [];
echo print_r($test,true);
```
This sets $test = 1 (true) and the array content is lost

Squidly271:
> This sets $test = 1 (true) and the array content is lost

Yeah, I was looking for PHP errors when I did the patch on my system. and a foreach (true as $xx) does not generate a warning / error, so missed the fact that I lost all the volumes on the page.  

I'm allowed to make a mistake once a year ;)

bergware:
No worries, that's why we have safety nets aka code reviews ...

PR #381, opened 2018-09-08
CPU Isolation: Prevent PHP error if no CPUs isolated

Squidly271:


bergware:
Yeah, I overlooked some initialization.
$isol is actually a string, better to set as an empty string

Squidly271:
> $isol is actually a string, better to set as an empty string. 

Technicality.  I thought about that originally setting it to be "", decided that it didn't matter either way and just deleted the line.  But, have updated it accordingly

bergware:
I believe it is okay, but for semantics it is clearer/neater to initialize $isolcpus as an empty array. As you initially did.

Squidly271:
As you wish.  But, it did work as is.  Different styles.

bergware:
PERFECT :)

Squidly271:
Yes, coding style is now "Dutchified"

bergware:
Yeah, we like to be on top of things (or it is perhaps we are just tall people)

Squidly271:
https://i.redd.it/tvxqbdux6uqz.jpg

Squidly271:
[https://i.redd.it/tvxqbdux6uqz.jpg](url)

Squidly271:
image

bergware:
We wear the wrong (wooden) shoes ...

Squidly271:
Gotta say though, that out of all the caricatures, the Canadian is the only one that looks cool :)

PR #383, opened 2018-09-09
Diagnostics: Don't include share's cfg if share isn't present

Squidly271:


eschultz:
Only bad thing about this is if they stopped their array then all the share's cfgs aren't included

bergware:
Instead of skipping the file, it can be given an extension to indicate it wasn't found or the array was offline.

limetech:
Actually please leave as-is.  The absence of presence of a share's cfg file can explain other things.  For example, if a system share cfg file exists (say domains.cfg) but share does not, the domains share will not be auto-created.  If someone asks, "Why didn't my domains share get created?" that could be the answer.

Squidly271:
It's just an annoyance when looking at diagnostics from people who have previously misconfigured their system, and have tons of .cfgs that aren't relevant anymore.

limetech:
Agreed.  Might be better to add some kind of cleanup utility, or even have system auto-delete or rename orphan cfg files.  We should give this some thought.

PR #390, opened 2018-09-27
Make font sizes relative

bergware:


bergware:
Font sizes are now all relative to the font size setting of the browser (default is 16px).
This means people can make the font go larger or smaller by changing their browser font size setting.
All text scales automatically.

Squidly271:
And why do I have a sneaking suspicion that an update to CA will now be required...

bergware:
That's only a suspicion... CA is still looking good (same as before) unless you made hardcoded font sizes.

Squidly271:
> CA is still looking good (same as before) unless you made hardcoded font sizes.

And that's why it's still looking good - hardcoded font sizes.  Problem with CA is that the individual apps are set to take up x-px across, and if we increase the font-size on the browser (and CA uses that size), then it will start looking terrible due to massive amounts of scroll bars suddenly appearing on the templates.

On the other hand, CA has for the longest time automatically adjusted itself to the zoom level of the browser.

That being said, there are actually a couple of spots where a font-size change on the browser does mess up CA, and I will have to make adjustments accordingly.

PR #397, opened 2018-10-05
Change anonymized email address

Squidly271:
lol  Whenever I look through someone's diagnostics, I always think that they have a porn addiction


bergware:
You don't like "sexy" code ;)

PR #399, opened 2018-10-07
Add global CPU core pinning for containers

realies:
Should make possible to (re)allocate CPU resources across all containers faster. Currently, it requires manual clicking on thread numbers per container individually, which in a large pool of containers can be very repetitive. The change allows for thread numbers to be clickable, doing so enables or disables thread pinning across all containers.

### DIFF
--- a/plugins/dynamix/CPUvms.page
+++ b/plugins/dynamix/CPUvms.page
@@ -119,6 +119,12 @@ function vm() {
     buttons(document.vm);
   });
 }
+function thread2containers(n) {
+  const selector = $('form[name=ct]').find(`[name$=":${n}"]`);
+  const checkboxes = selector.length;
+  const checked = selector.filter(':checked').length;
+  selector.prop('checked', (checkboxes - checked > checked ? true : false)).change();
+}
 function ct() {
   // fetch the current container assignments
   $.post('/webGui/include/CPUset.php',{id:'ct',cpus:'<?=$cpuset?>'},function(d){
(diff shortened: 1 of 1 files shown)


bergware:
Couple of things to further improve
1. Add a "title" message so users can see what it means when clicking on a core number
2. Make it a toggle function. Selected cores become unselected and vice versa

realies:
Thanks for the comment
1. Assuming you mean an alt message for the links? Or an expanding info box explaining what that is?
2. Was refraining from toggling as it would not be as straightforward to (re)assign all containers to specific cores but only invert the existing selection.

realies:
Thinking that this should cover most use-cases.

realies:
How do you get the table to reset?

bergware:
Make a change, the "Done" button changes to "Reset" (to undo the changes)

bergware:
Please change to regular quotes in the line below to adhere browser standards.
elem.innerHTML = elem.innerHTML.replace(/(\d+)/g, `<a href="#" onclick="thread2containers(this.innerText);return false;" title="Toggle thread to containers">$1</a>`);

realies:
Which standards? If you mean the back-tick quotes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

bergware:
Seems Safari doesn't support it.
Those back-tic quotes are not used anywhere else in the GUI. I think it is better/safer to use regular quotes, funtionalwise all stays the same.

realies:
Tested on iPhone/iPad/macOS Safari and everything works as expected.

Issue #400, opened 2018-10-09
Documentation on setting up a dev environment. A README.md?


williamkunz:
I am a long time user and would love some direction on how to get a dev environment going so I can contribute.

ljm42:
Sorry, we aren't really setup to handle support/feedback here in GitHub. I'm closing out these really old issues. 

If you are still interested in developing for Unraid I'd recommend checking out the API https://github.com/unraid/api

PR #403, opened 2018-10-16
Improve private docker registry support (correctly pulling and updating containers)

rhurling:
This is a rewrite to allow better support for private docker registries (in my case, free gitlab.com).
This rewrite is not tested on an unraid install, since I'm not sure how to set that up (see #400).

The code itself is tested in so far, as I ran it outside of unraid to get the api flow correctly and to make sure the `/root/.docker/config.json` is parsed correctly.

Following things could be problematic:
- Needs read access to `/root/.docker/config.json`
- Anybody wishing to use this would need to make sure `/root/.docker/config.json` is persistent across reboots

Otherwise it would basically be following a normal usage of private docker registries, for example Gitlab:
- Create gitlab account and repo, push docker image to docker repository
- Login to private docker repo via `docker login [...]` on unraid via ssh
- Installing and updating docker images via unraid webui should now work

### DIFF
--- a/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -14,6 +14,8 @@
 <?
 $docroot = $docroot ?? $_SERVER['DOCUMENT_ROOT'] ?: '/usr/local/emhttp';
 
+require_once "$docroot/plugins/dynamix.docker.manager/include/Helpers.php";
+
 $dockerManPaths = [
 	'autostart-file' => "/var/lib/docker/unraid-autostart",
 	'update-status'  => "/var/lib/docker/unraid-update-status.json",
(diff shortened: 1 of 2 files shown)


rhurling:
Great! I think there could still be improvements made in regards to public docker images on a different registry (not hub.docker.com). See 1. in this comment https://forums.unraid.net/topic/75023-docker-improve-support-for-private-registries/?do=findComment&comment=691020
Because at the moment if no auth data is found in `/root/.docker/config.json` it will still use the default registry api (hub.docker.com) instead of a given one.

Should i try to make some improvements here or should i do that in a separate pull request after this one is merged?

limetech:
eschultz is in process of testing current PR with aim of merging for next Unraid OS 6.6.3 release which is scheduled for later today.  If you want to take some time and make additional improvements we can skip this PR for 6.6.3 and get into the following release.

You can create the symlink in your 'go' file in order to keep the config.json retained across reboots.

rhurling:
I'll create a new merge request for the improvements after 6.6.3 is released.

PR #408, opened 2018-11-01
Diagnostics Add share storage location

Squidly271:
To help with the usual questions of why are all my drives spinning up etc.  Also helps with diagnosis with mover logging being disabled by default.


limetech:
thank you

PR #409, opened 2018-11-02
Diagnostics Update to #408

Squidly271:
More generalized.  Actually itemize which disks the share is stored on.

### DIFF
--- a/plugins/dynamix/scripts/diagnostics
+++ b/plugins/dynamix/scripts/diagnostics
@@ -28,7 +28,7 @@ function exert($cmd, &$save=null) {
   return implode("\n",$save);
 }
 function shareDisks($share) {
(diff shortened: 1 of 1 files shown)


limetech:
What's "/mnt/RecycleBin"?  That a mount point?  If so, what's mounted there?

Squidly271:
On my system with the recyclebin plugin installed, that is present in /mnt  Contains (not surprisingly) the files that have been deleted.  Don't ask me why dmacias chose that location.  Simply wanted for the change to only pick up unRaid stock items.

limetech:
If you look at this file:
https://github.com/limetech/webgui/blob/master/plugins/dynamix/include/Browse.php

On line 39 it uses 'getfattr' command to fetch Unraid-specific extended attribute called 'system.LOCATIONS' that lists all the disks that a share is present on.  You could use this extended attribute value in the diagnostics report.

For completeness: there is also 'system.LOCATION' which is valid for any file in Unraid user share file system that tells you what disk that file is located on.  We use this, eg, when a VM is started to pass the actual vdisk disk paths, so that I/O bypasses FUSE.

Squidly271:
Updated

PR #411, opened 2018-11-09
Black & White theme: make select arrow consistent for all browsers

bergware:


limetech:
I don't mind if it defaults to what's called 'advanced view'. However to swap the meaning will lead to unnecessary confusion.  Maybe get rid of basic/advanced terms and use a 'Raw' (yes/no) toggle instead?  (if toggle set to yes then you see and can edit the raw file)  Default: no

bergware:
I admit, I was lazy.
True it is confusing as it is now...

bergware:
Updated the code to properly reflect "menu view" and "raw view".
Changed the label from basic/advanced to menu view/raw view

PR #412, opened 2018-11-10
Revert FAT32 filename collisions

Squidly271:
#363 has been implicated in a glitch in the docker template system.  While I cannot replicate the issue at all, revert the PR as the edge case issue that it solved has only ever occurred during synthetic tests.  IE: revert the code and see if the reported "issues" still occur.

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -81,16 +81,7 @@ if (isset($_POST['contName'])) {
   $userTmplDir = $dockerManPaths['templates-user'];
   if (!is_dir($userTmplDir)) mkdir($userTmplDir, 0777, true);
   if ($Name) {
-    $filename = sprintf('%s/my-%s', $userTmplDir, $Name);
-    // look for FAT32 filename hits due to different case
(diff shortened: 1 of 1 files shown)


limetech:
Ok I'll accept this PR, and here's what I'll say about the issue.  First, I see the concern that because FAT32 (and NTFS too btw) filenames are case insensitive, be nice to user if there is a collision.  However I would say the better way to detect and handle this is using javascript on form where user sets the name to begin with and don't let them enter a name that collides with existing names.

That said, the real flaw is the docker config data being stored on the USB flash.  It was done this way because 'dockerMan' predates Unraid 'system' share support.  But really the docker container config files should be moved to /mnt/user/system/docker.  This is a 'todo' item for us on the ole 'laundry list'.

Squidly271:
> That said, the real flaw is the docker config data being stored on the USB flash. It was done this way because 'dockerMan' predates Unraid 'system' share support. But really the docker container config files should be moved to /mnt/user/system/docker. This is a 'todo' item for us on the ole 'laundry list'.

Would've been nice if I could replicate the problem that users were having with my original code.

Moving everything over to /mnt/user/system across the board is probably way over due..

PR #416, opened 2018-11-25
Add per DIMM information toggle

realies:
Hidden by default and showing when **Memory** is clicked.

image

### DIFF
--- a/plugins/dynamix/include/SystemInformation.php
+++ b/plugins/dynamix/include/SystemInformation.php
@@ -134,7 +134,7 @@ for ($i=0; $i<count($socket); $i++) {
 echo $size;
 ?>
 </div>
-<div><span class="key">Memory:</span>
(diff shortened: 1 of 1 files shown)


bergware:
Two things to address/correct.
1. Skip empty slots
2. Ident the text like all other entries

image

realies:
Maybe the modal window should also be wider to accommodate for the long text.

ljm42:
Hey @realies check out the "Dynamix System Information" plugin.  It adds a "more" button to the "info" popup that gives this detail and quite a bit more.

bergware:
What is the significance of the third entry? When removed it will fit the window.

realies:
@ljm42 thought the system information modal can benefit from some aggregated memory info.
@bergware f50650a adds `Skip empty slots, Ident the text like all other entries`. Row width depends on the string lengths per line. Some UI magix may guarantee they are not going to overflow.

PR #419, opened 2018-12-27
Position context menu always left + below icon

bergware:


Squidly271:
Problem with this on the docker tab is that the last 1-2 containers will require you to scroll down the display to see / be able to click on the entries in the pop up

bergware:
The default setting is "auto" for both horizontal and vertical alignment. This may cause the menu to be partially hidden at the top.
Forcing it to be displayed below the icon, makes it always available, albeit scrolling may be needed.
Unavoidable on lower screen resolutions, I think.

Do you think tooltipster does a better job in these kind of situations? It is not a straightforward switch from context.js which is used now.

Squidly271:
CA uses it up the ying-yang, and it's never put anything off the screen (so long as the screen is large enough to display the popup)

PR #420, opened 2018-12-28
Fix/Update default-white template for smaller screens

moekhalil:
This @media rule's min-width requires a minimum width that is not necessary since the website renders fine at 940px wide without any horizontal scrolling.  
  
With the min-width rule @ 1280px wide, I have monitors that don't display the table properly and also, some of us like to have unRaid open in a small window alongside other apps.  

Before Change  
  

After Change


limetech:
Does similar change have to be made in the other 3 templates?

moekhalil:
@limetech I went ahead and lowered the minimum widths on the other templates as well, and tested.

I could do more work around the tables css rules, since it could use some refactoring, but I'll leave that for when I have more time. CSS rules like this scream "refactor me!":

```table.disk_status tr>td+td+td+td{text-align:right;padding-left:0;padding-right:12px}```

bergware:
Making the width smaller than 1280 pixels would not allow all items to be displayed in the menu.
There are a number of optional items which can be added to menu to take into account when setting a minimum width. 940px is diffinitely too small.

PR #433, opened 2019-01-12
Fix custom parity check schedule

Squidly271:
Basically a revision of a commit in #380.  Obviously can't test all of the various combinations.  But it should work as there were no complaints before, and it doesn't work now (https://forums.unraid.net/bug-reports/stable-releases/scheduled-parity-check-did-not-run-r339/)

Would appreciate someone else checking also.

### DIFF
--- a/plugins/dynamix/include/update.parity.php
+++ b/plugins/dynamix/include/update.parity.php
@@ -25,12 +25,29 @@ if (isset($_POST['#apply'])) {
     $write = $_POST['write'] ?? '';
     $term  = '';
     switch ($dotm) {
-      case '28-31': $term = '[[ $(date +%e -d +1day) -eq 1 ]] && '; break;
-      case 'W1'   : $dotm = '1-7'; break;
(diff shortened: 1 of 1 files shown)


bergware:
Ok, didn't see that report.
I suppose I relied too much on cron ;(

And specially for @Squidly271 : 
Notepad++ -> Settings -> Preferences -> Language -> Tab size = 2  + Replace by space

Squidly271:
I know about that.  Thing is that I just can't stand spaces.  If/when CA ever gets integrated, then at least everyone will have to conform to me

Squidly271:
> Ok, didn't see that report.
You replied to it and I quoted you lol

> Notepad++ -> Settings -> Preferences -> Language -> Tab size = 2 + Replace by space
Just a pet peeve.  Personally I can't stand spaces

Squidly271:
> Ok, didn't see that report.

You replied to it, and I quoted you lol

PR #434, opened 2019-01-13
Miscellaneous updates

bergware:


Squidly271:
Can we also change the default settings to log options to be 50M and 1 file?

bergware:
The 10m option is only a safety fallback, normally the value is set under docker settings.

PR #437, opened 2019-01-16
Escape quotes on a container's template

Squidly271:
Escape double quotes on any config entry's _value_ Should handle https://forums.unraid.net/bug-reports/stable-releases/ver-666-quotes-are-not-parsed-correctly-on-docker-configs-r346/

Just need someone else to double check.


bergware:
Your PR seems to undo a change from an earlier PR of mine.

$cfg = parse_ini_file('boot/config/docker.cfg');

Is missing (see files changed details)

Squidly271:
Good catch.  Don't know how I messed up that one.

PR #451, opened 2019-01-25
telegram notification agent bug fixes

realies:
verbose script execution, fix variable assignment, exit upon false chatid retrieval, replace cat in command substitution, mute curl

### DIFF
--- a/plugins/dynamix/include/NotificationAgents.xml
+++ b/plugins/dynamix/include/NotificationAgents.xml
@@ -227,9 +227,10 @@
       ############
       {0}
       ############
+      set -x
 
       if [ ! -f /usr/local/bin/jq ]; then
-        curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o /usr/local/bin/jq;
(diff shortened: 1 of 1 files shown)


limetech:
Taking a closer look at this... I don't like how this will download an executable from github, and then use the 'go' file to install on the fly each boot, and then invoke that executable as part of notification processing.  This is fine for a plugin, provided user is aware, but for stock OS, this is a security issue.

Anyway I will merge this but I'm going install 'jq' into the bzroot and delete lines 232-239.

realies:
@limetech, I could perhaps look into using regex to parse the `.result | last .message .chat .id` from the getUpdates API endpoint so `jq` is not a dependency. How come it would be a security issue? If someone compromises the `jq` repository?

limetech:
I already added both 'jq' and 'oniguruma' (dependency) packages.  The 'jq' tool may be useful for other things so let's leave it for now.

Right, if someone compromises 'jq' repo or introduces some other bug, then something that worked before now suddenly doesn't (or worse).  Also don't like putting these kinds of dependencies on flash and especially not auto-modifying the 'go' file - the 'go' file is supposed to be under complete control of the user.  You had no way of knowing this so no worries.  Thanks for adding this support.

PR #453, opened 2019-01-30
Docker page: single column for CPU/Memory load

bergware:


bergware:
Good point. I changed to memory usage instead.

PR #454, opened 2019-01-31
Docker: use memory usage in advanced view

bergware:


bergware:
Replaced memory percentage with memory usage information. As recommended by @Squidly271 (good idea).

Squidly271:
> Replaced memory percentage with memory usage information. As recommended by @Squidly271 (good idea).
Much better.  TY

PR #456, opened 2019-02-04
Added syslog server functionality

bergware:


limetech:
If you let user specify /mnt/user/system (or anywhere in /mnt) and array not Started, then this will go to RAM - how to avoid this?

PR #461, opened 2019-02-10
This code checks for a theme-specific JS file

webprogrammingsolutions:
And it includes it in the page render if it exists.

Gives theme author opportunity to enhance UI//UX.

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -624,5 +624,14 @@ $(function() {
   watchdog.start();
 });
 </script>
+
+<?
+  $themescript = "/webGui/javascript/dynamix-{$display['theme']}.js";
+	$themescript_file = $docroot.$themescript;
+	if (file_exists($themescript_file)) {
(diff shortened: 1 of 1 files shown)


bergware:
Any example where this is used?

webprogrammingsolutions:
Not yet. Working on a new Material Design theme and would be nice to have a re-sizeable sidebar which would require some custom jQuery.

bergware:
There exist already a possiblity to insert (conditional) javascript code.
You need to create a page file, e.g. /plugins/myplugin/MoreJS.page
The content of MoreJS.page looks like this

Menu="Buttons"
Cond="$display['theme']=='azure'"
---
<script>
...
...
</script>

webprogrammingsolutions:
ok, I can work with that. The code I suggested may still be helpful down the road. Do you have a roadmap for anything theme related in unRAID? Or are things staying as-is for the forseeable future?

bergware:
I don't see the core framework of the GUI changing in the forseeable future.
E.g. the function I was referring to, exists since the introduction of Unraid v6.

PR #465, opened 2019-02-23
Use cookie for display setting font size

Squidly271:
To allow multiple devices to have a different setting


bergware:
Better change
`$display['font'] = $_COOKIE['fontSize'];`

to
`$font = $_COOKIE['fontSize'] ?? $display['font'];`

This would allow systems which have a setting already but no cookie to retain the original setting.
Ps. You'll need to adapt the later IF statement as well to use $font

bergware:
I made a small test, but cookies locally created using javascript are not stored in the PHP $_COOKIE variable.

Squidly271:
Changed to $display['font'] = $_COOKIE['fontSize'] ?? $display['font'];

Has to modify $display just in case dynamix / plugin ever checks that variable to see what the current size actually is, not what is saved.

And, this PR does work.  Tested on multiple devices.  $_COOKIE works so long as it comes before the HTML

bergware:
Ok :)

Squidly271:
lol  Does that mean you tried it or are you just taking my word for it?  I'd prefer you actually try it.

Squidly271:
There is one minor issue with the cookie though.  If you change the domain, the display setting will revert to the last saved value.  But, since we don't have a mobile interface and people routinely (at least I do) bounce between tablets / phones / desktops to manage the server depending upon where they are the pro's outweigh the cons

bergware:
> lol Does that mean you tried it or are you just taking my word for it? I'd prefer you actually try it.

First I took your word and then I tested it 👍

PR #476, opened 2019-03-24
Docker: Add More Info link (docker registry) to context menus

Squidly271:
Because more and more maintainers don't include in the support thread enough info to configure the container, but rather include on the registry page

### DIFF
--- a/plugins/dynamix.docker.manager/include/DockerContainers.php
+++ b/plugins/dynamix.docker.manager/include/DockerContainers.php
@@ -57,7 +57,8 @@ foreach ($containers as $ct) {
   $webGui = html_entity_decode($info['url']);
   $support = html_entity_decode($info['Support']);
   $project = html_entity_decode($info['Project']);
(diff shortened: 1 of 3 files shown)


bergware:
There is already a key "Registry" (notice the capital R).
Adding the key 'registry' is going to give confusion. Can you change the name?

Squidly271:
It already grabs the existing key.  As for the mixed case, was simply following the same style as the previous entries

Squidly271:
nm  I see what you're saying...

bergware:
The template has already the field "Registry"

`<?xml version="1.0"?>
<Container version="2">
  <Name>cAdvisor</Name>
  <Repository>google/cadvisor:latest</Repository>
  <Registry>https://registry.hub.docker.com/u/google/cadvisor/</Registry>
  <Network>br1</Network>
`
Adding another field "registry" is confusing, or are you referring to the existing field?

Squidly271:
I'm not adding any field.  Using the existing one.  The "r" vs "R" comes from this line in DockerClient` $tmp['registry'] = $tmp['registry'] ?? $this->getTemplateValue($image, 'Registry');`

And I'm not about to modify the existing code to keep Registry vs registry consistent, when other lines may reference registry

PR #484, opened 2019-04-15
Add Elapsed Time and increments columns to Parity History

itimpi:
Adds two new columns to the Parity History report (Elapsed and Increments).   On the default Unraid system these will be displayed as 'Unknown'    If the Parity Check Tuning plugin is installed then there will be values displayed in these new columns when relevant.

### DIFF
--- a/plugins/dynamix/include/ParityHistory.php
+++ b/plugins/dynamix/include/ParityHistory.php
@@ -37,16 +37,17 @@ function his_duration($time) {
 <link type="text/css" rel="stylesheet" href="<?autov("/webGui/styles/default-popup.css")?>">
 </head>
 <body>
-<table class='share_status'><thead><tr><td>Date</td><td>Duration</td><td>Speed</td><td>Status</td><td>Errors</td></tr></thead><tbody>
+<table class='share_status'><thead><tr><td>Date</td><td>Duration<td>Elapsed</td><td>Speed</td><td>Status</td><td>Errors</td>
+<td>Increments</td></tr></thead><tbody>
 <?
 $log = '/boot/config/parity-checks.log'; $list = [];
 if (file_exists($log)) {
   $handle = fopen($log, 'r');
   while (($line = fgets($handle)) !== false) {
-    list($date,$duration,$speed,$status,$error) = explode('|',$line);
+    list($date,$duration,$speed,$status,$error,$elapsed,$increments) = explode('|',$line);
     if ($speed==0) $speed = 'Unavailable';
     $date = str_replace(' ',', ',strtr(str_replace('  ',' 0',$date),$month));
-    if ($duration>0||$status<>0) $list[] = "<tr><td>$date</td><td>".his_duration($duration)."</td><td>$speed</td><td>".($status==0?'OK':($status==-4?'Canceled':$status))."</td><td>$error</td></tr>";
+    if ($duration>0||$status<>0) $list[] = "<tr><td>$date</td><td>".his_duration($duration)."</td><td>".($elapsed==0?'Unknown':his_duration($elapsed))."</td><td>$speed</td><td>".($status==0?'OK':($status==-4?'Canceled':$status))."</td><td>$error</td><td>".($increments==0?'Unknown':$increments)."<td></td></tr>";
   }
   fclose($handle);
 }


bergware:
I appreciate your effort, but two reasons against this
1. A dependency on a plugin is created. Users who don't have the plugin should see no difference with the 'default' display (people will question what the "unknown" means)
2. This only shows the last result, but nothing is stored in the history, which requires adjustments in several places

itimpi:
I think the change I am suggesting makes sense even when my plugin is not installed, so in that sense there is not really a dependency on the plugin.

1.   The current history is invalid in that it shows the wrong results for the duration and speed if a parity check is ever paused and resumed (whether by the user or automatically).    The meaning of Unkown in the extra columns is exactly what the situation is now in Unraid with the current implementation of pause/resume as it does not track those values.  That means that showing Unknown in the extra columns is accurate as far as the current Unraid implementation of parity checks with pause-resume capability is concerned regardless of whether the plugin is present or not.   The plugin is just filling in the correct values when it is installed until (or if) the core Unraid system can be enhanced to do it directly.

2.  I do not understand this point.    I have not created any extra entries or removed any above what existing code already creates.    I could not find anywhere else in the code that ends up being affected by the change I suggested so I am not sure what adjustments you are talking about.    If you mean that Unraid does not correctly record the values in the existing entries it makes, then yes I agree this will take significant work on core Unraid code to correct.   Are you therefore saying that I should raise a defect report about the fact the current history entries have bogus/misleading values?

It would be relatively trivial for me to add a “History” button to the plugin’s setting page that displays the history in the manner my suggested change would do.   It just seemed sensible to get the built-in support for displaying history to be more correct.

bergware:
Sorry my view is a little different.

Adding those two columns which are filled-in with "unknown" at the moment do not bring extra information for the user but will raise the question why it isn't filled in properly.

It is true that pausing and resuming the parity check results in wrong duration and average speed values. This needs to be addressed, but requires deeper coding on emhttp level since the GUI can only read the results.

The 'monitor' function which runs in the background updates the history log when it detects a new entry. This update does not include any of your "new" variables.
There is also a helper function used in the Array Operations page to read the "current" or "last" parity results and displayed accordingly, again not including your new variables.
All this results in wrong reading and writing because column assignments have changed in your proposal.

It seems for the moment the safest approach is to make this a separate item in your plugin. Maybe call it a bit different, such as "More History" or whatever you prefer to indicate it is not the same as the standard history.

itimpi:
OK.   I have an alternative proposal then for a different pull request:

1.  Update the ParityHistory.php file so that those extra columns are only displayed if at least one entry in the parity-checks.log file contains the extra information.  Hopefully that will handle the objection of the fields being displayed when they will not mean anything to the end-user as  it will mean that if my plugin is not installed the results will look identical to the current version.
2. The Monitor function does not cause a problem.  If it adds an entry without the extra fields then they get displayed as 'Unknown' by my changes to ParityHistory.php.   When the parity-check finally completes my Plugin updates the entry added by the monitor to correct the Duration and Speed fields to cover the whole operation rather than just the last increment and adds the Elapsed time and increments field as additional fields on the end of the record.  Because it leaves the timestamp information alone Monitor does not try and add an additional entry.
3. Add the changes to the ArrayOperations page and its associated helper to display the status information correctly alongside the History button for the current/last operation when my plugin is installed.   I had avoided doing that as it seemed a bigger change but I am happy to submit the changes needed.  It is not possible to correct it if the plugin is not installed because as you say that requires core level changes.   You also mentioned wrong reading/writing due to column alignment issues but this does not seem to happen as I did not change the order of the fields in the history file but only added some additional ones on the end of each record that get ignored by the current code.   If at a later time the core system starts adding these extra fields they will then get picked up if my changes go ahead.   If this ends up being handled in a different way then at that point I can simple remove from the plugin the tracking of history and rely on the new system level implementation.

If you think the above would be acceptable then I can work on submitting an alternative pull request that addresses the points you have raised.

As an interim I have the plugin install a patched version of the ParityHistory.php file as the extra columns will be meaningful to anyone who has installed the plugin.   I also added a History button to the plugin's settings page that displays the history in the extended format.   It does leave the fact that the status information alongside the Main tab History button only applies to the last increment as an outstanding issue but the History that is displayed on clicking the History button DOES display the correct information for the full check.

At the moment the plugin is patching the existing parity-checks.log file with updated/extra information.   If there are deep objections to this I can change it to instead only store this additional information in its private folder on the flash drive (which is what my early implementation did).   However it did seem that patching the existing history file had benefits as it corrected for deficiencies in the current Unraid code.

limetech:
It's pretty easy for me to add 'duration' and 'increments' to data maintained by driver.  This might be better anyway if/when pause/resume across reboots is supported.  Would this help?

itimpi:
In this context I assume you mean 'elapsed time' when you say duration?  Maybe a new term/phrase needs to be introduced to handle the execution time rather than duration?  Without pause/resume these have been synonymous so it has not really mattered.  In my plugin I used 'Elapsed Time' as I wanted to leave duration as meaning execution time but maybe it is time to use a new term and deprecate duration as meaning execution time?

I think it would definitely help the plugin help as if this information is made available via the /proc/mdstat fields as then I could remove the code from my plugin that tries to track this within the plugin.    It would also mean that the values were accurate for manual pause/resume which currently are harder for my plugin to detect.   It would also make it easy for the existing monitor script to be modified to capture and save this information.   The plugin would then end up being simplified o only be responsible for automated pause/resume functionality with all its current history tracking code removed.

The other point is to get agreement as to whether this additional information should be recorded for historical reasons as additional fields at the end of the records in the existing parity-checks.log file (in the manner my plugin is currently saving this information).   I chose that approach as it maintains maximum compatibility with existing historical records.   It would also mean that  any such information currently being gathered and recorded by the plugin would remain valid when the built-in implementation arrives.   If the information is going to be saved in a different manner then I need to amend my plugin to stop adding it to existing records ASAP to avoid causing problems in the future.

If the change to the core code goes ahead I would be happy to contributing my time/effort to making the full set of changes to the GUI code required to support this.   I would also be happy if you decided you needed to keep that work 'in house' as long as I knew that was happening and could plan changes to my plugin appropriately.

itimpi:
Had a thought - do you want the Elapsed time and number of increments columns to be shown in the Parity History if the user has never made use of the pause/resume capability?    I have already worked out code in my customised version of the ParityHistory.php script that avoids displaying those columns if the user has no entries with Elapsed Time or Increments set in their parity history.

limetech:
What I mean by 'duration' is the total time spent doing a Resync.  If there are no pauses, this would be the same as elapsed time from starting the operation, to when operation completes (or is terminated).  If there are pauses, it would not count the time in the paused state.

For history, all you care about really is duration, right?  Does it matter to keep track of elapsed time and number of segments (which I presume is number of times operation was paused + 1)?

I don't think it's worth the extra effort to do more than that.

itimpi:
I think that both the total elapsed time and the execution time (when the check is actively running) are of interest.

My plugin is tracking both the execution time and the elapsed time.     The execution time is then used to calculate the speed while the parity-Check is actively running.   I think the average user will be more interested in that figure as it gives an idea of how well the system is handling the parity check.    Also it means the speed is not dependent on whether you run the check in small or large increments as the total execution time should normally be relatively consistent.    The execution time can also be used to decide on how many increments and increment length you want to use.

Whether the number of increments is required is perhaps debatable.   As you say it is equivalent to pauses+1.  Since my plugin is storing pause/resume times during a check to enable the elapsed time and execution time to be derived, counting the increments was trivial and I found it to be of interest when looking at parity history.   However that may just be me as in the case where you have configured for one increment per day this then becomes the number of days the parity-Check takes to complete which can also be inferred by the elapsed time.   The increment count might become of more interest if you are running on a more complicated schedule than one per day.    I was thinking that one day it might be possible to auto-detect system activity and using this to pause/resume parity checks.   In such a case the conversion between elapsed time and increments is not inferable so measuring it directly might be desirable and is easy to implement.   Note that this assumes that the parity-history only has one record per check  (which is current implementation) - not separate records per increment.   One record per check seems the sensible approach to me

BTW:  while I think about it (you may already have done this) it might be a good idea to have an exit code of zero (or a new value) on a successful pause rather than the -4 that is in rc7.

limetech:
> I was thinking that one day it might be possible to auto-detect system activity and using this to pause/resume parity checks.

Item on my 'todo' list forever has been to implement a throttling mechanism where Resync would be less aggressive if I/O was taking place, if it simply paused, a Resync operation might never finish :)

re: stored data: I think it's important to minimize plugin-specific vars stored on the usb boot flash.  Instead, since super.dat is already managed there, it's better to save info in there, especially if we ant to pause/resume across reboots.  Hence in your design it might be better to specify what you want to see added.  I already mentioned 'duration', if you think 'elapsed time' and 'pause count' is important, I can probably add that info pretty easily.

> it might be a good idea to have an exit code of zero (or a new value) on a successful pause rather than the -4 that is in rc7

Sure.

itimpi:
If I have execution time (= duration?), elapsed time and increment count then the plugin would have no reason to store any dynamic data other than its settings on the USB drive.

limetech:
Here's an update on this: I have a series of md/unraid driver changes coming in 6.8 which will permit pause/resume across reboots and provide better reporting of total time spent during 'sync'.  This will require changing a handful of both 'md' and 'sb' variables.  Therefore I'd like to request that you retract this particular pull request for 6.7 release.

Squidly271:
It's really annoying my OCD having this still open

PR #486, opened 2019-04-28
Add GameServers to category for docker containers

Squidly271:


limetech:
We should make the category list part of the feed somehow.

Squidly271:
Easy enough to do.   Changes aren't that common, but your idea does make sense.

PR #494, opened 2019-06-26
Modified notify script to allow overriding email recipients in notification settings

Influencer:
This allows for a little more flexibility if users or plugins want to tie into the systems notification system. If there are one off notifications or notifications that only one person would be interested in, a script can pass in -r "myemail@email.com" and the notify system will ignore the recipients in the notification settings, instead sending to the provided email argument.

See this [post](https://forums.unraid.net/topic/81268-modified-notify-system/) for more information on the reason behind the modification.

Also important to note that these changes do not affect default functionality at all and require no updates to anything using the notification system currently.


limetech:
Nice feature, thank you!

PR #496, opened 2019-07-14
download docker icon with every edit

Squidly271:
Fix bug with dockerMan that's been present forever where it will not allow you to simply change the icon URL and have the system update the icon.

### DIFF
--- a/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -137,7 +137,15 @@ if (isset($_POST['contName'])) {
   }
   if ($startContainer) $cmd = str_replace('/docker create ', '/docker run -d ', $cmd);
   execCommand($cmd);
- echo '<div style="text-align:center"><button type="button" onclick="done()">Done</button></div><br>';
+  // force re-download of the icon
+  if (! strpos($Repository,":")) {
+    $Repository .= ":latest";
+  }
+  $iconPath = $DockerTemplates->getIcon($Repository);
+  @unlink("$docroot/$iconPath");
+  @unlink("{$dockerManPaths['images-usb']}/".basename($iconPath));
+	
+  echo '<div style="text-align:center"><button type="button" onclick="done()">Done</button></div><br>';
   goto END;
 }
 
@@ -156,6 +164,7 @@ if ($_GET['updateContainer']){
       continue;
     }
     $xml = file_get_contents($tmpl);
+
     list($cmd, $Name, $Repository) = xmlToCommand($tmpl);
     $Registry = getXmlVal($xml, "Registry");
     $oldImageID = $DockerClient->getImageID($Repository);


bergware:
At the time I removed the unnecessary icon download when a container gets updated, this prevents an "empty" icon when something goes wrong during the update process and speeds up the process altogether.

What is your fix exactly doing?

Squidly271:
Because if a user decides to change the icon URL, if the icon has already been downloaded, the OS as it stands now will not download a replacement to reflect the new URL.  A better solution is to determine if the URL has actually changed from the original, but this was a simple fix, and determining the difference was simply more work than I was willing to invest for an issue that only happens on occasion

Personally, though I would prefer that the icons are never downloaded and instead are grabbed live from whatever URL the user dictates.  Never could quite understand why unRaid decided to cache the icons locally for dockerMan, instead of letting the browser do what it does best.

limetech:
> Personally, though I would prefer that the icons are never downloaded and instead are grabbed live from whatever URL the user dictates. Never could quite understand why unRaid decided to cache the icons locally for dockerMan, instead of letting the browser do what it does best.

I guess it was done that way to handle a case where server does not have internet access, with side benefit of not unnecessarily writing the flash.  To the latter point, we've been wanting to migrate /boot/config/plugins/dockerMan to /mnt/user/system/docker for some time now.

Is it possible to re-download icon only as part of 'apply update'?

Squidly271:
Yeah, I just haven't gotten around to it yet to check if the URL has actually changed.

limetech:
Could be that url stays the same but icon still changes?

Squidly271:
Sure, of course.  PR as it stands now would handle that by redownloading on every edit of the container.  But since nobody likes this PR as is, I'll get around to checking for a changed URL and then the user can deal with the case of when the app changes it's icon.

limetech:
Instead of checking on every edit, check when it checks for a container update?  Maybe I'm missing something...

Squidly271:
In my mind, this isn't about if Plex decides to change their icon (and keep the URL the same - although this does handle that at edit time).  It's all about if the user wants to change the icon they use by supplying a different URL.  And that is only going to happen at edit time.

And if we're trying to reduce flash wear and tear, edits happen far less often than updates.

This is all because currently its a pita to change the icon because its non-intuitive what you have to do.

bergware:
How often does an icon change?
Most of the time when a container gets updated, the icon doesn't change and hence it isn't needed to download the icon.
If/when an icon changes, its URL and name usually stays the same.

Squidly271:
> How often does an icon change?
Rarely

> If/when an icon changes, its URL and name usually stays the same.
Unless the user wants to change it.

How often are containers edited?

Rarely.

I did this PR not because I cared personally, but because some guy complained and he is right.  If you change the URL, then the icon should change and that is not how the system currently works

Let's just close this until I redo the PR.

bergware:
Haven't seen the complaint, but will await your next debate :)

PR #498, opened 2019-08-04
Allow redefineable showUpgrade / hideUpgrade functions

Squidly271:
May seem like a rather strange PR, but CA has been working towards redefining the entire banner notification system (and opening it up to 3rd party apps, including CA).   

The hold-back on this currently is that on the Apps page which is making extensive use of the banner notification bar, any OS update notification is lost and won't be displayed. 

Allow CA to redefine those 2 functions to allow CA (and other 3rd Party Apps) to also utilize the banner without losing the OS upgrade notification.

Too see my vision, disable the docker service AND cookies in your browser and go to the apps tab. (ideally have CA up to date also and watch it for 20 seconds+)

See also this  https://forums.unraid.net/topic/79010-ca-api-for-plugin-update-checks/ for rev 1 of what CA's officially released and documented

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -210,7 +210,7 @@ function showFooter(data, id) {
 function showNotice(data) {
   $('#user-notice').html(data.replace(/<a>(.*)<\/a>/,"<a href='/Plugins'>$1</a>"));
 }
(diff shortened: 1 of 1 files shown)


Squidly271:
Closing for now.  Will instead work on integrating CA's banner warning system into unRaid proper as there are inconsistencies present (ie: OS update available and system notifications disabled cannot be displayed at the same time)

PR #500, opened 2019-08-11
Revamp Banner Warning System

Squidly271:
A number of advantages

- From my point of view, the most important is that it opens up the banner warning to 3rd party plugins, without CA's function having to create its banners with a z-index greater than the OS update z-index
- Allows multiple banner warnings ( browser will cycle through them ) (Previously, System Notifications Disabled couldn't be displayed if an OS or Reboot Warning was present ) (If desired, ACS Override can now via suitable programming toss up a reboot now banner instead of only appearing on its own page)
- If user dismisses an OS update, if another becomes available, then the banner will re-appear unlike previous operation if the browser window was not closed.
- Any given warning is able to be dismissed without any programming
- Warnings are able to be removed by the OS / plugin itself if the circumstances that result in it being displayed are no longer present

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -220,59 +220,59 @@ var bannerWarningInterval = false;
 var osUpgradeWarning = false;
 
 function addBannerWarning(text,warning=true,noDismiss=false) {
-	var cookieText = text.replace(/[^a-z0-9]/gi,'');
-	if ( $.cookie(cookieText) == "true" ) { return false; }
-	
-	if ( warning ) {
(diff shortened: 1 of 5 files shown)


limetech:
This looks good to me, let's ask the gui-guru: Ronald: looks good to you too?

Squidly271:
Easy tests are make sure CA is up to date, force the system to show an OS upgrade, disable docker, disallow cookies in browser and then hit Apps

Squidly271:
Easy tests are make sure CA is up to date, force the system to show an OS upgrade, disable docker, disallow cookies in browser and then hit Apps

eschultz:
@bergware looks good to you too?

bergware:
I did some tests, it looks good.
Thanks Squid.

PR #506, opened 2019-08-20
Made notify script compatible with 6.8 new security scheme

bergware:


limetech:
Thanks for finding this one.

PR #510, opened 2019-08-27
Added Dokcer/VM XML files to diagnostics

bergware:


bergware:
Are these advanced fields (available under "more settings" in the GUI) or regular fields?

One option is to anonymize any advanced fields.

bergware:
Yeah, current status quo about docker info seems the best (easiest) approach

I am not using any MACs.
Any example of this "not-so-secret" entry?

bergware:
Docker container state fix, courtesy of @bluemonster.

PR #512, opened 2019-09-02
enable group chat IDs, update helper description

realies:
A discussion in the unRAID Telegram group has revealed that adding the notifications receiving Telegram bot to a group chat does not work as group chat IDs start with a negative sign and the current regex does not account for that, this PR should fix it.

There's also the suggestion from a user that we expose the chat ID as a text field to the notification agent settings, although I am having the concern that this might confuse users that are not that familiar with the Telegram API and am leaning towards the automatic chat ID fetching. Thoughts?

### DIFF
--- a/plugins/dynamix/include/NotificationAgents.xml
+++ b/plugins/dynamix/include/NotificationAgents.xml
@@ -215,9 +215,10 @@
     <Variables>
       <Variable Help="[a href='https://telegram.me/botfather' target='_blank'][u]BotFather[/u][/a] is the one bot to rule them all.[br][br]
         1. Make a bot using BotFather[br]
-        2. Paste bot token in this field[br]
-        3. Select bot via Telegram and press /start[br]
-        4. Test bot (first test will be slow)[/a]" Desc="Bot Access Token" Default="">BOT_TOKEN</Variable>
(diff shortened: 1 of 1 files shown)


realies:
@bergware, thanks for adding the code for updating the Telegram config location. Wondering if `/boot/config/plugins/dynamix/notifications/agents/Telegram.sh` would be reset on system reboot?

bergware:
Scripts should get updated when disabling and re-enabling the agent

realies:
> Scripts should get updated when disabling and re-enabling the agent

@limetech, would be great if the update log lets the users know that re-setting the agent adds the new feature/fix

Wondering if we can do any additional modifications that would not require the user to use the terminal to reset their cached `chatid` value. Do you think that exposing the value in an input field would over-complicate the configuration?

Tooa:
> Wondering if we can do any additional modifications that would not require the user to use the terminal to reset their cached `chatid` value. Do you think that exposing the value in an input field would over-complicate the configuration?

I would love to see the `chatid` in the GUI actually. I stumbled across this feature request while searching for a way to receive notifications to a Telegram channel. As default behavior, the `chatId` could be populated automatically.

PR #520, opened 2019-09-22
dockerMan: Redownload Icon if URL changes

Squidly271:
Continuing on my debate with @bergware from #496 .  This PR will only re-download the icon if the URL changes for the Icon

Additionally, also moved the storage location from the USB to within the docker image to further lessen writes to the flash


bergware:
To avoid double slashes (//), can you change

filemtime("$docroot/{$info['icon']}")

into

filemtime("{$docroot}{$info['icon']}")

Squidly271:
I don't see that

``

Squidly271:
Thanks.  Didn't affect functionality, but did miss that.

limetech:
I'm not crazy about storing unraid-specific information in the docker vdisk image.  This is because it's better to treat that as opaque so as not to pollute with anything non-docker, and also so that one can delete entirely and just recreate by re-downloading images.

It's been a laundry list item of ours for a long time now to get all docker container-related config stuff off the flash an into /mnt/user/system/docker instead.  Just haven't had a great appetite to do so yet.

Anyway for these icons, how about we store them in /mnt/user/system/docker/images instead?

bergware:
**/mnt/user/system/docker/images** is dangerous, because it assumes that users have/use this structure.

I, for (bad?) example, don't use this folder structure at all, and am completely diffrently organized.

limetech:
I bet you're one of those people who don't let windows install to default 'program files' or use default 'my documents' folder too  - haha

In your case then, the icons should maybe stay on the flash but seriously, why not use a 'system' share?

bergware:
I don't bet with you ;) Indeed I install applications in my "own" folder C:\Tools and moved 'my documents' to a separate disk.

In "my case" it is all history. I created my own structure before the default paths were introduced in Unraid. And ever since sticked to it. Yes, this means no "system" folder for me :)

limetech:
That's alright I like non-conformists!

Maybe we can introduce a config setting that says where to look for docker stuff.  By default for new installs it would be:
/mnt/user/system/docker
but you can set to
/boot/config/plugins/dockerMan

What I want to do is move everything from that dockerMan directory, which includes the templates, images, etc., to system/docker.

bergware:
but wait ....

Having the templates on the USB stick is super convenient. Don't you think?
It would allow easy recovery/re-installation of containers in case something goes amiss with the array/cache

limetech:
Makes it hard if something goes amiss with the usb flash device.

The flash device is only supposed to store config data which is needed when array/pools are offline.  The docker stuff is really 'application data' and should be kept off the flash.  Recall we used to have libvirt VM xml and other config on the flash too but we moved that also to system share (system/libvirt) a couple years ago.  We should have migrated docker config back then.

We can easily add "DOCKER_CONFIG_PATH" to /boot/config/docker.cfg.

limetech:
So my vote for this PR is to restore the image path to /boot/config/dockerMan/images and then we can address moving dockerMan in a separate PR, which would probably include updates to unRAIDServer.plg to either set DOCKER_CONFIG_PATH appropriately and/or move dockerMan/* to system/docker.

bergware:
For bad USB flash devices, we invented the flash backup feature ;)
But you are right, there is inconsistency between VMs and Docker file placements.
I support your vote! Waiting for Squid ...

limetech:
Yes flash backup will solve most of these problems but not everyone is going to use it and also forget to backup their flash.  Then we have to spend a lot of time in support to help them recover and they get irritated when discovering their docker templates are lost.

Hey while we're at it, notice in the docker image this file:
unraid-update-status.json
Can we also move to flash?  (that is, get it out of the image)

limetech:
Also, to add to flash backup idea, the templates could be considered 'private data' and for that reason we don't really want them included in flash backup to unraid.net.

bergware:
Everything can be moved, but that file is in the same location as "unraid-autostart" because they both hold specific docker info. Why not move both? It allows to retain this information after the image is recreated (convenience)

bergware:
_Also, to add to flash backup idea, the templates could be considered 'private data' and for that reason we don't really want them included in flash backup to unraid.net._

What, you received another complaint from user BennTech ;)

limetech:
> Everything can be moved, but that file is in the same location as "unraid-autostart" because they both hold specific docker info. Why not move both? It allows to retain this information after the image is recreated (convenience)

Yes agreed, I forgot about that file because I don't have autostart for containers on my test servers.

> What, you received another complaint from user BennTech ;)

Anticipating ;)

Squidly271:
> I'm not crazy about storing unraid-specific information in the docker vdisk image. This is because it's better to treat that as opaque so as not to pollute with anything non-docker, and also so that one can delete entirely and just recreate by re-downloading images.

I don't see the issue with storing the icons within the docker.img, because other unRaid specific files are already there (as noted), and upon a recreate of the image file (and re-download of the container), the icon will simply download again.  No big deal.  Until such time as unRaid is ready for a complete revamp and everything winds up going to /mnt/user/system, I think that all users would appreciate any lessening of writes to the flash.

> Hey while we're at it, notice in the docker image this file:
> unraid-update-status.json
> Can we also move to flash? (that is, get it out of the image)

Then that becomes a pointless write to the flash drive.  It has nothing to do with configuration information at all and is simply a cache for the update status of any given container.  

> Everything can be moved, but that file is in the same location as "unraid-autostart" because they both hold specific docker info. Why not move both? It allows to retain this information after the image is recreated (convenience)

Sure, the argument can be made that this is a configuration setting and should be on the flash drive.

> Also, to add to flash backup idea, the templates could be considered 'private data' and for that reason we don't really want them included in flash backup to unraid.net.

Isn't unraid.net secure?




> But you are right, there is inconsistency between VMs and Docker file placements.

Yeah, slightly different history between the two of them that resulted in that.  But, anecdotally there are less problems with how dockerMan stores the files vs vmMan.  Not quite so easy to get a single XML out of libvirt.img  And, if somehow that single file becomes corrupted, you've now complicated how to recover all of your VMs, not just a single one


TLDR, my vote is to move the icons into the image file.

ljm42:
The icons are just cached files, easily re-downloaded if lost. It isn't important config data, so there is no value in storing it on the flash. This just results in unnecessary flash reads and writes, and larger than necessary flash backups (which is bad for Unraid.net, right?)

The main reason we have docker.img is so you can delete and start over if there is corruption. Storing icons there would not hurt that since the system could easily re-download the icons when you start over.

What I really think we need though is a guaranteed system share for storing information after the array has started. I don't see the point in complicating everything by making a bunch of config settings when we could standardize on /mnt/user/system/

If that is a long way out then I think putting icons in docker.img now is a step in the right direction.

But please don't move unraid-update-status.json to the flash. That is not important config information that warrants constant updates to the flash drive. It would be a great candidate for a system share though :)

Squidly271:
> This is because it's better to treat that as opaque so as not to pollute with anything non-docker, and also so that one can delete entirely and just recreate by re-downloading images.

To elaborate on my previous comment, is your intent to have the docker.img file only store data that's directly involved in the actual docker system?  My opinion is that anything that's not config related, but 100% related to the containers running should be in there, especially if its automatically created at docker create time.  Simplifies everything, lessens the files that would be stored on either the flash or ultimately within /mnt/user/system.  This means that the icons are now treated as something akin to an internal data structure with no need to expose the files to the user.  And this PR does effectively do that by handling the long standing bug where dockerMan never changed the icons no matter what you did.

Beyond that, I am really surprised at the reaction that IMO a very minor change generated.

limetech:
> Beyond that, I am really surprised at the reaction that IMO a very minor change generated.

No your change is fine and I'll merge the PR.

As mentioned earlier, the clean-up of the placement of "stuff" related to both docker and libvirt has been a longstanding laundry list item.  More to the point: really tweaks my OCD :)

But making this kind of change now is way beyond scope of 6.8 and even 6.9 and your PR improves things, thank you!

PR #523, opened 2019-09-28
CSP: block-all-mixed-content

ljm42:
If a page is loaded via https, prevent it from loading resources via http


limetech:
Thank you, nice changes!

PR #525, opened 2019-09-28
Allow outside click to close popups

Squidly271:
Drives me nuts having to click the "x", as so used to CA allowing an outside click.  Only allows an outside click if when closing the window nothing happens


bergware:
This was done to prevent 'accidental' clicking and the user closing the window before actions were finished...

Squidly271:
Which is why if no action happens (ie: the page doesn't reload or anything like that) the system only will allow the outside click.

Fundamentally, this change only affects things like the system info button, plugin change logs, and the like.

If you can find another popup that allows an outside click when it shouldn't let me know and I'll withdraw / rework the popup

limetech:
I'll merge and we'll see how wider testing goes.

PR #526, opened 2019-10-02
Added 'F1' key to toggle help text

bergware:


limetech:
I like the 'F1' key shortcut but not so sure about the others, especially since may conflict with system hotkeys (such as Ctrl-P to print) or maybe user-defined hotkeys.  Even during extensive use of Main page for testing, I never thought "hey I wish there was a hotkey for doing xxx".  I think best to keep F1 for help but let's not do the others.

bergware:
Well, thought it was a nice addition, but not really essential. Reverted.

PR #527, opened 2019-10-10
Log webUI login attempts

Squidly271:
Maybe I'm just paranoid, but with ports open, a reverse proxy thru unraid.net, I wouldn't mind seeing any and all login attempts to the server


limetech:
Good idea - I'll sneak this into -rc1 release.

PR #529, opened 2019-10-14
Improve Use Cache option

itimpi:
Try and make the Use cache option for a User Share less likely to be misinterpreted by new Unraid users


bergware:
Looks good.
Should the options "no" and "only" be explained separately? E.g.
No = "Files reside on array only and don't make use of cache"
Only = "Files reside on cache only and are never moved to array"

itimpi:
I was reluctant to do that as it is not always true if the user has changed the setting so that Existing files are already in the wrong location.

limetech:
Looks good, thank you!

PR #538, opened 2019-10-23
Expanded help for Use Cache setting

itimpi:


limetech:
Thank you.  Wording of this will change again once multiple-pool feature released.

PR #548, opened 2019-11-07
Switch #nav-item to .nav-item

OmgImAlexis:
Currently we use an ID for each nav item, the issue with that is IDs should be unique per page so I've moved them all to using a class instead.

Visually this doesn't change anything but there's chance this could break things with custom themes and plugins that rely on using the `#nav-item`.

PING: @limetech

### DIFF
--- a/plugins/dynamix/HelpButton.page
+++ b/plugins/dynamix/HelpButton.page
@@ -17,7 +17,7 @@ Code="e934"
 ?>
 <script>
 function HelpButton() {
-  if ($('#nav-item.HelpButton').toggleClass('active').hasClass('active')) {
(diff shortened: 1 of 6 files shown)


bergware:
Have you tested this?
Chances are big something got broken...

OmgImAlexis:
Yes but it’s possible plugins could be effected. With a vanilla install it should be fine.

I don’t mind when this is merged.

Squidly271:
Quick search shows that dynamix system buttons would be affected, along with something else I've been very slowly working on for a future release.

OmgImAlexis:
Any chance this could be merged yet?

PR #550, opened 2019-11-10
Fixed diagnostics errors when array was never started

bergware:


limetech:
I'll merge this PR but what is the case where diags are gathered and those ini files are not present?

bergware:
When the server is started but not the array (autostart = no).

limetech:
I was wondering if there was a bug report or other report of this issue because even with autostart set to No, those ini files should still have been generated.

bergware:
I guess this is one of those edge cases, which are not really fully examined

ljm42:
Here is the bug report: https://forums.unraid.net/bug-reports/prereleases/680-rc5-command-line-diagnostics-throws-errors-r714/

limetech:
Ok the one missing is 'diskload.ini' - that makes sense.

PR #553, opened 2019-11-15
More prominent format warning

itimpi:
Users still seem to think that format can be used as part of a data recovery or disk rebuild despite the current warning in the GUI.   This is an attempt to make it even more obvious that this is not the case and harder for the user to ignore by accident.


Squidly271:
I would personally add
`type: 'warning'`
into the swal options

itimpi:
I am not at all hung up on the wording/presentation of the alert.  The main idea was whether the idea of making it an alert was deemed to be a good idea in the first place and likely to reduce the incidence of users doing a format when it was the wrong action. 

My brother thought it might be worth making an explicit reference to doing a file system check as often being the correct thing to do instead of a format.   I preferred to keep the wording more minimalist as I feel that makes it more likely the user will actually read it.

bergware:
If we go the "alert" way, I prefer **type: warning** to be added, it makes it more explicit that user attention is required.

itimpi:
OK  - I will resubmit with warning type.

PR #554, opened 2019-11-20
Add shares: slashes are not allowed in share name

bergware:


OmgImAlexis:
Unrelated but just a thought.

Since shares are auto created when a new directory is added to /mnt/user; What happens a user creates a directory with a slash?

limetech:
> Since shares are auto created when a new directory is added to /mnt/user; What happens a user creates a directory with a slash?

Can't create a directory with a forward-slash because that's a separator.  A directory containing a back-slash or only a back-slash is valid, however it's not valid as a component of a file name on FAT32 which is where the same-named cfg file is created/looked for.  This is the actual bug.

limetech:
There are other places in webGUI that need this 'fix'.

limetech:
Nice work! thx

PR #558, opened 2019-11-21
Warning alert for Format operations

itimpi:


limetech:
Thank you.  May change wording slightly but good idea.

PR #559, opened 2019-11-24
Diagnostics: Remove OSK info from VM xml

Squidly271:
Original code wasn't working.  It was attempting to open the literal file ..../*.xml instead of applying the wildcard


Squidly271:
Now doesn't work from GUI.  Closing

PR #560, opened 2019-11-24
Diagnostics: Adjust for timezone from webGUI

Squidly271:
The filename generated by the webGUI's diagnostics wouldn't take into account the user's time zone.  (Time would always be UTC/GMT)

Set the filename according to the date / time on the browser

### DIFF
--- a/plugins/dynamix/Diagnostics.page
+++ b/plugins/dynamix/Diagnostics.page
@@ -22,8 +22,9 @@ $zip = htmlspecialchars(str_replace(' ','_',strtolower($var['NAME'])));
 
 <script>
 function zipfile(){
-  var d = new Date();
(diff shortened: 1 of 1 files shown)


bergware:
Not sure what you try to accomplish, because there is nothing wrong with the current implementation.
The "new Date" function of javascript takes the local browser's time and hence the generated file name includes the local time of the user, which corresponds to the times mentioned in the logs.

Squidly271:
It does if you're in GMT.  For me in EST, the filename's time is out by 5 hours if generated by the GUI, but correct if generated by the CLI

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

> The timezone is always zero UTC offset, as denoted by the suffix "Z".

bergware:
I am in CET (UTC+1) and works as expected.

Squidly271:
Stock diagnostics:
image

My changes
image

bergware:
yep

PR #566, opened 2019-12-24
Add default to theme switch

Skitals:
This was hardcoded for four specific theme filenames. This line adds a default case so custom themes are better supported.


bergware:
Please close this PR.
I have made another approach to better support custom themes.
Reasons:
- The white theme is the default theme, not black
- This PR is specific to your needs

PR #572, opened 2020-01-11
Miscellaneous updates

bergware:

PR #573, opened 2020-01-11
autocapitalize="none"

ljm42:
Prevent mobile browsers from changing username "root" to "Root"


ljm42:
No longer needed, is included in #572

PR #576, opened 2020-01-16
Add Network:VPN as an application category

Squidly271:


Squidly271:
@bergware I put a PR in against wireguard.xml to reflect this change also

PR #584, opened 2020-02-06
Miscellaneous updates

bergware:


Squidly271:
There's a conflict between this PR and #583 the download_url routine used for icon downloads in DockerClient (Lines 71+)

bergware:
I merged your PR into mine, you can close yours.
Thanks

Squidly271:
There's also one other line you need to add  ~line 319

`	public function getIcon($Repository) {
		global $docroot, $dockerManPaths;
		$imgUrl = $this->getTemplateValue($Repository, 'Icon');
		if ( ! $imgUrl ) return "";`

the final if statement there needs to be added to your PR

Squidly271:
But, you should test the rest of the system's downloads (more repositories, etc).  I only changed the download_url routine as it was only used by the Icon system and had an issue with connections

Squidly271:
Issue with your changing plugins/dynamix.plugin.manager/include/PluginHelpers.php line 20 

The original code was 

return $retval==0 ? implode("\n", $output) : false;

whereas the new code is 

  return $retval==0 ? end($output) : false;

Why are you switching to only return the last entry of what is potentially a multiple line array?  This results in (the most obvious example) the changelogs for any given plugin only listing the last line in the <Changes> section instead of the entire changlog unless you run the bash version instead of the PHP version.

IE: CA consistently runs the PHP version of plugin, and it's messing up all the changelogs because of this change.

bergware:
Yeah, I was already afraid I did "too much" theres ;)

Made an update which restores the original code and compatibility.

limetech:
we good to merge?

bergware:
Yeah, the brain storm is over ... good to go

PR #585, opened 2020-02-08
Prevent referrals on HTML links

Squidly271:
Since Support, Project and MoreInfo links in context menus are in control of the app authors, not limetech


Squidly271:
Referrals on icon downloads are handled via #583

PR #588, opened 2020-02-09
Block referrals to 3rd Party Sites

ljm42:
This adds `<meta name="referrer" content="same-origin">` to all pages, which will prevent browsers from sending referrer information when linking offsite.  See:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

### DIFF
--- a/logging.htm
+++ b/logging.htm
@@ -3,6 +3,7 @@
 <head>
 <meta name="robots" content="noindex, nofollow">
 <meta http-equiv="Content-Security-Policy" content="block-all-mixed-content">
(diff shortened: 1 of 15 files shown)


Squidly271:
And while it may not be necessary, referrals on dockerMan icon downloads are handled via #583

bergware:
Resistance is futile, you will be assimilated :)

Squidly271:
Seeing as how many of my ideas have already been assimilated, it's fair to say I'm already part of the collective.

PR #592, opened 2020-02-14
Users: Display Password Strength

Squidly271:
Concept.  I fully expect @bergware to refine the styling


OmgImAlexis:
Doesn’t this add a false sense of security?

Squidly271:
Gives people a clue of how weak their password is.  Pretty common nowadays to have this check.

limetech:
Please translate that regex to English: what is this doing?

OmgImAlexis:
Somewhat of an issue I've had in the past with "strength meters".

```
validatePassword('Should this long secure password comes up as medium just because it has no numbers and such in it')
```

bergware:
Medium
"paid $10 dollar for this password with more than 8 characters"

Strong
"Password1#"

Squidly271:
Thinking about it overnight I believe that we should limit this to just checking against the root password

bergware:
If we are going to use this, I would make the rules for "medium" more stricter.
Just adding a uppercase letter isn't really that much more secure ... just saying!

Squidly271:
Personally I don't think that's necessary.   Nobody ever suggests to make sure that your password is at least medium.   All suggestions everywhere is to create a strong one

Squidly271:
And it's not just a capital letter.   password1 or password$ is also medium

bergware:
So why do we use medium?

Squidly271:
Because it's better than weak.   It's not like theres any rule enforced that the password has to be strong

bergware:
So "password123" is better than "password" ;)

Squidly271:
Touche

bergware:
https://youtu.be/_JNGI1dI-e8

limetech:
Amazing! That's the same combination on my luggage too!

limetech:
Keypass has a "quality meter", examples:

image

image

image

bergware:
@Squidly271 with your permission, I changed the password validation from regex to javascript "zxcvbn". From github:

zxcvbn is a password strength estimator inspired by password crackers. Through pattern matching and conservative estimation, it recognizes and weighs 30k common passwords, common names and surnames

It classifies passwords as worst, bad, weak, good and strong.

Entering "password123" as password gives

image

(for sharp eyes: I made a bar which grows with the length of the entered password and changes color depending on how "zxcvbn" rates the password strength)

Squidly271:
That works for me.   Like stated in the OP this was a concept  didnt invest much time in it, but with where unraid is going something like this is needed on the root password.   Dont believe we need to put it on any users.

Closing this PR

bergware:
I would leave that strength indicator for every user's password, to make the user more aware

ljm42:
Thanks for starting this conversation @Squidly271 , this is a really good idea.  And wow @bergware , zxcvbn is a great find

OmgImAlexis:
Since were adding this maybe we should also look at adding pwned password checking? There are a few js libs that interface with this. 

https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/

PR #593, opened 2020-02-14
Users: Display Password Strength (take 2)

bergware:


limetech:
This is quite interesting... looking at their Issue list, wondering how often they update their 'dictionary' - also lots of non-PC words in there.  We need to mull this one over.  Personally I think it's better to use a password manager and have it create a long random string.

Squidly271:
>  Personally I think it's better to use a password manager and have it create a long random string.

Sorry, but when you state this

>  These days, having a bug that presents a security risk is far more worrisome

You should do everything you can to encourage people to have a strong root password, and IMO that means you need to do a check on the strength.

bergware:
> @bergware I would've thought you'd at least add "unraid" to the list

"Calling names" is bad ... added

limetech:
A lot of profanity in the dictionary - I guess I'm going to have to change all my passwords :/

bergware:
> A lot of profanity in the dictionary - I guess I'm going to have to change all my passwords :/

What does that say about your love of Unraid ;)

Squidly271:
I'm honored that you think there's the possibility that anyone would ever use Squid or Squidly as a password.

bergware:
I am wondering whether we should store zxcvbn.js on flash and not ram, because of its size?

limetech:
> I am wondering whether we should store zxcvbn.js on flash and not ram, because of its size (800K)?

Yes I am concerned about the size, and how long it might take to load.  What if we host it in our Amazon S3 download bucket and then load only when user clicks Users page?  I guess problem would be if user doesn't have internet connection...

bergware:
I moved zxcvbn.js from RAM to flash.
There is a check built-in now to see if the file exists and only load/use it when present.
The javascript file loading happens asynchronously, which means there is no delay when the page is opened the first time.

In the next version of Unraid the file zxcvbn.js needs to be included in the package and stored in /boot/config/plugins/dynamix/zxcvbn.js

Squidly271:
Can we call the file something other than zxcvbn.js?  I can see a ton of people asking "What is this" and deleting it since they may be under the mistaken belief that its malware...

bergware:
> Can we call the file something other than zxcvbn.js? I can see a ton of people asking "What is this" and deleting it since they may be under the mistaken belief that its malware...

We went a slightly different way and made a plugin which installs the "zxcvbn.js" file to flash.
This approach leaves a choice to the users whether they want password validation or not, with a recommendation to install the plugin.

If want to try the plugin, this is the URL for manual installation
https://raw.github.com/bergware/dynamix/master/unRAIDv6/dynamix.password.validator.plg

limetech:
Nice work!  I need to pull the PR in order to see what's in the PR :)

limetech:
The password checker plugin is dope!

PR #594, opened 2020-02-15
dockerMan: Preserve \n on overview in basic mode

Squidly271:
\n not being preserved between advanced view and basic view of overviews, making for some strange looking overviews in basic.


limetech:
Thank you!  this has annoyed me for a while.

Squidly271:
It's not quite a perfect situation.  Some of the authors have used [br] and \n within the templates.  Those wind up double spaced, but the majority have forgotten all about encoding html within descriptions, and expect \n to format it.

PR #598, opened 2020-02-22
SysDevs overhaul with vfio-pci.cfg binding

Skitals:
For your review and feedback.

### DIFF
--- a/plugins/dynamix/SysDevs.page
+++ b/plugins/dynamix/SysDevs.page
@@ -4,8 +4,8 @@ Icon="icon-hardware"
 Tag="server"
 ---
 <?PHP
-/* Copyright 2005-2018, Lime Technology
- * Copyright 2012-2018, Bergware International.
+/* Copyright 2005-2020, Lime Technology
+ * Copyright 2012-2020, Bergware International.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License version 2,
@@ -28,6 +28,29 @@ $(function(){
   $('#t3').load('/webGui/include/SysDevs.php',{table:'t3'});
   $('#t4').load('/webGui/include/SysDevs.php',{table:'t4'});
 });
+function applyCfg() {
+  var message = "System Devices: A reboot is required to apply changes";
+  var string = "BIND=";
+  var elements = document.getElementById("vfiopci").elements;
+  for (var i = 0, element; element = elements[i++];) {
+    if (element.type === "checkbox"  && element.checked === true)
+    string = string + element.value + " ";
+  }
+  string = string.trim();
+  if (string === "BIND=") {
+    string = "";
+  }
+  $.get( "/plugins/dynamix/include/update.vfio-pci-cfg.php", { cfg: string } )
+  .done(function(d) {
+    if (d==1) {
+      addRebootNotice(message);
+      document.getElementById("warning").innerHTML = "<b>ALERT: Changes saved. Reboot to take effect.</b>";
+    } else {
+      removeRebootNotice(message);
+      document.getElementById("warning").innerHTML = "<b>No changes.</b>";
+    }
+  });
+}
 </script>
 
 **PCI Devices and IOMMU Groups**
@@ -39,7 +62,12 @@ $(function(){
 To see natural IOMMU groups for your hardware, go to the <b><a href="/Settings/VMSettings">VM Manager</a></b> page and set the <b>PCIe ACS override</b> setting to <b>Disabled</b>.</p>
 <?endif;?>
 
-<pre><table id='t1' class='pre'><tr><td><div class="spinner"></div></td></tr></table></pre><br>
+<? if (is_file("/boot/config/vfio-pci.cfg")) $file = file_get_contents("/boot/config/vfio-pci.cfg"); ?>
+<pre><form id="vfiopci" onsubmit="return false"><table id='t1' class='pre'><tr><td><div class="spinner"></div></td></tr></table></form>
+<input id="applycfg" type="submit" value="Bind selected to VFIO at Boot" onclick="applyCfg();"><span id="warning"></span>
+</form>
+</pre><br>
+
 
 **CPU Thread Pairings**
 
--- a/plugins/dynamix/SysDevs.page
+++ b/plugins/dynamix/SysDevs.page
@@ -63,8 +63,6 @@ To see natural IOMMU groups for your hardware, go to the <b><a href="/Settings/V
 <?endif;?>
 
 <pre><form id="vfiopci" onsubmit="return false"><table id='t1' class='pre'><tr><td><div class="spinner"></div></td></tr></table></form>
-<input id="applycfg" type="submit" value="Bind selected to VFIO at Boot" onclick="applyCfg();"><span id="warning"></span>
-</form>
 </pre><br>
 
 
--- a/plugins/dynamix/SysDevs.page
+++ b/plugins/dynamix/SysDevs.page
@@ -56,15 +56,23 @@ function applyCfg() {
 **PCI Devices and IOMMU Groups**
 
 > This displays a list of IOMMU groups available on your system along with the output of the `lspci` command for each IOMMU group.  The numeric identifiers are used to configure PCI pass-through.
+>
+> Devices you select will be bound to the vfio-pci driver at boot, which makes them available for assignment to a virtual machine, and also prevents the Linux kernel from automatically binding them to any present host driver.
+>
+> **Note that selecting a device will bind not only the specified device(s), but *all* other devices in the same IOMMU group as well.**
+>
+> &nbsp;<i class="fa fa-circle orb green-orb middle"></i>&nbsp;This symbol indicates the device is currently bound to the vfio-pci driver.
+>
+> &nbsp;<i class="fa fa-retweet grey-orb middle"></i>&nbsp;&nbsp;&nbsp;This symbol indicates the device supports FLR (Function Level Reset).
+>
+> <input type="checkbox" value="" disabled>&nbsp;&nbsp;If a checkbox is greyed out it means the device is in use by Unraid and can not be passed through.
 
 <?if (strpos(file_get_contents('/proc/cmdline'), 'pcie_acs_override=') !== false):?>
 <p class="notice" style="line-height:30px;height:auto">Warning: Your system has booted with the PCIe ACS Override setting enabled. The below list doesn't not reflect the way IOMMU would naturally group devices.<br>
 To see natural IOMMU groups for your hardware, go to the <b><a href="/Settings/VMSettings">VM Manager</a></b> page and set the <b>PCIe ACS override</b> setting to <b>Disabled</b>.</p>
 <?endif;?>
 
-<pre><form id="vfiopci" onsubmit="return false"><table id='t1' class='pre'><tr><td><div class="spinner"></div></td></tr></table></form>
-</pre><br>
-
+<pre><form id="vfiopci" onsubmit="return false"><table id='t1' class='pre'><tr><td><div class="spinner"></div></td></tr></table></form></pre><br>
 
 **CPU Thread Pairings**
 
--- a/plugins/dynamix/SysDevs.page
+++ b/plugins/dynamix/SysDevs.page
@@ -62,7 +62,6 @@ function applyCfg() {
 To see natural IOMMU groups for your hardware, go to the <b><a href="/Settings/VMSettings">VM Manager</a></b> page and set the <b>PCIe ACS override</b> setting to <b>Disabled</b>.</p>
 <?endif;?>
 
-<? if (is_file("/boot/config/vfio-pci.cfg")) $file = file_get_contents("/boot/config/vfio-pci.cfg"); ?>
 <pre><form id="vfiopci" onsubmit="return false"><table id='t1' class='pre'><tr><td><div class="spinner"></div></td></tr></table></form>
 <input id="applycfg" type="submit" value="Bind selected to VFIO at Boot" onclick="applyCfg();"><span id="warning"></span>
 </form>
--- a/plugins/dynamix/include/SysDevs.php
+++ b/plugins/dynamix/include/SysDevs.php
@@ -72,7 +72,6 @@ case 't1':
           echo "<i class=\"fa fa-circle orb green-orb middle\" title=\"Kernel driver in use: vfio-pci\"></i>";
           $isbound = "true";
         }
-        unset($outputvfio);
         echo "</td><td>";
         if ((strpos($line, 'Host bridge') === false) && (strpos($line, 'PCI bridge') === false)) {
           if (file_exists('/sys/kernel/iommu_groups/'.$iommu.'/devices/0000:'.$pciaddress.'/reset')) echo "<i class=\"fa fa-retweet grey-orb middle\" title=\"Function Level Reset (FLR) supported.\"></i>";
@@ -80,7 +79,10 @@ case 't1':
           echo in_array($iommu, $iommuinuse) ? ' <input type="checkbox" value="" title="In use by Unraid" disabled ' : ' <input type="checkbox" value="'.$pciaddress.'" ';
           echo (strpos($file, $pciaddress) !== false) ? " checked>" : ">";
         } else { echo "</td><td>"; }
-        echo "</td><td>$line</td></tr>";
+        echo '</td><td title="';
+        foreach ($outputvfio as $line2) echo "$line2&#10;";
+        echo '">'.$line.'</td></tr>';
+        unset($outputvfio);
         switch (true) {
           case (strpos($line, 'USB controller') !== false):
             if ($isbound) {
--- a/plugins/dynamix/include/SysDevs.php
+++ b/plugins/dynamix/include/SysDevs.php
@@ -13,28 +13,6 @@
 <?
 switch ($_POST['table']) {
 case 't1':
-  if (is_file("/boot/config/vfio-pci.cfg")) $file = file_get_contents("/boot/config/vfio-pci.cfg");
-  $disks = (array)parse_ini_file('state/disks.ini',true);
-  $devicelist = array_column($disks, 'device');
-  $lines = array ();
-  foreach ($devicelist as $line) {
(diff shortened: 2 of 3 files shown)


bergware:
This looks good.
Can you add "help" information to explain the different symbols, like a 'green orb' and the 'refresh icon', with some explanation too about the check-boxes (some are disabled).

For code consistency, can you use "spaces" instead of "tabs" (1 tab = 2 spaces).
Thanks

Skitals:
> This looks good.
> Can you add "help" information to explain the different symbols, like a 'green orb' and the 'refresh icon', with some explanation too about the check-boxes (some are disabled).
> 
> For code consistency, can you use "spaces" instead of "tabs" (1 tab = 2 spaces).
> Thanks

Is there an example somewhere else in the webgui how this "help" information should be presented? All the items mentioned are explained via the html title attribute when you mouse over them.

bergware:
Help text is indicated by lines starting with a > sign.
You can expand the help text under "PCI devices and IOMMU groups"
E.g.
`> This displays a list of IOMMU groups available on your system...
>
> <i class='fa fa-circle orb green-orb'></i> kernel driver in use by system
>
> <i class='fa fa-retweet grey-orb'></i> Function Level Reset (FLR) supported
`

bergware:
One other question: have you tested this on a system without IOMMU groups?

Squidly271:
> One other question: have you tested this as well on a system without IOMMU groups?

Isn't that pointless for what it does?  Unless your talking about a condition where the page tosses up a message if IOMMU is disabled

bergware:
> Isn't that pointless for what it does? Unless your talking about a condition where the page tosses up a message if IOMMU is disabled

Reason to ask is to make sure the display isn't messed up when no groups are present. I.e. give a normal listing

Skitals:
> Help text is indicated by lines starting with a > sign.
> You can expand the help text under "PCI devices and IOMMU groups"

Ahh, I have never pressed the "help" button before so I wasn't sure where that info was presented. I will expand on this.


> One other question: have you tested this as well on a system without IOMMU groups?

I have not, but I did not change the output if there are no IOMMU groups. That is still handled by the "if (empty($groups))" statement. The "Bind Selected" button would still be visible, so perhaps it should also be hidden if there are no IOMMU groups.

bergware:
Looks good.
One small thing, in the file "update.vfio-pci-cfg.php" add the "-f" flag to the copy command.
This will force an overwrite when the .bak file already exist

Skitals:
> Looks good.
> One small thing, in the file "update.vfio-pci-cfg.php" add the "-f" flag to the copy command.
> This will force an overwrite when the .bak file already exist

Done. The backup copy is a holdover from my plugin which had an "undo" function that I removed to streamline/simplify things. Could probably just remove it all together, but I figured creating a backup doesn't hurt.

ljm42:
This is really cool!

According to the 6.7.0-rc1 release notes:
 https://forums.unraid.net/bug-reports/prereleases/unraid-os-version-670-rc1-available-r351/?tab=comments&fbclid=IwAR20LHp-FI_IPs_45OXtacPDFoXbCLxGK3vGz7Im4w4KArwGt4l4AuJClyU

> "either all devices of an IOMMU group are bound to vfio-pci or none of them are."

If that is still accurate, can we make sure that all three devices in this IOMMU group are greyed out, and not just the disk controller?
image

Skitals:
> This is really cool!
> 
> According to the 6.7.0-rc1 release notes:
> https://forums.unraid.net/bug-reports/prereleases/unraid-os-version-670-rc1-available-r351/?tab=comments&fbclid=IwAR20LHp-FI_IPs_45OXtacPDFoXbCLxGK3vGz7Im4w4KArwGt4l4AuJClyU
> 
> > "either all devices of an IOMMU group are bound to vfio-pci or none of them are."
> 
> If that is still accurate, can we make sure that all three devices in this IOMMU group are greyed out, and not just the disk controller?
> image

Well, this hurt my brain more than it should have, but it now takes the array of devices in use and builds a new array of IOMMU groups in use and excludes all devices belonging to those groups.

Skitals:
> This is really cool!
> 
> According to the 6.7.0-rc1 release notes:
> https://forums.unraid.net/bug-reports/prereleases/unraid-os-version-670-rc1-available-r351/?tab=comments&fbclid=IwAR20LHp-FI_IPs_45OXtacPDFoXbCLxGK3vGz7Im4w4KArwGt4l4AuJClyU
> 
> > "either all devices of an IOMMU group are bound to vfio-pci or none of them are."
> 
> If that is still accurate, can we make sure that all three devices in this IOMMU group are greyed out, and not just the disk controller?
> image

Please try it out, all of my in use devices were already in their own IOMMU groups so the output hasn't changed on my machine.

ljm42:
Woot!  Looks great:
image

ljm42:
My system has two NICs:
image

would it make sense to mark the one(s) that are in use by Unraid as not being available for vfio-pci? If users bind the wrong one then they will lose access to the network when their system reboots.

Some ideas... from here:
```
root@Tower:~# readlink /sys/class/net/eth*
../../devices/pci0000:00/0000:00:1c.2/0000:02:00.0/net/eth0
../../devices/pci0000:00/0000:00:1c.3/0000:03:00.0/net/eth1
```
we can see my system has two eth* devices:

-  eth0 at PCI address 02:00.0
-  eth1 at PCI address 03:00.0

I don't know if there is a better way, but using "ip link" we can see that eth0 is in use while eth1 is not:
```
root@Tower:~# ip link show eth0 | grep "state UP"
10: eth0: <BROADCAST,MULTICAST,PROMISC,UP,LOWER_UP> mtu 1500 qdisc mq master br0 state UP mode DEFAULT group default qlen 1000
root@Tower:~# ip link show eth1 | grep "state UP"
```
So if we add eth0's PCI address of "02:00.0" to the $inuse[] array, then it wouldn't be able to be assigned to vfio-pci and we don't have to worry about users inadvertently losing access to their network

If users *want* to assign an in-use NIC to vfio-pci, they would first need to change their Unraid settings to disable the NIC.

Skitals:
> My system has two NICs:
> image
> 
> would it make sense to mark the one(s) that are in use by Unraid as not being available for vfio-pci? If users bind the wrong one then they will lose access to the network when their system reboots.
> 
> Some ideas... from here:
> 
> ```
> root@Tower:~# readlink /sys/class/net/eth*
> ../../devices/pci0000:00/0000:00:1c.2/0000:02:00.0/net/eth0
> ../../devices/pci0000:00/0000:00:1c.3/0000:03:00.0/net/eth1
> ```
> 
> we can see my system has two eth* devices:
> 
> * eth0 at PCI address 02:00.0
> * eth1 at PCI address 03:00.0
> 
> I don't know if there is a better way, but using "ip link" we can see that eth0 is in use while eth1 is not:
> 
> ```
> root@Tower:~# ip link show eth0 | grep "state UP"
> 10: eth0: <BROADCAST,MULTICAST,PROMISC,UP,LOWER_UP> mtu 1500 qdisc mq master br0 state UP mode DEFAULT group default qlen 1000
> root@Tower:~# ip link show eth1 | grep "state UP"
> ```
> 
> So if we add eth0's PCI address of "02:00.0" to the $inuse[] array, then it wouldn't be able to be assigned to vfio-pci and we don't have to worry about users inadvertently losing access to their network
> 
> If users _want_ to assign an in-use NIC to vfio-pci, they would first need to change their Unraid settings to disable the NIC.

I have already put some thought into this and my thought was to parse state/network.ini the same way I parse state/disks.ini and search it for eth*.

ljm42:
> I have already put some thought into this and my thought was to parse state/network.ini the same way I parse state/disks.ini and search it for eth*.

Good idea! I'm happy to help test

Skitals:
> > I have already put some thought into this and my thought was to parse state/network.ini the same way I parse state/disks.ini and search it for eth*.
> 
> Good idea! I'm happy to help test

This only took a bit of copy pasta, test it out and let me know.

ljm42:
This looks good!
image

But this entry is now marked in use as well, and I'm not sure why?
image

I reverted to the previous commit and confirmed "[8086:8c14] 00:1c.2" was not marked in use before, it is something to do with this latest commit

Here are some more details:
```
root@Tower:~# cat /usr/local/emhttp/state/network.ini 
[eth0]
DHCP_KEEPRESOLV="yes"
DNS_SERVER1="8.8.8.8"
DNS_SERVER2="8.8.4.4"
DNS_SERVER3=""
DHCP6_KEEPRESOLV="no"
BONDING="no"
BONDNAME=""
BONDNICS=""
BONDING_MODE="1"
BONDING_MIIMON="100"
BRIDGING="yes"
BRNAME="br0"
BRNICS="eth0"
BRSTP="no"
BRFD="0"
DESCRIPTION:0=""
PROTOCOL:0=""
USE_DHCP:0="no"
IPADDR:0="192.168.10.51"
NETMASK:0="255.255.255.0"
GATEWAY:0="192.168.10.1"
METRIC:0=""
USE_DHCP6:0=""
IPADDR6:0=""
NETMASK6:0=""
GATEWAY6:0=""
METRIC6:0=""
PRIVACY6:0=""
MTU=""
TYPE="access"
```

```
root@Tower:~# readlink /sys/class/net/*
../../devices/virtual/net/br0
../../devices/virtual/net/docker0
../../devices/virtual/net/erspan0
../../devices/pci0000:00/0000:00:1c.2/0000:02:00.0/net/eth0
../../devices/pci0000:00/0000:00:1c.3/0000:03:00.0/net/eth1
../../devices/virtual/net/gre0
../../devices/virtual/net/gretap0
../../devices/virtual/net/ip_vti0
../../devices/virtual/net/lo
../../devices/virtual/net/sit0
../../devices/virtual/net/tunl0
../../devices/virtual/net/veth0c3153d
../../devices/virtual/net/veth3f2a4a5
../../devices/virtual/net/veth40bcff1
../../devices/virtual/net/veth4aa7b88
../../devices/virtual/net/veth859448b
../../devices/virtual/net/veth908b39a
../../devices/virtual/net/veth93873ba
../../devices/virtual/net/veth971ae08
../../devices/virtual/net/veth9dd03c4
../../devices/virtual/net/vethaa07888
../../devices/virtual/net/vethabe9703
../../devices/virtual/net/virbr0
../../devices/virtual/net/virbr0-nic
../../devices/virtual/net/vnet0
../../devices/virtual/net/wg0
../../devices/virtual/net/wg1
```

```
root@Tower:~# ls /sys/kernel/iommu_groups/*/devices/0000* -1 -d
/sys/kernel/iommu_groups/0/devices/0000:00:00.0@
/sys/kernel/iommu_groups/1/devices/0000:00:14.0@
/sys/kernel/iommu_groups/10/devices/0000:02:00.0@
/sys/kernel/iommu_groups/11/devices/0000:03:00.0@
/sys/kernel/iommu_groups/12/devices/0000:04:00.0@
/sys/kernel/iommu_groups/12/devices/0000:05:00.0@
/sys/kernel/iommu_groups/2/devices/0000:00:16.0@
/sys/kernel/iommu_groups/2/devices/0000:00:16.1@
/sys/kernel/iommu_groups/3/devices/0000:00:1a.0@
/sys/kernel/iommu_groups/4/devices/0000:00:1c.0@
/sys/kernel/iommu_groups/5/devices/0000:00:1c.2@
/sys/kernel/iommu_groups/6/devices/0000:00:1c.3@
/sys/kernel/iommu_groups/7/devices/0000:00:1c.7@
/sys/kernel/iommu_groups/8/devices/0000:00:1d.0@
/sys/kernel/iommu_groups/9/devices/0000:00:1f.0@
/sys/kernel/iommu_groups/9/devices/0000:00:1f.2@
/sys/kernel/iommu_groups/9/devices/0000:00:1f.3@

```

Skitals:
../../devices/pci0000:00/0000:00:1c.2/0000:02:00.0/net/eth0

00:1c.2 is the root port for the controller.

ljm42:
Oh! I see it now :) In that case I'd say this is working great!

Skitals:
> Oh! I see it now :) In that case I'd say this is working great!

That being said, I don't think you can pass it through either way, and I don't believe adding 00:1c.2 to vfio-pci.cfg would do anything at all.

It is probably safe to simply remove the checkbox for all "PCI bridge" and "Host bridge" device classes. Is there ANY use case I'm not thinking of? Limiting the total number of checkboxes would certainly make it visually easier to see what are potential pass-through device candidates.

Skitals:
New commit completely removes the checkboxes (and FLR icon) for PCI bridges and Host bridges. This greatly declutters the interface and makes it easier to see what you might want to actually select. Someone please chime in if there is a use case I am missing.

bergware:
More and more promising...
Is it possible to lock the USB controller which is used for the flash device of Unraid?

ljm42:
Here is what my system looks like now. I question the "Intel C220 Series Chipset Family MEI Controllers", should those have the checkbox removed?
image
image

ljm42:
`lspci -nk` shows the kernel driver and kernal modules in use by each device, is there room to add this to the page? It seems like useful information to have on a page that lists out all of the system devices, whether IOMMU groups are available or not.

```
root@Tower:~# lspci -nk
00:00.0 0600: 8086:0c08 (rev 06)
        Subsystem: 1849:0c08
        Kernel driver in use: ie31200_edac
        Kernel modules: ie31200_edac
00:14.0 0c03: 8086:8c31 (rev 05)
        Subsystem: 1849:8c31
        Kernel driver in use: xhci_hcd
00:16.0 0780: 8086:8c3a (rev 04)
        Subsystem: 1849:8c3a
00:16.1 0780: 8086:8c3b (rev 04)
        Subsystem: 1849:8c3b
00:1a.0 0c03: 8086:8c2d (rev 05)
        Subsystem: 1849:8c2d
        Kernel driver in use: ehci-pci
00:1c.0 0604: 8086:8c10 (rev d5)
        Kernel driver in use: pcieport
00:1c.2 0604: 8086:8c14 (rev d5)
        Kernel driver in use: pcieport
00:1c.3 0604: 8086:8c16 (rev d5)
        Kernel driver in use: pcieport
00:1c.7 0604: 8086:8c1e (rev d5)
        Kernel driver in use: pcieport
00:1d.0 0c03: 8086:8c26 (rev 05)
        Subsystem: 1849:8c26
        Kernel driver in use: ehci-pci
00:1f.0 0601: 8086:8c56 (rev 05)
        Subsystem: 1849:8c56
00:1f.2 0106: 8086:8c02 (rev 05)
        Subsystem: 1849:8c02
        Kernel driver in use: ahci
        Kernel modules: ahci
00:1f.3 0c05: 8086:8c22 (rev 05)
        Subsystem: 1849:8c22
        Kernel driver in use: i801_smbus
        Kernel modules: i2c_i801
02:00.0 0200: 8086:1533 (rev 03)
        Subsystem: 1849:1533
        Kernel driver in use: igb
        Kernel modules: igb
03:00.0 0200: 8086:1533 (rev 03)
        Subsystem: 1849:1533
        Kernel driver in use: igb
        Kernel modules: igb
04:00.0 0604: 1a03:1150 (rev 02)
05:00.0 0300: 1a03:2000 (rev 21)
        Subsystem: 1849:2000
        Kernel modules: ast
```

Skitals:
I already have lspci run on each device to check if it was bound to the vfio driver, this commit shows the whole verbose output as a title tooltip for each device. I've only tested in chrome so far, hopefully all browsers respect &#10; as a line break in the tooltip.

ljm42:
This is a good solution. It allows users to answer the two questions from my last post without having to go to the command line, and it doesn't clutter the screen.

The linebreaks in the tooltip work for me in both Chrome and Firefox on Windows

bergware:
Browsers have no issue with the 10 (LF) character code.
Tested the last version and looking good.

Skitals:
> Here is what my system looks like now. I question the "Intel C220 Series Chipset Family MEI Controllers", should those have the checkbox removed?
> image
> image

https://www.kernel.org/doc/Documentation/misc-devices/mei/mei.txt

It looks like something that COULD be passed through it you wanted to give you guest an interface to the Intel Management Engine.

ljm42:
> It looks like something that COULD be passed through it you wanted to give you guest an interface to the Intel Management Engine.

Ah, thanks for that!

Skitals:
With this commit, when you check/uncheck a device, it checks/unchecks all other selectable devices in the same iommu group. This drives home the fact that it is all or nothing when passing through a device that is not in its own iommu.

ljm42:
The "select all in IOMMU group" feature works perfectly, both in the UI and in the data that is added to vfio-pci.cfg. nice!

Skitals:
I simplified the js on the previous commit.

ljm42:
In this line:
https://github.com/limetech/webgui/blob/399110f6baf87ebfb0fad90b9f272541d7719293/plugins/dynamix/include/SysDevs.php#L91

How about sorting the output of lsusb:
  `lsusb -s $bus:|sort`

That turns this output:
image
into this:
image
which matches the list of USB devices below

ljm42:
thanks!

ljm42:
I have one more request for this... how about disabling the "Bind Selected" button by default, then enabling it when the user checks or unchecks a box? 

This should make the interface more clear at a glance.

ljm42:
I've been studying the vfio-pci script and I think we're taking a shortcut that we shouldn't.  We are currently writing "Bus:Device.Function" (i.e. 00:00.0) to the config file, but the script really wants the full "Domain:Bus:Device.Function" identifier (i.e. 0000:00:00.0). I don't have hardware with multiple domains to test with, but it seems risky to build a gui that assumes the Domain is always "0000".

I'd recommend that we do what vfio-pci does:
- If the config file specifies the full "Domain:Bus:Device.Function", use it as is
- If just "Bus:Device.Function" is specified, prepend "0000:" to it before processing

Then modify the rest of the code so that it doesn't assume the domain is "0000", and write the full "Domain:Bus:Device.Function" back out to the config file:
`BIND=0000:03:00.0`

That should allow this solution to work on a wider range of hardware

ljm42:
One little thing, the code currently puts `\n` in an "empty" vfio-pci.cfg file, so when you pass it to the vfio-pci script you get this error:

```./vfio-pci: line 158: declare: `=': not a valid identifier```

Can we make sure that an empty file really is empty? Or maybe just delete it?

PR #602, opened 2020-03-04
Multi language support

bergware:


Squidly271:
Very nice addition.

Is there a wrapper (or can there be) in javascript or php to handle dynamically generated content?

IE:  While I can't do anything about CA's descriptions of templates, there is also a ton of other sections within CA that can benefit from this.  While some of them are static, others are actually dynamically generated and everything is returned via post calls.

bergware:
I am in the process of writing a guide for developers on how to add multi
language support to plugins, like CA.

Dynamically generated content usually needs some adjustment of the code and
post calls are handled by changing the source (i.e. the code which is
called).

Cheers,
Ronald

Squidly271:
Happy Happy Joy Joy.  More compatibility code.

ljm42:
Can we rethink the use of session_start() in Translations.php? This is a big departure from the current code, which only starts sessions for authenticated users. This is because the session timeout is set extremely long, and we don't want every random Unraid.net connection creating a long-lived session file.

bergware:
The multi language implementation relies on session binding, which is needed to preserve the "locale" setting across the different php scripts being invoked.

Session binding is achieved by setting the initial session in template.php, which is the root page for all GUI pages. This makes the "locale" available to all *.page files, but also to php scripts invoked by post and which need translations too (these scripts need to do a 'require_once Translations.php' to get access to translations).

The SESSION variable then looks like this
```
 [_SESSION] => Array
        (
            [unraid_login] => 1583476727
            [unraid_user] => root
            [locale] => nl_NL
        )
```
If you believe there is another way to store session information and preserve the "locale" setting,

I am not sure how this would affect "authenticated" users, you still need to login to gain access and can logout as before.

ljm42:
I want to avoid the chance of having thousands of open sessions due to the random script kiddies on the Internet. Since the tmp dir for session files is in RAM, all of those session files will eventually eat up the available RAM.  Unraid has a super long session timeout value which exacerbates this.

How about starting the session only if the session cookie already exists or if the user changes the locale away from the default? The code would simply use default values if the session has not been started.

Also, it seems like the session start code should move to auth_request.php so all sessions are started in one place (well, two places counting login.php)

Note that in auth_request.php the session is closed as soon as possible after data is written to it.   This allows PHP to support concurrent requests for the same session. If the session stays open for the entire time it takes a script to run, it basically becomes a single threaded app.

bergware:
> Also, it seems like the session start code should move to auth_request.php so all sessions are started in one place (well, two places counting login.php)

This works only for any *.page file loaded, but not for any php script invoked thru a javascript post call (post calls btw are protected by the csrf_token mechanism)

> How about starting the session only if the session cookie already exists

This is a catch-22 situation. A script invoked by post does not have the session variables available, and need session binding first to read them.

My testing shows that at the time "template.php" is loaded it doesn't have any session variables preserved and I need to do session binding first to get them populated, similar story for "translations.php" (btw I added session_write_close() now too for improved page handling).

As far as I can see what happens here is simply session binding, but not the creation of new sessions. There are no "thousands of sessions", just a single one, but of course you may prove me wrong.

ljm42:
Every time session_start() is called, if a session cookie does not exist a new session is created.

Take a look at the list of session files in /var/lib/php/. Now use incognito mode to access the site running this code and you'll see another session file has been added. Now multiply this by all the script kiddies out there. Pretty sure even zero byte files use a 4K block of space. 

Most production sites don't have an issue with this because they don't have 7 day session lengths (see /etc/php/ ) . And they probably don't have their session files in RAM, or they can just add more RAM to handle it.

What I'm suggesting is instead of always calling session_start(), first check to see if the session cookie exists. If it does, then the user has a session already and you should use it. If the cookie does not exist then just use default values and don't call session_start()

bergware:
My test:

1. Logout of the GUI
2. Close the browser
3. Delete all session files in /var/lib/php
4. Re-open the browser
5. Login as usual (I am using Dutch translations)
6. Go to the Main page (screen updates happen here every second, including calling of Translations.php and the subsequent session binding each second)
7. After a minute check how many session files exist ==> ONE (the initial session after logging in)

In short, it is exactly working as I expect it to work :)

ljm42:
Thanks for testing. You are still using a single browser though right? I'm glad that has a single session file, otherwise things are really broken: ) 

Now use a different browser, don't even login just hit the site. I expect that with this code you will see a new session file. Now imagine many bots doing the same, except they won't store cookies, so many bots hitting the site many times will generate many*many session files over a period of 7 days.

Sessions for non-authenticated users are useless.

bergware:
Ok, next test

1. current number of session files ==> 1 (using Brave browser)
2. Open GUI (login screen only) in Chrome
3. Open GUI (login screen only) in Firefox
4. Open GUI (login screen only) in Edge
5. Check number of session files ==> 1 (still my Brave browser which is logged in)

ljm42:
OK that is unexpected. Does that make sense to you?

bergware:
Yes, fully expected, it means no sessions are created unless I authenticate and enter the GUI after login.

ljm42:
Oh I was assuming login.php was translated too, is it not?

bergware:
No, the login screen is not translated.

If I include the translations file in the login file, before authentication is done, then new session files are created each time the login page is requested, as you explained.

So, to stay safe ... no translations at login level, which I think is acceptable.

Thanks for clearing that.

ljm42:
> So, to stay safe ... no translations at login level, which I think is acceptable.
>
> Thanks for clearing that.

Thanks, sorry I didn't realize that sooner. Good discussion anyway :) 

Would it be possible to dynamically set the html language code on the files that support translation? https://github.com/limetech/webgui/blob/c8e7e3de42e3b7cc9aae49defc2db4e41097db18/plugins/dynamix/include/DefaultPageLayout.php#L22

BTW, I installed this patch on my 6.8.3 VM to test it, but I'm having trouble switching languages. Under Settings -> Display -> Language, the only option is English?

Squidly271:
> BTW, I installed this patch on my 6.8.3 VM to test it, but I'm having trouble switching languages. Under Settings -> Display -> Language, the only option is English?

Waiting for a merge before I try this, but isn't that a catch-22?  Multi-Language support, but you have to be able to read and understand English in order to switch the language.  Shouldn't there be a little flag or something on the menu bar?

Squidly271:
Assuming that limetech is going to wind up hosting a repository of the languages, what I'd like to see on the Language page is that it indexes that online repository and then downloads the zip once the user chooses the language they want. (along with offering the ability to upload your own custom one)   Make it easy to switch if you can't read the default English.

ljm42:
Ah, thanks for the hint @Squidly271 . I wasn't expecting a separate install process.

Are there any language files to test with?

How about adding a Pig Latin option? I'd have a better chance of navigating with that :)

Squidly271:
And Klingon

bergware:
There are two aspects in multi-language support.

1. A new language needs to be installed.
   Herefore the "Language" option under Tools is made.
   The language import tool has a number of predefined languages (written in their native wording)
   If somebody makes a translation for a language not available in the preset selection,
   he/she can make a request and it will be added (this is explained in the README guide)
   By default Unraid has English only, and any other language needs to be installed.
   Installing means obtaining a ZIP file with the translations and add this as a new language

2. Switching to the new language in the GUI
   This is selection is done under Display Settings.
   Here you can select from the languages which are installed, initially this is English only.
   Any other language added thru the Language tool, will be made visible here.
   Select a language and press "Apply" and voila .. de nieuwe taal wordt gebruikt

The proposal is to have a language section on the forum, where translations can be offered and be downloaded. We are going to rely on the community to get different languages included.
As a starter I have made Dutch translations, and I hope others will follow with their respective language. This is something which needs to grow over time...

Qapla'! - that's good luck in Klingon :)

bergware:
> Would it be possible to dynamically set the html language code on the files that support translation?

Yes, this is possible, but I wonder what will be the added value?

bergware:
If you are wondering how a translated page looks like, below the Main page in Dutch.

image

Squidly271:
> If somebody makes a translation for a language not available in the preset selection,
he/she can make a request and it will be added (this is explained in the README guide)

Didn't notice Polish in there.  Not that I can speak it (except for the swear words).

Dazog:
I would suggest a hosted version of Zanata for translators to translate in browser.

Pfsense uses it with great success for their project.

Squidly271:
Dziękuję

bergware:
Zapraszamy

I see some task for you ...  (you can't hide your roots)

Squidly271:
> I see some task for you ... (you can't hide your roots)
Sure.  If you want say the tabs to be "Pizda", "pierdolić", "kogut" etc.

But seriously though, the problem with any translation (especially a machine translation) is that context may get lost.  A direct translation isn't always appropriate as a foreign language speaker might look at the translation and shake his/her head as while it does translate correctly, it doesn't make any sense in the context of unRaid.

limetech:
> >   So, to stay safe ... no translations at login level, which I think is acceptable.
> > 
> >     Thanks for clearing that.
> 
> Thanks, sorry I didn't realize that sooner. Good discussion anyway :)

Yes very good discussion and a suggestion:  I'm going to merge this now but in a future PR we should add a comment in login.php that the login page cannot be translated or else we get script kiddie fun.  For sure someone will come along in the future thinking, "hey these guys forgot to translate the login page, let me fix that ..."

Squidly271:
Now that this is merged, any updates @bergware 

> I am in the process of writing a guide for developers on how to add multi
language support to plugins, like CA.

> Dynamically generated content usually needs some adjustment of the code and
> post calls are handled by changing the source (i.e. the code which is
> called).

bergware:
Sure, I have made a PDF document with instructions.

Squidly271:
TLDR: CA won't support this without an auxiliary dynamix plugin for compatibility.

Sent you a PM, but it's going to be basically an extreme PITA for plugins to support this without setting the minimum version of the plugin to be 6.9+ or maintaining 2 separate code bases.

What's needed for plugin adoption is an auxiliary dynamix plugin that can be conditionally installed by the plugin that will supply the necessary javascript / php functions  to prevent the various errors being generated when running on older versions of unRaid.  We don't need all of the 185 files that have been modified by this PR, but simply the bare minimum files needed to prevent the errors from appearing and thus have the  .page files simply have a conditional to include those scripts based upon the unRaid version.

Otherwise, adoption of this system for the 3rd party plugins will be basically 0 until such time as it's reasonable for a plugin to drop support for all previous versions.

For a simple plugin, a simple conditional test on the unRaid version and whether or not to include the tag isn't a big deal.  For CA to even begin to support this and maintain compatibility, it requires either 2 separate code bases or hundreds of conditional tests.  A dynamix plugin to install the necessary support files on previous unRaid versions will allow CA to maintain a single code base.

PR #604, opened 2020-03-15
Cache plugin .cfg reads from flash drive

Squidly271:
Not really sure if this is an actual thing or not https://en.wikipedia.org/wiki/Flash_memory#Read_disturb , but it is rather disturbing the amount of flash drive corruption that happens, seemingly out of no where.  (And I personally wouldn't ever count on a flash drive's controller being halfway decent and detecting this) (Dynamix' monitor reads the same file every minute on the minute).  If nothing else, it will give a speed bump

### DIFF
--- a/plugins/dynamix/include/Wrappers.php
+++ b/plugins/dynamix/include/Wrappers.php
@@ -23,7 +23,7 @@ function parse_plugin_cfg($plugin, $sections=false, $scanner=INI_SCANNER_NORMAL)
     @mkdir(dirname($cache),0777,true);
     @copy($rom,$cache);
   }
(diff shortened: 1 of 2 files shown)


bergware:
I was always under the believe that Linux would cache these files already and not access the USB all the time.

Squidly271:
My understanding with any caching is that it's not a guarantee depending upon what else is going on

limetech:
ELI5 what this is doing.

Squidly271:
It explicitly caches any plugin's .cfg files in RAM to lessen reads from the flash drive if read-disturb is an actual thing.  Depending upon the plugin, the .cfg file may be read quite often (dynamix' .cfg is read every minute), CA with notifications enabled is read every hour.  Who knows how often other plugins .cfgs are read.

PR #607, opened 2020-03-20
Multi cache pools

bergware:


bergware:
The first steps, which allow adding and removing pools from the GUI, see Main page

bergware:
Next step, dashboard

PR #619, opened 2020-03-23
Multi-language support

Squidly271:
Required for CA to support multi-language.

Here's the story

When on multi-language, if the user selects "Dutch", then CA operates in Dutch.  However when installing a container from CA, the language on the Add Container screen will revert to English.

The workaround for that is for CA to specify the install URL as /Docker/Apps/AddContainer... instead of the current /Apps/AddContainer.

But, that change breaks the "Done" Button in returning to CA.  This change fixes that.  Doesn't appear to affect anywhere else in the UI where the Done button is also present, but @bergware might want to double check that

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -165,6 +165,7 @@ function settab(tab) {
 }
 function done(key) {
   var url = location.pathname.split('/');
+  // Ensure Done() goes back to CA if referrer is CA
   if (location.pathname.indexOf("/Apps/") != -1 ) 
     var path = "/Apps";
   else 
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -165,7 +165,7 @@ function settab(tab) {
 }
 function done(key) {
   var url = location.pathname.split('/');
-  if (url.indexOf("/Apps/")) 
+  if (location.pathname.indexOf("/Apps/") != -1 ) 
     var path = "/Apps";
   else 
     var path = '/'+url[1];
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -165,7 +165,7 @@ function settab(tab) {
 }
 function done(key) {
   var url = location.pathname.split('/');
-  var path = '/'+url[1];
+  var path = '/'+url[url.length - 2];
   if (key) for (var i=2; i<url.length; i++) if (url[i]==key) break; else path += '/'+url[i];
   $.removeCookie('one',{path:'/'});
   location.replace(path);
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -165,7 +165,10 @@ function settab(tab) {
 }
 function done(key) {
   var url = location.pathname.split('/');
-  var path = '/'+url[url.length - 2];
+  if (url.indexOf("/Apps/")) 
+    var path = "/Apps";
+  else 
+    var path = '/'+url[1];
   if (key) for (var i=2; i<url.length; i++) if (url[i]==key) break; else path += '/'+url[i];
   $.removeCookie('one',{path:'/'});
   location.replace(path);


bergware:
Instead of this change, try the following in your CA code:

Change
<input type="button" value="_(Done)_" onclick="done()">
to
<input type="button" value="_(Done)_" onclick="done('Apps')">

Squidly271:
That won't work.  CA doesn't have any "Done" button anywhere.  This is about changing the Done button on the Add Container Screen / After the container is installed, etc.

bergware:
Ok, but your change has the side-effect that anything called as "sub-page" won't return to the "root" page. I can't tell for sure if this is 100% ok for the GUI.

Squidly271:
An alternative if you're ok with it is to put a special case in for CA on the Done() function where it sees if /Apps/ is present in the URL and then sets the URL to be /Apps/ instead of the root / parent as it currently does.  Net result is the same for me.

Squidly271:
I want to keep after any install the user lands back on CA.

bergware:
I understand why you want to do this, it was the reason at the time to introduce a "key" parameter with the done function to make a selective return. Alas not suitable for CA.

Let me think over it, how to approach this. It's about bedtime overhere and need some sleep ;)

bergware:
I changed the Docker part to look for /Apps and you can call the page as normal, i.e.

/Apps/AddContainer

This means no workarounds in the done() function needed.

bergware:
Ok, your last PR is the most plausible solution.

Squidly271:
You dont get much sleep do you? Lol

limetech:
suggestion: special cases deserve comments in the code:
//  Ensure Done() goes back to CA if referrer is CA

bergware:
Rethinking this again.
The current translation code allows to refer to multiple sections, e.g. set URI as follows

$_SERVER['REQUEST_URI'] = 'docker/apps';

The above would first load docker translations and next add the apps translations.

Can you write the CA code to make use of this? and keep the your original page reference /Apps

Squidly271:
OK, that worked for our other problem with my  require_once's taking precedence over CA's translation files.

It also worked for CA going to any other plugin's settings (they stayed in Dutch, and the Done would go back to CA).

But, for adding a container it reverted back to English.

There is one notable difference between how the two different links work though.  CA going to a Plugin's settings is handled via `<a href='....'>`  CA installing / editing a container is actually handled via javascript where it first has to create the xml template from what the feed gave CA, and then the javascript does a `window.open(url,"_self")`

bergware:
One of the design goals of multi-language support, is to ensure minimum impact on the GUI operation, in other words, no noticable slowness due to translations.
This is the background of the sections approach, just load the necessary translations required to view the active page.

All sections in the stock GUI (i.e. no plugins) work properly with the mutli-language implementation, because they operate autonomously. That means the Dashboard and Main pages have their own translations next to the general translations, so have the other sections, like Docker or VMs.

Frankly speaking, I needed to make modifications to the Docker and VMs parts, to get them aligned. I took the effort, because in the longer run it is much easier to maintain when everything works the same way.

CA needs "specials" because it does cross-section operation and the challenge here is to get all necessary translations in place due to the way CA operates, which differs from the rest of the GUI.

We should explore a bit more how to solve it in CA itself, not saying you need to completely rewrite CA, but if possible tackle the problem at the source. At the end, we may still come to the conclusion that an exception is requried, And call it "done" :)

How does the "url" look like which is used in your javascript?

Squidly271:
`/Docker/Apps/AddContainer?xmlTemplate=default:/tmp/community.applications/tempFiles/templates-community-apps/ich777sRepository/IW4xServer.xml`

Squidly271:
```
function openNewWindow(newURL) {
	var popUp = window.open(newURL,"_self");
	if ( !popUp || popUp.closed || typeof popUp == "undefined" ) {
		addBannerWarning(translate("Popup Blocked.  CA requires popups to be enabled under certain circumstances.  You must white list your server within your browser to allow popups"),true,true);
	}
}

function xmlInstall(type,xml) {
	post({action:'createXML',xml:xml},function(result){
		if ( result.status == "ok" ) {
			openNewWindow("/Docker/Apps/AddContainer?xmlTemplate="+type+":"+xml);
		}
	});
}

```

bergware:
I did some investigation to find out what is exactly happening.

The problem is not missing translations in this case, but a chicken and egg situation.

When the page is called as /Apps/AddContainer it will load the docker translations, but AFTER the "AddContainer" page is loaded, hence the docker specific items are not translated.

When the page is called as /Docker/Apps/AddContainer, then the docker translations are present BEFORE the "AddContainer" is loaded and translation is done.

To solve the issue I made an update to the "AddContainer" and "UpdateContainer" pages to add the docker translations if not present BEFORE calling the page content.
With this in place, CA can simply stick to the existing reference /Apps/AddContainer.

Side note, not relevant but you may want to change the window.open function (and relying on the browser to have this allowed) to a simple location function, which doesn't need specific allowance. 
```
function openNewWindow(newURL) {
  location.replace(newURL);
}

```

Squidly271:
> When the page is called as /Apps/AddContainer it will load the docker translations, but AFTER the "AddContainer" page is loaded, hence the docker specific items are not translated.

OK that explains why on occasion it would appear translated but then quickly after it would switch itself to English.

Let me know when you do a PR for this change and I'll test and withdraw my PR at that time.

bergware:
PR is already done...

Squidly271:
OK.  That worked.

Now on another note,

The system doesn't think I have a cache drive installed (no pool on it), and each and every one of my containers also only see that there is only a mapping for appdata and NOTHING else when editing them.  Is it safe to assume that this is because I'm on beta1 with the new GUI?  (Can't say 100% if this started with the last merge, or if it was also on the merge I did the other day)

bergware:
Yes, you are running code which is only supported on the next beta version, which includes multi cache support. Wait for Limetech to release the next version or ignore for the time being.

Squidly271:
No worries.  Just noticed it, and wanted to confirm it wasn't a bug in your code 

Closing this PR

PR #624, opened 2020-03-29
swal style enhancement

bergware:


Squidly271:
Replacement looks a ton better!  Just have to now put in compatibility code on CA for it

image

Squidly271:
Are you also going to be updating the CSS on openBox() to match the themes, or is sticking with white for all themes the game plan?

bergware:
For the moment it stays white.
I would need to rethink the current implementation to make it theme dependent, too much effort att.

PR #626, opened 2020-03-30
Multi-language Support

Squidly271:
Remove any translations that attempt to inject code into the OS

### DIFF
--- a/plugins/dynamix/include/Translations.php
+++ b/plugins/dynamix/include/Translations.php
@@ -21,7 +21,7 @@ function _($text) {
   return strpos($data,'*')===false ? $data : preg_replace(['/\*\*(.+?)\*\*/','/\*(.+?)\*/'],['<b>$1</b>','<i>$1</i>'],$data);
 }
 function parse_lang_file($file) {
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
Does this protect against attack like so?

```
<b onmouseover=alert(‘XSS testing!‘)></b>
```

Squidly271:
Good point.  Thanks.  Now I've got to think about this mainly with regards to the application feed as formatting is allowed within descriptions which significantly complicates the situation.  And formatting is also allowed (and necessary) within translations

OmgImAlexis:
What I've seen is using something like this `_t(`The {0} {1} is ready for download`, red, 'ball')` for the translations where they need formatting and such.

Is there any other formatting stuff you'd need?

Squidly271:
Goes beyond that.  This isn't about the phrase to be translated.  Its about what the translated text is.

OmgImAlexis:
Whatever way we work it out it needs to be compatible with the way we’re going todo this in vue too.

OmgImAlexis:
Might be worth looking into http://htmlpurifier.org

PR #630, opened 2020-04-01
Allow markdown within container descriptions

Squidly271:
re: https://forums.unraid.net/topic/90433-dockerman-descriptions-etc-security/

Regardless if the final removal in the future is handled by the OS disallowing HTML or the appfeed itself enforcing it, authors still require a method to create and preview their descriptions within dockerMan


bergware:
Presence or absence of tags probably is not a good test indicator, because markdown content may include tags.

Squidly271:
Tags within the markdown are being specifically disallowed.  Otherwise there's no point in allowing Markdown as it wouldn't bring any security.  Whenever the system switches over to pure markdown, a strip_tags() will be done to remove them.

PR #640, opened 2020-04-04
Multi-language support

Squidly271:
Display aberration on Settings - Docker - Advanced View, VM xml view


bergware:
I must have fat fingers and like that space bar a bit too much ;)
Thanks for spotting

PR #645, opened 2020-04-09
Multi language support

bergware:


bergware:
Correct regression errors in the VM form templates

PR #650, opened 2020-04-11
Added support for private docker registries with basic auth or no auth, and improvements for token based authentication

rystaf:
This fixes the "Check for updates" button not working for images from private docker registries that don't use Bearer tokens for authentication. I tested this using the DockerRegistry container from Community Applications and Traefik providing the basic auth.

### DIFF
--- a/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -851,7 +851,7 @@ class DockerClient {
 		foreach ($this->getDockerJSON("/containers/json?all=1") as $ct) {
 			$info = $this->getContainerDetails($ct['Id']);
 			$c = [];
-			$c['Image']       = DockerUtil::ensureImageTag($ct['Image']);
+			$c['Image']       = DockerUtil::ensureImageTag($info['Config']['Image']);
 			$c['ImageId']     = $this->extractID($ct['ImageID']);
 			$c['Name']        = substr($info['Name'], 1);
 			$c['Status']      = $ct['Status'] ?: 'None';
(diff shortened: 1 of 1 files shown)


bergware:
I did a quick test to see if regular registry downloads still work, this seems okay.
Did you test this against the regular registy too?

One request: can you comment out the debug statement (I presume you used it for troubleshooting)

rystaf:
I tested it with public images on docker hub, but I did not test private docker hub images or public images on 3rd party registries that use bearer tokens. Turns out the original code did not work for me  at all. I just pushed a commit that fixes this. I tested it with a private docker hub image and a public image at dock.mau.dev/maubot/maubot. Also removed those debug calls.

SquareMesh:
@rystaf @bergware Just feedback to let you know that this fix worked for me.  I was previously encountering the Not Available error.  My private docker registers sits behind SSL (LetsEncrypt docker) and this fix works to allow changes to be detected and allow upgrade to be applied.  I have other dockers, some based on Unraid templates and some direct from docker hub public images no errors occurring.   Thanks for the fix.  Will monitor and report if any surprise issues occur.

SquareMesh:
Hi @limetech,

Any chance this can make next release?  Since applying fix locally it has been working with existing docker images aswell as private docker.

cc @rystaf @bergware

SquareMesh:
Hi @bergware ,

Sorry to hassle, just wondering if you are considering this for the next release?  I'm continuing to use the changes made by this pull request in my system and fix appears to be working for both private docker and public dockers.  Is there any further tests or changes you are looking for?  Let me know if I can assist with further testing.

Apologies for questions, just keen to see this progress.

cc @rystaf, @limetech

PR #665, opened 2020-05-02
Task Plugins: Providing Ability to define Display_Name

dorgan:
Gives the ability to provide a Display_Name for when displaying task menu bar as having spaces in the file name seems to mess things up further.
Screen Shot 2020-05-02 at 9 17 59 AM
Screen Shot 2020-05-02 at 9 17 48 AM

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -433,9 +433,9 @@ foreach ($tasks as $button) {
   echo "<div id='nav-item'";
   echo $task==$page ? " class='active'>" : ">";
   if (!isset($display_name)) {
-    echo "<a href='/$page' onclick='initab()'>$page</a></div>";
+    echo "<a href='/$page' onclick='initab()'>". _($page) ."</a></div>";
   } else {
-    echo "<a href='/$page' onclick='initab()'>$display_name</a></div>";
+    echo "<a href='/$page' onclick='initab()'>". _($display_name) ."</a></div>";
   }
 }
 unset($tasks);
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -426,9 +426,17 @@ $(function() {
 echo "<div id='menu'><div id='nav-block'><div id='nav-left'>";
 foreach ($tasks as $button) {
   $page = $button['name'];
+  $display_name = null;
+  if (isset($button['Display_Name'])) {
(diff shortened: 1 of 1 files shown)


bergware:
Why are there two properties "Name" and "Display_Name"? Neither is used by Unraid.

A single property "Name" sounds more logical, and makes a very small 1 line change.

dorgan:
@bergware isnt Name used for nesting of items in the menu, and is more of an ID, such as when its a tabbed interface or nesting different sections.  Thats my current understanding

bergware:
Page nesting is done using the "Type" property.
E.g. Type="xmenu", tells the parser that this page has one or more "children" attached to it.

Page tabbing is controlled automatically, or can be forced using the "Tabs" property.

bergware:
I propose you make your change like this

`echo "<a href='/$page' onclick='initab()'>"._($button['Name'] ?? $page)."</a></div>";`

Just a one line change with "$button['Name'] ??" inserted.
This allows you to overrule the automatic name generation for an item in the header by using the "Name" property in a page file.

bergware:
Scratching my memory here.

In the pre-dynamix days (Unraid v5) there used to be a "Name" property, but I can't remember how it was exactly used.

dorgan:
Maybe that's why I have it, there no official documentation so I've had to
reverse engineer alot.  I can make the proposed changes a little later.

dorgan:
@bergware name Vs Name is very confusing of property, is still think Display_Name might be appropriate here, thoughts?

bergware:
I don't think there is any confusion and "Name" is the most suitable solution.

When creating page files and setting the appropriate properties in the file, you won't have to take into consideration any of the internal properties (there are more internal properties).

dorgan:
@bergware the current assignment for $page is `$button['name']` do you not think it's confusing if my code just checks for `$button['Name']`?

dorgan:
@bergware sorry my server was down due to hardware issues, so didnt want to follow up if i couldnt make changes.  Are you OK moving forward with current code, being that `$button['namae']` is the current value assigned to $page, I think having code the then conditionally sets $page to `$button['Name']` is just confusing, but I am open to changes.

bergware:
Please, see my proposal a few posts up.

dorgan:
@bergware thats still just making use of $button['name'] and $button['Name'], where $button['name'] is the current value used today.

dorgan:
guess its silly to knit pick here, updated.

bergware:
Ok, thanks

PR #666, opened 2020-05-03
Miscellaneous updates

bergware:


Squidly271:
Seems like this might be a devilish PR, and a bad omen...

PR #671, opened 2020-05-06
Duplicate file....

dorgan:
scanning the code both plugins/dynamix/template.php and webGui/template.php both reference the /webgui/include/PageBuilder.php.  Honestly even template.php in both of these locations look like duplicates and could probably been consolidated down.


bergware:
/webGui is a symlink to /plugins/dynamix.

There are no duplicates

dorgan:
it doesnt look that way in the repo itself...

PR #675, opened 2020-05-09
Multi Language Updates

Squidly271:
The fix for user never going to settings prior to changing languages got lost somehow

### DIFF
--- a/plugins/dynamix/include/LanguageReset.php
+++ b/plugins/dynamix/include/LanguageReset.php
@@ -14,5 +14,7 @@
 $docroot = $docroot ?? $_SERVER['DOCUMENT_ROOT'] ?: "/usr/local/emhttp";
 
 $lang = $_POST['lang'] ?? '';
+if ( ! is_file("/boot/config/plugins/dynamix/dynamix.cfg") )
+  copy("$docroot/plugins/dynamix/default.cfg","/boot/config/plugins/dynamix/dynamix.cfg");
 exec("sed -ri 's/^(locale=\")[^\"]*/\\1$lang/' /boot/config/plugins/dynamix/dynamix.cfg 2>/dev/null");
 ?>


bergware:
Please revert.
It is not intended behaviour to copy the default file to flash.
The configuration implementation always checks for the presence of a file on flash and loads default if it doesn't exist.

I have solved the "issue" with 'sed' by suppression the error when the file is not present.

Squidly271:
> I have solved the "issue" with 'sed' by suppression the error when the file is not present

Suppressing the error doesn't solve the problem.  As we discovered with the problems of binhex-preclear, dynamix.cfg does not necessarily exist on the flash drive.  And I tested by doing a completely stock install of 6.8.3, upgrading to 6.9.0-beta9 without hitting any settings (and simply assigning drives), and at no point was dynamix.cfg created.

The issue is that if a user never changes Settings - Display (not an unlikely circumstance) and then goes to CA, and installs a language they will be unable to switch to it unless they go to Settings - Display

Squidly271:
ie: your switchLanguage("nl_NL") assumes that dynamix.cfg is already created.  Which on many user's system it is not.

bergware:
There are several configuration files which initially do not exist, and therefor the construction with "default.cfg" files was invented. Everything in the GUI works perfectly fine with that.

Switching language is indeed a special case. It should only set one setting (locale), copying the complete default file isn't the right approach, it is a workaround for only one case.

And sorry for binhex, but that is an oversight in his plugin. We do not guarantee that every cfg file exists.

Our discussion here, made me rethink the solution for language switching, thanks for that :)
I made a revised PR.

Squidly271:
No problems.  So long as it's fixed.

Can't help but comment though that my elegant solution to switching languages wouldn't have needed this fix implemented  :) 

Closing

bergware:
In the future this is "solved" for new users, when the Unraid Wizard is introduced.
This Wizard creates "dynamix.cfg" upon start up of a virgin system.

And as always we had a fruitful discussion :)

PR #684, opened 2020-05-14
Multi-language support

bergware:


bergware:
Ignore the first PRs and reverts. Trying to solve a non-existing error

PR #685, opened 2020-05-15
Multi-language support

bergware:


bergware:
Fixed: help text sections are not fully converted

bergware:
Proper handling of double quotes inside help text

PR #687, opened 2020-05-16
Multi-language support

bergware:


bergware:
Replaced single quotes for double quotes.
This to avoid issues when translations may have apostrophes.

PR #688, opened 2020-05-17
Multi-language support

bergware:


bergware:
More singles to doubles quotes to prevent potential translation issues

bergware:
Added language setting in html header.
<html lang="xx">

and added right-to-left display for Arabic and Farsi languages (when they become included).

bergware:
Fixed regression error

PR #690, opened 2020-05-18
Fix regression error

bergware:


bergware:
Went thru all the files and found two more regressions :(

PR #694, opened 2020-05-21
Use update image dialog for update entry in docker context menu

Tooa:
This pull request uses the docker update dialog also for the update entry in the context menu. This makes sure that users do not accidentally update their containers.

Reference: https://forums.unraid.net/bug-reports/stable-releases/docker-image-update-dialog-missing-cosmetic-issue-r941/


bergware:
Ok, thanks

PR #701, opened 2020-06-05
Multi-language support

bergware:
Another modification needed, discovered when using the French language!


bergware:
I got tired of making code corrections each time to support the French language.
To be more future proof I have deprecated the function my_lang() and consolidated its options in the function _(), while at the same time do automatic conversion of apostrophes to HTML code.

limetech:
we don't need no stinkin' my_lang

bergware:
After the tabs and semi-colons, it was about time to dump these suckers :)

PR #703, opened 2020-06-08
Update PluginHelpers.php

Squidly271:


bergware:
The function _() needs to be defined, because some scripts include PluginHelpers.php without translations and would fail otherwise.

Squidly271:
It gets defined via the require_once

PR #715, opened 2020-06-17
Multi-language support

bergware:


limetech:
Please let's not change the unit designations, eg, "KB" to "Ki".

In the storage industry since forever, "MB", "GB", "TB" always refer to million, billion, trillion, that is base 10 values.  The absence of 'B' means base 2**10.  This convention is followed many places, for example, 'dd' command.

bergware:
I find it very "unfortunate" that back then somebody at IBM *thought* it was a clever idea to call 1024 a kilo (it is not).

In the old days the relative difference between 1000 and 1024 was okay to accept (during my university time), but over time when numbers grew the discrepancy became bigger and bigger, and made the awkward difference more visible.

In Europe (and most other parts of the world) we are educated with the metric system. This means prefixes are always a multiple of 1000 in whatever context.

When I buy 1 kilo of coffee, it is 1000 gram.
When I drive 10 kilometer per hour, it is 10000 meter per hour
My LAN connection at home is 10 Gbps, this is 10000000000 bps

Back to Unraid...
Unraid has always displayed disk sizes in the metric system, this is conform to what disk manufacturers are doing, a 1 TB drive = 1000000000000 bytes. The same is true when the browser in the GUI shows file sizes (consistency). We should not change this!

I have removed the IEC names, see PR,  though personally I prefer these above the convention to omit 'B', but will accept it.

Translation of prefixes is needed for those languages which use different writing, such as Chinese or Arabic.

limetech:
> I have removed the IEC names, see PR, though personally I prefer these above the convention to omit 'B', but will accept it.

Alright, I've had a change of heart (a little grumpy yesterday).  I now agree that we should adopt the IEC suffix names.  It doesn't affect how Unraid uses KB, MB, TB, etc. because as you rightly point out those units are already base10.  But it does remove ambiguity in the UI to use, e.g., "GiB" instead of simply "G".  Since we have introduced multi-language into webGUI it is more important to remove ambiguity wherever possible.

PR #718, opened 2020-06-22
Multi-language support

bergware:


limetech:
When it comes to regular expressions, the hairier the better I always say .... ;)

PR #719, opened 2020-06-27
Multi-language support

bergware:


bergware:
Support for more languages using the Cleas Sans font

PR #725, opened 2020-07-10
Fixed: monitor may spinup disk

bergware:


limetech:
In monitor, why not just delete the line: "if (file_exists( .." ?
Also a similar line to remove in DeviceList.php

PR #733, opened 2020-07-23
VM manager: new setting "Network Model"

bergware:


limetech:
we don't need no stinkin' <?php tags!

PR #739, opened 2020-07-31
Multi-language support

bergware:
Fixed references in Dashboard


bergware:
Doc says $var['**regBuildDate**'] but implementation is $var['**regBuildTime**']
Changed _template.php_ to reflect implementation.

PR #746, opened 2020-09-19
Support links on notifications

Squidly271:
Allow notifications to link to pages in the GUI.

Pet Peeve: Clicking on the browser notification doesn't do anything.  This allows for clicking on the notification to actually take you to where you need to be to handle the notification.

### DIFF
--- a/plugins/dynamix.apcupsd/apcupsd.notify
+++ b/plugins/dynamix.apcupsd/apcupsd.notify
@@ -2,4 +2,4 @@
 # Send system notify message from apcupsd
 #
 read MESSAGE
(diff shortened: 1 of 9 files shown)


limetech:
Nice feature, thanks!

Squidly271:
lol  In notepad++, 2 spaces.  In GitHub 4 spaces.  It lets me know when I've done the styling mistake.

One of these days I'm going to do a PR against the whole webGUI and switch it to tabs.  Not that I would ever expect it to be merged, but I can't keep complaining about it if I don't at least make the attempt to fix it....  

I just can't get over it.

Squidly271:
> Why not set a editorconfig so you’ll can use whatever you want and it’ll normalise on save.
> 
> https://editorconfig.org/

TBH, I don't see how what amounts to hiding the spaces will fix either the problem, or the enjoyment I get from complaining about them...

PR #750, opened 2020-09-26
Add customizable timeout when stopping containers

Squidly271:
https://forums.unraid.net/topic/94360-customizable-docker-stop-timeout/


limetech:
Thx for the ref to helptext.txt PR

PR #758, opened 2020-10-20
Fix/j growl notification styles

zackspear:
Unfortunately the introduction of `#menu {…z-index:99999;}` introduced a few bugs – like with the Shadowboxes and other various elements that should have z-index priority. 

These suggested top value changes prevent `jGrowl` from overlapping the #menu DOM element.

There's another bug here with `jGrowl` creating an empty `<div class="jGrowl-notification"></div>`. Regardless these changes are still needed.


Squidly271:
Thanks.  I never run the OS on a small resolution, so never thought of openBox etc overlapping the menus...

PR #788, opened 2021-02-16
Switch Diagnostics to web socket

Squidly271:
Switches diagnostics to run via a Web Socket

- Will eliminate the fairly common timeouts prevalent, especially with wide arrays and drives spun down
- If a command crashes, it's helpful to the forum to know exactly where it hung up
- Even if the web socket isn't working (browser issues), the file is now also saved to the flash drive.
- Safeguards in place to handle if the user runs multiple diagnostics at the same time.

### DIFF
--- a/plugins/dynamix/Diagnostics.page
+++ b/plugins/dynamix/Diagnostics.page
@@ -4,8 +4,8 @@ Icon="icon-diagnostics"
 Tag="tv"
 ---
 <?PHP
-/* Copyright 2005-2020, Lime Technology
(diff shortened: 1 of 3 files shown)


ljm42:
The `at` command is sending an empty 'Output from your job 9' email to root

Squidly271:
Thanks

ljm42:
No more email :)  This is a great improvement to diagnostics, thanks @Squidly271!

limetech:
awesome, thanks!

PR #789, opened 2021-02-17
SysDevs - fix warn if leave page without saving

ljm42:
Prevent incorrect warnings on systems with no IOMMU groups

### DIFF
--- a/plugins/dynamix/SysDevs.page
+++ b/plugins/dynamix/SysDevs.page
@@ -24,7 +24,7 @@ table tr td.thin{line-height:8px;height:8px}
 <script>
 $(window).bind('beforeunload',function(e){
   // warn user if they leave the page with unsaved changes
-  if (!$('#applycfg').prop('disabled')) return "You have unsaved changes";
+  if ($('#applycfg') && $('#applycfg').is(":visible") && !$('#applycfg').prop('disabled')) return "You have unsaved changes";
 });
 $(function(){
   $('#t1').load('/webGui/include/SysDevs.php',{table:'t1'});


plugins/dynamix/SysDevs.page
```diff
@@ -24,7 +24,7 @@ table tr td.thin{line-height:8px;height:8px}
 <script>
 $(window).bind('beforeunload',function(e){
   // warn user if they leave the page with unsaved changes
-  if (!$('#applycfg').prop('disabled')) return "You have unsaved changes";
+  if ($('#applycfg') && $('#applycfg').is(":visible") && !$('#applycfg').prop('disabled')) return "You have unsaved changes";
```
Squidly271, line 27: Should be "_(You have unsaved changes)_"

ljm42: It actually doesn't matter what the text is, browsers will ignore it and display their own message

ljm42: For instance, Chrome says "Leave site? Changes you made may not be saved." and Firefox says "This page is asking you to confirm that you want to leave - data you have entered may not be saved." They do this regardless of what message you return here.

Squidly271: Ah.  Perfect.  Learn something new every day.  Why not do this in DefaultPageLayout so that every page works the same way?

ljm42: This page already has logic that enables the "Bind Selected" button when the form is "dirty", so I am using the enabled/disabled status of that button to control whether the "save changes before leaving?" message is shown. 

We don't have a global way of determining whether any given form is "dirty", so the "dirty/clean" logic needs to be added individually to the forms/pages that need it.

It could potentially be added to every form, but It is particularly important here because the form is so long people could (un)check a box and not realize they also need to press the "Bind Selected" button for it to take effect. I believe that is what happened here: https://forums.unraid.net/topic/100365-iommu-networkdevice-issues-690-rc1/?tab=comments#comment-926351

This PR is to make sure that systems which don't have IOMMU (and therefore don't have a visible "Bind selected" button) will never be prompted to save changes. Oversight from my previous PR for this.

PR #790, opened 2021-02-17
Fix: properly set samesite cookie

eschultz:
Logging in with Safari works again 🥞


OmgImAlexis (approved):
Can confirm this now allows me to login with Safari.

PR #791, opened 2021-02-20
Add Progress for Mover

SimonFair:
Please find attached changes to allow Mover progress to show on main screen.

I will send mover script via a DM.

Progress is only updated per share as any more granular reporting will need move  to be updated.

Main screen Move button and status will change if the schedule or the mover is run outside of the main page.


SimonFair:
Will need to be reworked due to move to nchan.

PR #795, opened 2021-03-06
Diagnostics: include modprobe.d files

ljm42:


eschultz (approved):

ljm42:
Hmm... will the modprobe.d/*.conf files always be empty? If so, a simpler solution would be to add `/boot/config/modprobe.d` to the $folders array here: https://github.com/limetech/webgui/blob/056ded89021bdf924b082412c8e9d71c3269220c/plugins/dynamix/scripts/diagnostics#L23  and then the list of conf files will be placed in system/folders.txt

PR #796, opened 2021-03-06
Diagnostics: skip empty log files

ljm42:


eschultz (approved):

limetech:
This changes causes information to be lost.  That is, the log files are only created when that subsystem is started.  For example, docker.log is only created if Docker is enabled.  If the size of this file is 0 it just means nothing logged for Docker.  If you omit the file from diags then you can't tell if log is omitted because Docker not started, or simply Docker didn't log anything.  I say just leave 0 length files in there, causes no harm.

ljm42:
good point, let's drop this

PR #797, opened 2021-03-06
Diagnostics: include go as go.txt

ljm42:
Easier to view go script without having to extract it first


eschultz (approved):
LGTM.  good idea on making it easier to examine the go file with a .txt extension.

PR #798, opened 2021-03-07
Protect GUI from invalid .page files

Squidly271:
If the format of the .page is completely wrong, then every page gets corrupted, and the system becomes unusable.  Personally, I hate it that every .page file can potentially take down the entire GUI in case of an error.  Blocking the output of the parse_ini_file protects the GUI, and the file gets logged to inform the user / forum.

Realistically, a released plugin should never trigger this code, but it's really annoying when it does (https://forums.unraid.net/bug-reports/stable-releases/gui-partially-didnt-working-r1327/) as it's not obvious anywhere what plugin is causing it.


limetech:
Good idea!

PR #802, opened 2021-03-09
Fix links on notifications not being present if in summary display

Squidly271:


ljm42 (approved):
works great! notifications in "summarized" view are now clickable too

PR #808, opened 2021-03-14
Keep GUI in sync with diagnostics spinning up drives

Squidly271:
If there's a better way of doing it, feel free to not merge

https://forums.unraid.net/bug-reports/stable-releases/diagnostics-spins-up-drives-but-show-down-in-array-r1363/

### DIFF
--- a/plugins/dynamix/scripts/diagnostics
+++ b/plugins/dynamix/scripts/diagnostics
@@ -22,6 +22,17 @@ $var = (array)@parse_ini_file("$get/var.ini");
 $docroot = $docroot ?? $_SERVER['DOCUMENT_ROOT'] ?: '/usr/local/emhttp';
 $folders = ['/boot','/boot/config','/boot/config/plugins','/boot/extra','/boot/syslinux','/var/log','/var/log/plugins','/var/log/packages','/tmp'];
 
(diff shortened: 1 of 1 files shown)


limetech:
Instead of looping over disks you can use

cmdSpinupAll=apply

api call.

Squidly271:
Thanks.  As an aside, that's not how the GUI works.  It calls ToggleState on spin up / spin up all and loops through the drives.

limetech:
> As an aside, that's not how the GUI works. It calls ToggleState.php on spin up / spin down all and loops through the drives, hence the original coding.

Looking at code looks like ToggleState.php does take into account spin-up/down All and invoke either cmdSpinupAll or cmdSpindownAll.  For a group this really should be using spin-up groups now since that code works for array and pools as of 6.9 release.

Squidly271:
I traced the code, and it always seems to spin individually.  But, in retrospect that may be because on the array the spinup only spins up the array, etc.

limetech:
cmSpinupAll will spin up all devices: those in array, pools, and unassigned.

Squidly271:
Looks like this was fixed via different manner in 6.9.2  closing

PR #813, opened 2021-03-18
Make diagnostics messages left aligned

bergware:


limetech:
Suggestion: consider changing name of that javascript sprintf() function to something more specific to how it's being used.  This is because "sprintf" is a common function name, eg in C, php, etc. and might confuse someone looking at code later ;)

PR #814, opened 2021-03-20
Prevent autofilling of passwords on containers/vms

Squidly271:
Chrome / Firefox have very strange results when autofilling on containers:  Fixed IP will get "root" and the first password entry will have the root password.

Editing VMs have a share filled out of "root" and the VNC password of root password.


limetech:
thx!  should help

PR #819, opened 2021-03-30
Do not flag mover messages

bergware:


limetech:
re: Allow dot in user name

This might break something.  Refer to first answer here:
https://stackoverflow.com/questions/6949667/what-are-the-real-rules-for-linux-usernames-on-centos-6-and-rhel-6

Was there a request by someone to allow this?

bergware:
_Was there a request by someone to allow this?_

Yes, see https://forums.unraid.net/topic/104434-change-adduser-restriction/?tab=comments#comment-964343

If not sure, I can revert this change.

bergware:
Ok, I reverted this PR. Better safe than sorry!

limetech:
thx.  Yes I think we can support this but it's something I'd rather introduce during a beta/rc series rather than add to stable.

PR #821, opened 2021-04-10
Add reference to internal docker path / variable / port

Squidly271:
Continuing problem over the years where the description (if the maintainer does not fill anything in) bears absolutely no relationship to what the internal path / port actually is.   (And if the container path is edited, but the description isn't then the screen still shows the original path) Most recently reflected here: https://forums.unraid.net/bug-reports/stable-releases/692-changing-container-paths-not-reflected-in-gui-r1415/, but this has been going on forever, and is very confusing for the users and people trying to help them.

This eliminates it by never automatically filling in the description when adding a path / port / variable, and always displays on the edit template screen what the current internal setting is.

image

This will require (at time of merging) for me to issue a script which will remove the pre-existing descriptions from the installed templates if they are of the form "Container Port: 80" so that duplicate lines do not appear on the screen.

As a bonus, it also lessens the need to ever post the docker run command (which is a pain to get for new users) as what shows up on the screen is a reflection of what is actually going to be present in the command

This is more transparent and more secure for the user.


bergware:
Haha, right!
This is a nice solution, definitely an improvement.

Only thing, it makes me realize that translation is not done correctly here (and previously) and need to modify that.

_(Container)_ {5}: {1}

This becomes for example **Container Path: xxxxx**
Here we need to translate the complete phrase, because certain languages use different syntax, e.g. french: _Chemin du container_

I can solve this by introducing a new variable {12} which is the translated text, it means your PR needs a small change:

{12}: {1}

I will then follow up with the translation additions to support this.

Squidly271:
Done

Squidly271:
Actually @bergware , wouldn't it just be easier to do
_(Container {5})_ 
Instead of creating a new variable and then adding Container Path, Container Port etc to the translations?

bergware:
I made my PR to do the revised translation based on your PR.

Regarding your question: **(container {5})** doesn't work because the PHP translation happens before the javascript variable substitution. In other words {5} isn't substituted in time. The new approach does all translations in javascript.

Squidly271:
Here's the script (run via Install Plugin).  Doesn't actually install anything though.  Just a simple way to run a script for end user

https://raw.githubusercontent.com/Squidly271/misc-stuff/master/fix_description.plg

Doesn't handle multi-language, but I think that's a non-problem since most people wouldn't ever edit the default descriptions to begin with.

CA would have to on the next OS also begin to remove the descriptions being passed through.

Squidly271:
Updates done for CA

PR #824, opened 2021-04-16
Include links in email and Discord agent notifications

ljm42:
Users with the Discord agent currently installed will need to go to the Notification Agents page, make a small change to the Discord agent, and hit Apply to regenerate the script.

Note: When clicking links from web-based email clients you will likely be redirected to the login screen due to SameSite cookie restrictions. Links can be copy/pasted to the address bar instead.


ljm42:
Links are now available to all agents. On the Notification Agents page, when configuring the "Notification Message" for a given agent, tick the box for "Link" and hit apply.

PR #825, opened 2021-04-18
Validate WebGUI ports before applying

Squidly271:
@bergware, since we are starting to use more and more sprintf within the UI, can I suggest that we incorporate the sprintf library from CA (javascript/libraries.js) as it's a complete implementation

### DIFF
--- a/plugins/dynamix/ManagementAccess.page
+++ b/plugins/dynamix/ManagementAccess.page
@@ -5,8 +5,8 @@ Icon="icon-key"
 Tag="expeditedssl"
 ---
 <?PHP
-/* Copyright 2005-2020, Lime Technology
- * Copyright 2012-2020, Bergware International.
+/* Copyright 2005-2021, Lime Technology
+ * Copyright 2012-2021, Bergware International.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License version 2,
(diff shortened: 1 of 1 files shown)


bergware:
In the swal function, please add (for translation purposes)

`confirmButtonText:"_(OK)_"`

bergware:
For Unraid 6.10 we can include the javascript library to support sprintf. This will be included in dynamix.js, so it is available for all pages in the GUI (and do some rewrites of existing solutions).

To avoid conflicts though, you'll need to remove it from CA.

Squidly271:
Yeah, didn't even think about the OK translation.  Thanks.  Yeah, if / when you include the library in dynamix.js I'll conditionally remove it from CA.  

Just since we keep on using simplef / whatever else we call it, we might as well include the complete library (which CA does require, as simplef doesn't handle order transposition of the arguments)

bergware:
Make list of ports in use more selective
1. Check on IPv4 only (otherwise IPv6 entries will interfere)
2. Only include ports which listen to all interfaces or to the host address

This is the revised code

```
// Get ports in use by host
exec("lsof -Pni4|awk '/LISTEN/{print \$9}'", $output);
foreach ($output as $line) {
  [$ip, $port] = explode(':', $line);
  if (!in_array($port, $portsInUse) && ($ip=='*' || $ip==$eth0['IPADDR:0'])) $portsInUse[] = $port;
}

```

bergware:
Sorry, another update (after carefully rethinking what we need)

```
exec("lsof -Pni4|awk '/LISTEN/{print \$9}'", $output);

$bind = $var['BIND_MGT']=='yes';
$list = ['*', $eth0['IPADDR:0']];

foreach ($output as $line) {
  [$ip, $port] = explode(':', $line);
  if (!in_array($port, $portsInUse) && (!$bind || in_array($ip, $list)) $portsInUse[] = $port;
}
```

PR #826, opened 2021-04-24
Add vmxnet3 and e1000 into available NICs for VMs

SimonFair:


bergware:
A couple of things with your PR, which prevents it from pulling

1. It alters the default choice (virtio-net), new choices must be at the end of the list
2. New choices should also be made available in the template used for adding network interfaces (user clicks on the + button)
3. New choices should also be made available for the ELEC templates

Squidly271:
> New choices should also be made available for the ELEC templates

Maybe it's time to remove those altogether?  They're 2 major versions behind

SimonFair:
> A couple of things with your PR, which prevents it from pulling
> 
> 1. It alters the default choice (virtio-net), new choices must be at the end of the list
> 2. New choices should also be made available in the template used for adding network interfaces (user clicks on the + button)
> 3. New choices should also be made available for the ELEC templates

Thanks for feedback will review.

bergware:
> Maybe it's time to remove those altogether?

AFAIK there are people using these templates, albeit not many.

SimonFair:
> > Maybe it's time to remove those altogether?
> 
> AFAIK there are people using these templates, albeit not many.

I didnt change those ones as people would be unlikely to change from defaults.

Revised order and added 2nd NIC options.

SimonFair:
Defaults are not set by the list order, should it be alphabetical in the list?

bergware:
I prefer as you have done in your last PR (no need to make changes).

virtio-net should be the first choice with virtio as second choice, these are the recommended general choices.

SimonFair:
> I prefer as you have done in your last PR (no need to make changes).
> 
> virtio-net should be the first choice with virtio as second choice, these are the recommended general choices.

Ok thanks, I think the ELEC should be left with the virtio only NICs, as they OS works with them.

PR #836, opened 2021-05-08
Update alert text

Squidly271:
re: pr #829


bergware:
We should use "swal()" instead of "alert()".

Squidly271:
Completely agree.  Problem is that it's running within a modal, and is a pain to bring up the swal within the modal properly without investing a fair amount of time for what's going to be an error message that will pop up 1 in a million times.

Just not worth it.

PR #841, opened 2021-05-12
Nchan start/stop foundation

bergware:


plugins/dynamix/ArrayOperation.page
```diff
@@ -276,15 +276,15 @@ devices.on('message', function(data) {
     if ($('#'+table[0]).length) $('#'+table[0]).html(table[1]);
   }
   display_diskio();
+<?if ($var['fsState']=='Stopped'):?>
+  setTimeout(function(){devices.stop();},0);
```
OmgImAlexis, line 280: Just at tip, `setImmediate` would be a little quicker when using `0` as the time since `setTimeout` still needs to lookup the timer on run unlike `setImmediate`.

PR #842, opened 2021-05-13
Rename Unraid.net to My Servers

ljm42:


bergware:
Translation is broken.
Corrections were made to the translated text in DefaultPageLayout.php. Can you port these modifications to myservers1.php and myservers2.php?

ljm42:
Thanks you for taking the time to review those translations earlier, sorry I didn't realize it. They have been updated

PR #843, opened 2021-05-13
Nchan start/stop foundation

bergware:
Some delay needed, brrr.


bergware:
Testing on an iPad-Pro and page display is inconsistent. A 1 second delay makes it consistent.

PR #852, opened 2021-05-15
feat(docker): process bash ANSI colors in web log display

nysos3:
This adds support for properly displaying the colors output by some docker containers' logs.


bergware:
Do you have an example of a container which uses ANSI codes in its log?

OmgImAlexis:
Can you please an example of how it looks on a white background as the standard is based on a black background and that may cause issues here.

nysos3:
tdarr and tdarr_node both have ANSI codes in their log.

image

image

limetech:
Merged, thank you!

PR #870, opened 2021-05-28
Dashboard: fix context menu display

bergware:


Squidly271:
Does this fix when the context menu is at the very bottom and you can't acknowledge anything on it?

bergware:
Yes, it creates more space at the bottom of the table to allow the display of the complete context menu.

PR #874, opened 2021-06-02
Initialize variables properly with array values

bergware:


limetech:
my_explode('neck',$body);

PR #875, opened 2021-06-03
Initialize variables properly with array values

bergware:


bergware:
Next to **my_explode** there is **my_preg_split** :)

plugins/dynamix/include/Translations.php
```diff
@@ -36,7 +36,7 @@ function _($text, $do=-1) {
     foreach ($keys as $key) if (isset($language[$key])) $text = preg_replace("/\b$key\b/",$language[$key],$text);
     return $text;
   case 3: // device translation
-    [$p1,$p2] = preg_split('/(?<=[a-z])(?= ?[0-9]+)/i',$text);
+    [$p1,$p2] = array_pad(preg_split('/(?<=[a-z])(?= ?[0-9]+)/i',$text),2,'');
```
ljm42, line 39: should this use my_preg_split?

bergware: No, this is correct.
Translations.php can not depend on inclusion of Helpers.php and need to do its own coding

PR #878, opened 2021-06-15
fix: upc char encoding issue

zackspear:
AWS Amplify builds we not correctly setting charset (reason still unknown) – ex: `•` became `â€¢` and a few others.

Built the JS locally to circumvent issue.


OmgImAlexis (approved):
LGTM

PR #884, opened 2021-06-29
fix: #nav-item to .nav-item

OmgImAlexis:
Replaces https://github.com/limetech/webgui/pull/548

### DIFF
--- a/plugins/dynamix/HelpButton.page
+++ b/plugins/dynamix/HelpButton.page
@@ -17,7 +17,7 @@ Code="e934"
 ?>
 <script>
 function HelpButton() {
-  if ($('#nav-item.HelpButton').toggleClass('active').hasClass('active')) {
+  if ($('.nav-item.HelpButton').toggleClass('active').hasClass('active')) {
     $('.inline_help').show('slow');
     $.cookie('help','help',{path:'/'});
   } else {
--- a/plugins/dynamix/LanguageButton.page
+++ b/plugins/dynamix/LanguageButton.page
@@ -18,7 +18,7 @@ Code="e982"
 <script>
 <?if (count(glob('/var/log/plugins/lang-*.xml'))==0):?>
 // hide switch button when no other language packs
-$(function(){$('#nav-item.LanguageButton').hide();});
+$(function(){$('.nav-item.LanguageButton').hide();});
 <?endif;?>
 
 function LanguageButton() {
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -49,8 +49,8 @@ html{font-size:<?=$display['font']?>}
 #header{background-color:#<?=$backgnd?>}
 <?if ($themes1):?>
 #menu{background-color:#<?=$backgnd?>}
-#nav-block #nav-item a{color:#<?=$header?>}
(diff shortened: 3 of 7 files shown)


OmgImAlexis:
@Squidly271 if you have any plugins relying on this they need to be updated to use the class.

If you need to support older installs you'll need to add a fallback to the ID if it can't find the element.

Squidly271:
> if you have any plugins relying on this they need to be updated to use the class.

If and when it's merged

OmgImAlexis:
@Squidly271 is there a reason you can't add this before that? This **WILL** be getting merged as it's incorrect to have multiple items using the same ID.

Squidly271:
Because I don't know what version of the OS this will get included on.   Maybe I'm gun shy, but the last PR sat forever...

OmgImAlexis:
It mainly sat idle as I was waiting for plugins to get their fallbacks added. You should be able to use `.nav-item, #nav-item` in most cases for CSS. Query selectors shouldn't be much different.

In saying that I'm not even sure if anyone was using query selectors since I'm not sure this would return the correct element because of the duplicates IDs. 🤔

Squidly271:
> It mainly sat idle as I was waiting for plugins to get their fallbacks added.

Then you also have to go after @bergware.  There's (AFAIK) 3 places this is used.

- Dynamix System Buttons (without update this is a breaking change)
- GUI Search (without update this is a breaking change)
- CA (only used to replace the switch language function with something that works slightly better in CA)

OmgImAlexis:
@Squidly271 thanks for the update. 👍 

@bergware any chance I could get an update here?

bergware:
I haven't really looked into this (it is not causing a problem).

The consequence of this change is that plugins need to be updated too, and to maintain backward compatibility, these plugins need to deal with both situations. I don't know about plugins of other authors though.

OmgImAlexis:
> (it is not causing a problem).

Yes it is. Screen-readers don't work with this. This should never have been added like this as it's not even valid HTML.

> The consequence of this change is that plugins need to be updated too, and to maintain backward compatibility, these plugins need to deal with both situations. I don't know about plugins of other authors though.

Okay? It's going to have to happen eventually.

bergware:
I prepared an update for the system-buttons plugin, which can deal with the revised classification (compatibility between new and old).

Waiting for inclusion of this PR in 6.10 though

Squidly271:
There's more than just #nav-item.  Notification Settings - Notification Agents has multiple #slot_1 and #slot_2

(CA's next release now has "alt" on images and fixed a missing <ul> to comply with WCAG 1.x)

OmgImAlexis:
At some point we're going to have to introduce automated checks for this repo and plugins for CA that checks things like this before allowing them to be merged.

Squidly271:
If you're curious, support just had a ticket asking for the WebUI to be more usable for screen readers.

zackspear:
Can we please consider this PR again? We keep pushing and pushing but it's a real issue for accessibility as it's incorrect HTML.
If we're ok breaking functionality with other PRs that add things (https://github.com/limetech/webgui/pull/1226) then why not this PR that fixes?

bergware:
I have included these changes in my PR. This request can be closed

zackspear:
Thank you @bergware 🙏

limetech:
These long overdue changes were accomplished by @bergware in a separate PR.

PR #886, opened 2021-06-29
Fix/box shadow

OmgImAlexis:
box-shadow cannot have a 0 value. Instead it should be set to none.


zackspear:
`none` is the default value and interpreted as `0`.

https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow

PR #888, opened 2021-06-29
upc translation formatting & js updated

zackspear:


limetech:
Instead of string /boot/config you should simply say 'config', as in:

_('The key file should be located in the "config" directory on your USB Flash boot device')

I don't know about the quotes around "config" - would be good to have them or else display that word in monospace - not sure how to do that in a translated string.

The reason for this: when the USB flash is plugged into a PC you don't see a 'boot' diretory, you just see a 'config' directory.  Similar if you access the 'flash' share on the network - no 'boot' directory.

'/boot' is the name of the mount point on which the flash partition 1 is mounted.  So if you are referencing via Terminal window then you would reference '/boot/config' - but most people don't do this, and those that do, know what's happening.

zackspear:
10-4. Will run through all the copy and update `/boot/config` to `config`.

I'll attempt to emphasize "config" either with _italics_ or `<code>` if the translation setup allows for it. Because we don't ship translations.txt in `/usr/local/emhttp/languages` by default I'm not sure it'll be possible.

bergware:
For translations you can do something like this:

_('The key file should be located in the *config* directory on your USB Flash boot device')

This would display _config_ in italics
The translation file will have an entry like this:

The key file should be located in the config directory on your USB Flash boot device=The key file should be located in the *config* directory on your USB Flash boot device

zackspear:
Thanks @bergware for confirming that. Was going to try it but wasn't 100% certain on it working.

@limetech copy updated in latest commit per request.

PR #906, opened 2021-07-25
When viewing source, identify which .page file is responsible

Squidly271:
Up to @bergware.  As an example, I had no idea that UD injected code into every page of the UI.

### DIFF
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -534,6 +534,7 @@ unset($pages,$page,$pgs,$pg,$icon,$nchan,$running,$start,$stop,$row,$script,$opt
 <iframe id="progressFrame" name="progressFrame" frameborder="0"></iframe>
 <?
 // Build footer
+echo "\n<!--\n".str_repeat("#",30)."\n Footer\n".str_repeat("#",30)."\n-->\n";
 echo '<div id="footer"><span id="statusraid"><span id="statusbar">';
 $progress = ($var['fsProgress']!='')? "&bullet;<span class='blue strong'>{$var['fsProgress']}</span>" : '';
 switch ($var['fsState']) {
--- a/plugins/dynamix/include/DefaultPageLayout.php
+++ b/plugins/dynamix/include/DefaultPageLayout.php
@@ -421,7 +421,7 @@ if ($display['usage']) my_usage();
 echo "</div>";
 echo "<div id='nav-right'>";
 foreach ($buttons as $button) {
-  eval("?> \n<!--\n".str_repeat("#",30)."\n Button: {$button['file']}\n".str_repeat("#",30)."\n-->\n");
+  echo "\n<!--\n".str_repeat("#",30)."\n {$button['file']}\n".str_repeat("#",30)."\n-->\n";
   eval('?>'.parse_text($button['text']));
   if (empty($button['Link'])) {
     $icon = $button['Icon'];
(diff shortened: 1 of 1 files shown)


ljm42:
This is very enlightening! I'd like to go one step further and have it output "button:" or "nchan:" at the beginning of the line:
  button: {$button['file']}
and:
  nchan: {$page['file']}
so then we can search the source for "button:" or "nchan:" to find all of the instances

Squidly271:
No point in labelling "nchan" as the PR has nothing to do with nchan at all (you're just seeing the line above).  Labelling something as "Button"? sure.  Not that it's necessarily a "button".  Just that the .page file is *technically* an invisible button used to inject code to every page in the UI

bergware:
It seems like adding debug information, is there a specific reason why you want to see which page is adding code.
Most "buttons" are part of the stock GUI going, only a few plugins make use of this mechanism to add code.

bergware:
The GUI is built dynamically, this means only those pages which are needed for a particular section are loaded.
In other words you will get different results depending on what part of the GUI you are viewing.

Did it help you in identifying a problem?

Squidly271:
The PR just happened to have popped into my mind.  It costs absolutely nothing, and since the source is built dynamically, it would be highly beneficial to know which .page file is loading which code, and would also make it easier to find any particular code fragment when doing dev on the GUI when debugging code that is being conditionally added. since you can now jump easily to the .page file to see how it worked.

bergware:
I see, yes for development purposes this may be handy. It allows you to see where sections are coming from
(though for me I know them pretty well by heart)

For troubleshooting purposes I think it is less meaningful, I don't expect users to inspect source code. And in case the GUI isn't acting as expected, the first step should be: test in safe mode without plugins.

Squidly271:
Completely agree.   This is a dev tool only.   Very rarely would an end user be looking at the source.

bergware:
"eval" is a costly operation in PHP and only needed when code execution is in play, it is better to use "echo" to insert the statements.

  echo("\n\n");
..
..
  echo("\n\n");

ljm42:
> No point in labelling "nchan" as the PR has nothing to do with nchan at all

Oh right, wasn't looking closely enough.  How about "page: " then? So you can quickly find which page(s) were involved in rendering this.

I often have a hard time identifying which .page file(s) are used to render a page. For instance, the .page files that build Network Settings are not called Network.

Squidly271:
> I often have a hard time identifying which .page file(s) are used to render a page. For instance, the .page files that build Network Settings are not called Network.

I'm confused, because even without the identifier your want, this is exactly what this PR does.

ljm42:
Yes just saying in general I like this PR :)

bergware:
The header section at the top of a .page file, tells where in the GUI this page should be included.

Unfortunately we don't have a nice write up of the statements which can be used in the header section, nor a description of the dynamic construction of the GUI and how pages are interlinked.

Sounds like a task to add ... ;)

Squidly271:
> Sounds like a task to add ... ;)

Tom said he was doing this back in 2014 https://forums.unraid.net/topic/33322-how-does-the-plugin-system-work-documentation-added-wip/?tab=comments#comment-321212

Maybe it got forgotten? ;)

bergware:
_> Maybe it got forgotten? ;)_

I came across a document called **Plugin design rules** which I wrote in November 2014. Outdated and incomplete, but apparently I felt something was needed. Btw document was never published.

Squidly271:
Would've been nice way back when....  Of course nowadays everyone looks at either the dynamix plugins or mine, and we both still answer questions on how things get done.

PR #907, opened 2021-07-27
System devices additions

SimonFair:
Please find PR for addtiton of the physical port to usb lists in SysDevs,

ljm42 has already seen a preview, but let me know any feedback.


bergware:
Two minor changes to consider.
1. I would preset the value to "None" at the start and remove the 'else' statements
2. Multi-language support needs to be included
`  $physical_busid = _("None");`

PR #916, opened 2021-08-05
dockerMan: Remove code for template repositories

Squidly271:
Don't hold up an RC over this

### DIFF
--- a/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -27,7 +27,6 @@ require_once "$docroot/plugins/dynamix.docker.manager/include/Helpers.php";
 $dockerManPaths = [
 	'autostart-file' => "/var/lib/docker/unraid-autostart",
 	'update-status'  => "/var/lib/docker/unraid-update-status.json",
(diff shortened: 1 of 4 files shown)


bergware:
When you remove the download function, it will break the Docker update logic in the GUI.
We should leave that in.

Squidly271:
Yup, there is a single instance where it's used that I missed within DockerTemplates.  Thanks

bergware:
The function "downloadTemplates" is used by the GUI to retrieve the latest versions when clicking on "Check for Updates".
This won't work anymore after your PR, I believe.

I am not against code cleanup, but I think this one needs a bit more testing to understand the consequences.

Squidly271:
Yes the GUI did call that routine.  Hence why it was deleted ln the original PR.  I did miss a reference in the background schedule (dockerupdate) after your first post

I've been running this code since 8/1 on a production server.  If you want to test it out yourself, then install it.

It doesn't need to be merged anytime soon as it is simply a cleanup and doesn't affect functionality at all.

limetech:
Is this good to merge?

Squidly271:
Leave it until @bergware tests it

Squidly271:
Let's leave the code in for a while.   Closing this out.

PR #919, opened 2021-08-08
Set tabbed mode as default

Squidly271:
Probably @bergware will disagree, but tabbed mode cleans up the GUI significantly, and will make it look like it's had a complete makeover.  (But, this is also my personal preference. )


bergware:
Initially Dynamix knew only tabbed mode, but when it became integrated with Unraid, it was the explicit wish of Tom to support the original non-tabbed mode too and make this default. This would make the GUI appearance similar to what existing users are familiar with, but they have the opportunity to change it to their liking.

I am not sure we should change the default, and leave it up to Tom to decide.

limetech:
Yeah let's hold off on this one :)

PR #922, opened 2021-08-11
Docker updates

bergware:


Squidly271:
Hold off for a couple days on merging if possible until CA is updated to reflect these changes.  It's a breaking change across the board

PR #925, opened 2021-08-12
Add HOST_HOSTNAME and HOST_CONTAINERNAME env vars

TheFlyingBadger:
add extra environment variables to the container


OmgImAlexis:
Any reason for these changes?

TheFlyingBadger:
It was prompted because I wanted to set my window title better in the Firefox container ("Guacamole Client" wasn't cutting it for me). I did a sed in the container to set it to something better then I realised that using env vars would be better. I've always had a bit of grizzle to myself about containers not being "self-aware" about what they really were - so the automatic population of these vars would be a "blue pill" moment for our Unraid containers!

bergware:
Your new entries should follow the same construct as the existing entries:

```
$Variables[] = 'name="value"'

$Variables[] = 'HOST_HOSTNAME="'.$HOSTNAME.'"';
$Variables[] = 'HOST_CONTAINERNAME="'.$xml['Name'].'"';

```

PR #938, opened 2021-08-19

bergware:


plugins/dynamix/include/FileTree.php
```diff
@@ -24,70 +25,38 @@
...
+      $htmlRel  = htmlspecialchars($rootdir.$dir);
+      $htmlName = htmlspecialchars(mb_strlen($dir)<=33 ? $dir : mb_substr($dir,0,33).'...');
+      echo "<li class='directory collapsed'>$checkbox<a href='#' rel=\"$htmlRel/\">$htmlName</a></li>";
+    }
+    foreach ($names as $file) if (is_file($rootdir.$file)) {
+      $htmlRel  = htmlspecialchars($rootdir.$file);
+      $htmlName = htmlspecialchars($file);
+      $ext      = mb_strtolower(pathinfo($file)['extension']);
+      foreach ($filters as $filter) if (empty($filter)||$ext==$filter) {
+        if (empty($match)||preg_match("#$match#",$file)) echo "<li class='file ext_$ext'>$checkbox<a href='#' rel=\"$htmlRel\">$htmlName</a></li>";
```
OmgImAlexis, line 56: $_POST should be sanitised to prevent injecting into preg_match to cause a ReDos type attack.

ljm42:
This is a pretty major change to the "flash offline" check. It was meant to be a simple test to make sure certain files were readable and parseable, is this now doing a sha256sum on bzroot and bzimage on every page load?

plugins/dynamix/include/Report.php
```diff
@@ -12,6 +12,16 @@
...
 <?
 switch ($_POST['cmd']) {
+case 'vfat':
+  $check = "/var/tmp/check.vfat";
+  if (!file_exists($check)) {
+    $disks = parse_ini_file('/var/local/emhttp/disks.ini',true);
+    file_put_contents($check,exec("lsblk -lo NAME /dev/{$disks['flash']['device']}|awk '(NR>2)'"));
+  }
+  $dev = file_get_contents($check);
+  passthru("fsck.vfat -n /dev/$dev",$status);
```
Squidly271, line 22: From my 100% working system with NO issues and is not corrupted at all which would fail your tests

root@unRaidA:/# fsck.vfat /dev/sda
fsck.fat 4.2 (2021-01-31)
Logical sector size (1766 bytes) is not a multiple of the physical sector size.
root@unRaidA:/# echo $?
1

Squidly271: Or including the partition info,

root@unRaidA:/# fsck.vfat -n /dev/sda1
fsck.fat 4.2 (2021-01-31)
Dirty bit is set. Fs was not properly unmounted and some data may be corrupt.
 Automatically removing dirty bit.

Leaving filesystem unchanged.
/dev/sda1: 4322 files, 61969/1904448 clusters
root@unRaidA:/# echo $?
1

Now, considering that in *theory* Unraid does an fsck.vfat on bootup (which it's supposed to), this isn't exactly correct, and you can't be flagging a ton of flash drives as being "corrupt" when they're not.

bergware:
Personally I am not convinced that parsing config files is the best method, but in fact no method is perfect.

I have countered this by introducing a new setting under Flash setting, which allows the user to select the desired check method.

To keep user experience the same as before, it defaults to "configuration" check.

Squidly271:
But my point is that the first option in the drop down would flag a ton of flash drives as being corrupt even though they're not... (And I can't remember the last time I had an unclean shutdown)

The 2nd option more or less will never find anything because if the files are corrupt the odds are good that the system won't boot at all to begin with.

This default quick check (3rd option) was designed to be a stop-gap to the remounting POC since the user experience is beyond terrible when the flash drops offline, and would let the user know why all of those PHP errors are suddenly appearing in the GUI, why the name has been changed etc.  and was never particularly intended to be an actual "corruption check" but simply a notice to the user that these files which the system actually requires to operate are messed up.

ljm42:
Unraid 6.10 validates the bz* files during boot, I don't see value in revalidating them as you navigate the webgui. The original check was meant as a really low cost way of detecting if the flash went offline or if key config files were corrupted since that would bork the display.  The idea is to give users an immediate answer to the questions "why are there errors all over the webgui", "why has my theme reset", "why can't I save any changes", etc. I'd suggest we stick with the original logic without adding more options to the webgui.

bergware:
Eventually I like to have a more elaborate testing method, including the possibility for the user to repair stuff when something "wrong" is detected. A startup check is good, but during the operation things still may get corrupted (including bz* files).

For the moment I'll settle with the simple configuration check as we do today.
The check itself is moved to the background to improve GUI responsiveness.
In the test I have added more relevant configuration files (if existing)

Squidly271:
disk.cfg doesn't exist if you issue a new config.   Hence why it was excluded.   The original code contains the only files guaranteed to exist in all circumstances

Squidly271:
And shared.cfg doesn't always exist unless something has changed

bergware:
The following mechanism is in place

1. A file can exist mandatory or optional
2. If file is mandatory and does not exist : give error
3. If file exists and doesn't parse correctly: give error

This is done for all relevant cfg files to ensure they are present and correct (if present).
It makes the possibility of catching errors higher.

Squidly271:
> why can't I save any changes

It would be nice if we could find a way to determine if the flash drive is mounted RO without having to write something to it and toss up an applicable banner.

bergware:
Exist already!
The monitor script checks the status of the USB device and gives an alert when it is not r/w.

Squidly271:
Maybe look at re jigging it so that  instead of the notification it does both the notification and a banner since notifications can be disabled

bergware:
A notification alert should be enough. It is the default system thru which a user is informed and has the advantage that the user doesn't need an open browser to get the message.

limetech:
Merging, but let's discuss any further changes to 'flash corruption' before committing to code ;)

bergware:
Sure, we can. Any new ideas would be for 6.11 anyways...

PR #958, opened 2021-09-27
adding support for per user chroot jails to Unraid's stock FTP powered by vsftpd

imthenachoman:
# What and Why

I made some changes to enable per user chroot jails for vsftpd. This is the same functionality offered by ProFTPD but since vsftpd comes with Unraid and can also do it I figured it made sense.

- You can select the root folder for each user
- Users are restricted to this folder
- You can enable/disable logging to syslog

I couldn't figure out how to test the changes to the help text because I don't know how to update the helptext.dot file.

# One More File

Along with the three files listed in this commit, the `/etc/vsftpd.conf` file needs to be edited. I couldn't find it in this repo so I am putting those details here. These three settings need to be added:

    # new code start - IMTheNachoMan
    chroot_local_user=YES
    allow_writeable_chroot=YES
    user_config_dir=/boot/config/vsftpd.user_config_dir
    # end new code - IMTheNachoMan

# How It Works

- `chroot_local_user` forces users to stay within their "home directory". 
- `user_config_dir` tells `vsftpd` where to look for per user settings

I don't have a way to spin up an Unraid VM or second box so I can't do a full thorough test but testing it on my main/only/production Unraid box worked fine.

# Screenshot

image

# Known FTP Enabled Bug

I opened a bug report about this: https://forums.unraid.net/bug-reports/stable-releases/ftp-server-in-692-still-auto-starts-on-reboot-r1588/.

I think it happens because `/etc/inetd.conf` is replaced on reboot and in the stock one the ftp line is not commented out. 

To fix it, my thought was to save some FTP enable/disable status in `/boot/config/vsftpd.cfg` and then something in `/boot/config/go` that would read it but I am not convinced that is the right approach within the Unraid ecosystem.

### DIFF
--- a/plugins/dynamix/FTP.page
+++ b/plugins/dynamix/FTP.page
@@ -21,6 +21,8 @@ Tag="globe"
 $user_config_dir = "/boot/config/vsftpd.user_config_dir";
 $is_ftp_server_running = exec("lsof -i:21 -Pln|awk '/\(LISTEN\)/{print $2;exit}'") ? 1 : 0;
 $is_ftp_syslog_enabled = preg_match('/syslog_enable\s*=\s*yes/i', exec("grep syslog_enable /etc/vsftpd.conf"));
+$pasv_min_port = exec("grep pasv_min_port /etc/vsftpd.conf | awk -F '=' '{print $2}'");
+$pasv_max_port = exec("grep pasv_max_port /etc/vsftpd.conf | awk -F '=' '{print $2}'");
 ?>
 
 <script>
@@ -65,6 +67,16 @@ _(Log to Syslog)_:
   </select>
 
 :ftp_syslog_enable_help:
+
+_(PASV Min Port)_:
+: <input class='narrow' type="number" name="#arg[3]" min='1' max='65535' value='<?= $pasv_min_port ?>'>
+
+:ftp_pasv_min_port_help:
+
+_(PASV Max Port)_:
+: <input class='narrow' type="number" name="#arg[4]" min='1' max='65535' value='<?= $pasv_max_port ?>'>
+
+:ftp_pasv_max_port_help:
 
 &nbsp;
 : <input type="submit" value="_(Apply)_" disabled><input type="button" value="_(Done)_" onclick="done()">
--- a/plugins/dynamix/FTP.page
+++ b/plugins/dynamix/FTP.page
@@ -108,8 +108,6 @@ echo "</tbody></table>";
 
 <?PHP
 
-# copied from plugins\dynamix.docker.manager\include\CreateDocker.php
-
 $bgcolor = strstr('white,azure',$display['theme']) ? '#f2f2f2' : '#1c1c1c';
 
 ?>
--- a/plugins/dynamix/FTP.page
+++ b/plugins/dynamix/FTP.page
@@ -16,12 +16,38 @@ Tag="globe"
  */
 ?>
 <?
-$ftp_userlist_file = "/boot/config/vsftpd.user_list";
+// Don't need this file any more - IMTheNachoMan on 2021-09-16
+// $ftp_userlist_file = "/boot/config/vsftpd.user_list";
+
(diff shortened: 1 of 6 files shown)


OmgImAlexis:
> To fix it, my thought was to save some FTP enable/disable status in /boot/config/vsftpd.cfg and then something in /boot/config/go that would read it but I am not convinced that is the right approach within the Unraid ecosystem.

I'd avoid needing to change the `go` file.

plugins/dynamix/FTP.page
```diff
@@ -16,39 +16,142 @@ Tag="globe"
...
+
+?>
+
+&nbsp;
+: <input type="submit" value="_(Apply)_" disabled><input type="button" value="_(Done)_" onclick="done()">
+</form>
+
+<?PHP
+
+# copied from plugins\dynamix.docker.manager\include\CreateDocker.php
```
OmgImAlexis, line 111: Unneeded comment.

imthenachoman: I don't know enough about the core Unraid system and wanted to make sure folks know I just copied this from elsewhere. I was hoping it was a central library I could just reference instead of copying but since it isn't I wanted to make sure it was clear. I defer to y'all on if you do or don't want it. I'm just trying to help by adding a missing feature I think will help.

imthenachoman:
> I'd avoid needing to change the go file.

How else would I go about it? I assume `/etc/inetd.conf` is part of the stock OS file that doesn't get saved on reboot. So there would need to be someway to save and check user preference and update that file accordingly....

plugins/dynamix/FTP.page
imthenachoman: How exactly do I do this now? Do I have to change on my end and submit a new pull request?

OmgImAlexis: Pushing a new commit should be fine.

imthenachoman: Done. Thanks!

imthenachoman:
Doh. I found a bug/issue. I will fix.

imthenachoman:
I am adding support for `pasv_min_port` and `pasv_max_port` but hit a dilemma. Changes to `/etc/vsftpd.conf` do not persist on reboot. Users shouldn't have to muck with the `go` file to make their settings from the webGui stick.

Is there a way from the webGui/PHP to write changes to whatever file is being used to create the `/etc/vsftpd.conf` file on boot?

imthenachoman:
image

imthenachoman:
Hello. I was wondering if there is anything else I need to do for this pull request? I've been using it for weeks with no issue. I think it will make unRAID better. It addresses the risk and makes the experience better.

imthenachoman:
As of right now, everything works as expected except the PASV min/max port values reset after reboot. Changes to `/etc/vsftpd.conf` do not persist on reboot. Users shouldn't have to muck with the go file to make their settings from the webGui stick.

Is there a way from the webGui/PHP to write changes to whatever file is being used to create the /etc/vsftpd.conf file on boot?

imthenachoman:
I was wondering if anyone had looked at this?

If there is no intention of incorporating my code -- there are no hard feelings or issues on my part. I'd just like to know so I can "close the case" on my end. 

 And if it won't be incorporated, then I'd prefer to just nix the code I have on my Unraid box and use ProFTPD or something. I have to use FTP cause that is all my camera and printer support.

ljm42:
Apologies for letting this sit for so long. ftp is an insecure protocol that isn't seeing a lot of traction right now. I'd suggest switching to a different protocol, or an ftp server in a Docker container (which I imagine by now you have done).  Again, apologies for letting this sit.

imthenachoman:
Many things still only talk/understand FTP -- like some wifi security cameras. 

I know folks could run a Docker container but that seems like avoidable overhead when unRAID has it built in and the unRAID code just needs some tweaks to make it more secure.

PR #959, opened 2021-09-27
added Windows 11 template and OVMF TPM

ich777:


plugins/dynamix.vm.manager/templates/Custom.form.php
```diff
@@ -1254,7 +1264,8 @@ function resetForm() {
 	}
 
 	$('.advancedview').change(function () {
-		if ($(this).is(':checked')) {
+		if
```
OmgImAlexis, line 1267: Was this meant to be split onto 2 lines? 🤔

ich777: Will correct that since another change is necessary for Bitlocker to work.
Thank you @OmgImAlexis for the heads up! :)

PR #972, opened 2021-10-15
feat: make libvirt networks accessible via gui

F4IL3D:
This PR gives the user the option of using LibVirt networks in addition to bridges without having to edit the VM's XML.

I tested the changes locally on my Unraid and so far have not found any errors.

Example of a tested network:
```
<network>
  <name>sriov0</name>
  <uuid>bbb1fc77-0d0e-45c9-b577-597ac9d4792c</uuid>
  <forward mode='hostdev' managed='yes'>
    <pf dev='eth3'/>
    <address type='pci' domain='0x0000' bus='0x81' slot='0x10' function='0x1'/>
    <address type='pci' domain='0x0000' bus='0x81' slot='0x10' function='0x2'/>
    <address type='pci' domain='0x0000' bus='0x81' slot='0x10' function='0x3'/>
    ...
  </forward>
</network>
```

image
image

### DIFF
--- a/languages/en_US/helptext.txt
+++ b/languages/en_US/helptext.txt
@@ -1494,11 +1494,16 @@ Open the folder for the version of Windows you are installing and then select th
 Three drivers will be found. Select them all, click next, and the vDisks you have assigned will appear.
 :end
 
-:vms_network_bridge_help:
-Select the name of the network bridge you wish to use as default for your VMs,
-the setting 'virbr0' will let libvirt create a virtual bridge that utilizes NAT (network address translation)
-and act as a DHCP server to hand out IP addresses to virtual machines directly.
-More optional selections are present when virtual bridges are created under network settings.
+:vms_network_source_help:
+Select the name of the network you wish to use as default for your VMs.
+You can choose between **'bridges'** created under network settings or
+**'libvirt'** created with virsh command in the terminal.
+The bridge **'virbr0'** and the associated virtual network **'default'** are
+created by libvirt.  
+Both utilizes NAT (network address translation) and act as a DHCP server to hand out IP addresses to virtual machines directly.
+More optional selections are present for bridges under network settings or for libvirt networks with the virsh command in the terminal.
+
+**If your are unsure, choose 'virbr0' as the recommended Unraid default.**
 
 NOTE: You can also specify a network bridge on a per-VM basis.
 :end
--- a/languages/en_US/helptext.txt
+++ b/languages/en_US/helptext.txt
@@ -1507,8 +1507,8 @@ More optional selections are present for bridges under network settings or for l
 
 NOTE: You can also specify a network source on a per-VM basis.
 
-**IMPORTANT: Most LibVirt networks only become active as soon as the associated interface is up.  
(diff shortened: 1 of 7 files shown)


F4IL3D:
Hello Tom, 

thank you for your feedback. I understand what you mean. Especially with regard to the possible errors

To your question:
I think the best and most user-friendly solution would be a setting on the network page for the corresponding interface.
Maybe a check box to bring an interface up at startup or something like that. The prerequisite is of course that the settings on the network page are applied before libvirt is started. Otherwise the libvirt network would not be ready for a VM that is marked for autostart.
I also see another problem. At the moment it is not visible to the user in the gui that an interface is used by a libvirt network.
A hint or a graying out of the interface might be helpful.

Of course, all of this only works with networks that are persistent i think.

Other solutions are then udev rules or just a script. But I think, whether the user edited the go script or created a udev rule/script is actually the same. Apart from the possible user errors.

I don't see any other possibilities, unless unraid uses a network manager like nmcli or networkd that I don't know about. ;-)

What do you think?

bergware:
Two remarks:
1. All ethernet interfaces are automatically detected by Unraid upon system startup. The user can (need) to configure these interfaces in the GUI when these are to be used. You need to revise your help text to reflect this (it is not done to use to go file)
2. Code must have multi-language support to allow translation of text to other languages. The function _() is used for translations, e.g.  echo mk_option("", $key, "- "._($key)." -", "disabled");  and  _('default')

limetech:
> **IMPORTANT: Most LibVirt networks only become active as soon as the associated interface is up.  
This can be ensured with an 'ip link set <ethx> up' at the beginning of the go file.**

This is something that is problematic.  What we are saying is, to use a standard feature of Unraid OS a user may possibly have the requirement of making a custom mod to 'go' file which is easy for user to forget about and complicates support.  It also assumes that networking system is 'up' at time that 'go' script is executed, which is the case today but not necessarily a guarantee.

F4IL3D:
@limetech
Sorry, I forgot to push the helptext change. ^^

F4IL3D:
@limetech @bergware
Sorry for the late reply. I haven't had a lot of free time recently.

1. I have changed the help text and now refer to the NetworkSettings page. I hope the text is okay?
2. I added multi-language support. @bergware Unfortunately, I don't understand what you mean by `_('default')`?

I also added a test to the getValidNetworks () function. Without this, the VM Manager page will not be displayed as soon as Libvirt is deactivated.

I also noticed another problem. My changes seem to no longer work with version 6.10.0-rc2.
For example the libvirt-php function `libvirt_list_networks ($conn, VIR_NETWORKS_ALL);` does not return anything. At least one array with one value should be returned.
I have not found any changes to Unraid or to libvirt-php that explain this behavior.
Can you check this out, please? Maybe it's just my test system.

F4IL3D:
@limetech @bergware  
Happy new year.
The PR is ready for another review.

bergware:
PR looks good, thanks

SimonFair:
This PR has affected function of ELEC templates as missing key from array.

$arrValidNetworks[$key] is defined in Custom template, but ELECs are missing the key just have $arrValidNetworks

Which results in Array being shown and not able to set NIC source.

image

PR #980, opened 2021-11-07

bergware:


plugins/dynamix/include/OpenTerminal.php
```diff
@@ -0,0 +1,57 @@
...
+  $pid = exec("pgrep -a ttyd|awk '/\\/var\\/run\\/ttyd.sock:{print \$1}'");
+  if ($pid) break;
+  @unlink('/var/run/ttyd.sock');
+  exec("ttyd-exec -o -i '/var/run/ttyd.sock' bash --login");
+  break;
+case 'syslog':
+  $pid = exec("pgrep -a ttyd|awk '/\\/var\\/run\\/syslog.sock:{print \$1}'");
+  if ($pid) break;
+  $path = '/var/log/';
+  $file = realpath($path.$_GET['name']);
```
OmgImAlexis, line 28: Haven't run this but I have a feeling if I set `$_GET['name']` to `../../../etc/passwd` I'll get that file back. This needs a basename check to ensure the `$file` ends up in the `$path` after resolving.

bergware: realpath and the subsequent path check takes care of this.

plugins/dynamix/scripts/tty_font
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+if [[ -n $1 ]]; then
+  pkill ttyd
+  sed -ri "s/fontSize=[0-9]+/fontSize=$1/" /etc/default/ttyd
```
OmgImAlexis, line 4: Is `#arg[1]` escaped? If not when it's passed down to `$1` code injection is possible.

bergware: arguments are automatically escaped in update.php which prepares the request

limetech:
Nice work!

plugins/dynamix/include/DefaultPageLayout.php
```diff
@@ -235,6 +235,10 @@ function openWindow(cmd,title,height,width) {
   form.submit();
 }
 function openTerminal(tag,name,more,height,width) {
+  if (/MSIE|Edge/.test(navigator.userAgent)) {
```
OmgImAlexis, line 238: Isn't this going to trigger on both the old EdgeHTML based browser and the new chromium based one?

PR #990, opened 2021-12-15
Miscellaneous updates

bergware:


Squidly271:
@bergware Just noticed the ContextMenu: added option button in the change log.  Since CA now uses context menus, what exactly is this option?

PR #991, opened 2021-12-22
Miscellaneous updates

bergware:


limetech:
This looks like ability for arbitrary command execution, correct?  I don't think we want to do this.

bergware:
You are right. Reverted this PR

PR #1022, opened 2022-02-03
Always show "WebUI" for user specified URLs

shayne:
I often specify custom WebUI URLs based on my network/local-domain
setup. I rarely if ever use "http://[IP]:[PORT:8888]/" templated URLs.

This change maintains support for the existing templated
bridged/host-network URL schema while enabling the "WebUI" affordance
for containers where the user explicitly entered a WebUI url, e.g.
"https://grafana.mylan.io"

### DIFF
--- a/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -305,8 +305,14 @@ class DockerTemplates {
 			}
 			if ($ct['Running']) {
 				$port = &$ct['Ports'][0];
-				$ip = ($ct['NetworkMode']=='host'||$port['NAT'] ? $host : $port['IP']);
-				$tmp['url'] = $ip ? (strpos($tmp['url'],$ip)!==false ? $tmp['url'] : $this->getControlURL($ct, $ip)) : $tmp['url'];
+				$webui = $this->getTemplateValue($ct['Image'], 'WebUI');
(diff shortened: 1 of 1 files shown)


Squidly271:
Seems to work ok, but maybe I'm missing something - It also worked before.  I can set any of the webUI's to any random website and it works just fine.  What is the setup of a container (networking etc) where this didn't work prior?

shayne:
My fault! Forgot to explain the issue.

If a container doesn't have a port mapping, the existing logic keep it from displaying WebUI in the drop-down. If you're not using the Unraid "map ports to host IP" then the WebUI feature is kinda broken/useless.

This change promotes using WebUI for both cases, either in the tradition al way or if you're using something more advanced/custom. Simply setting a custom URL for a container will allow WebUI to appear.

Example:

I use Traefik (container on Unraid) and my containers can register themselves with Traefik via Docker labels. I access the containers via URLs like "https://grafana.mylan.io". The containers don't have port mappings back to the host as it's not required (either bridge or br0 networking). With this change I can edit the templates for my container and add these URLs and use the WebUI feature like anyone else, just with my custom URLs instead of the templated IP/PORT standard URLs.

Squidly271:
Doesn't look like it'll hurt anything (and may solve some edge cases that aren't easy to test), but I tested on RC3K and with a container set with no ports and network=none I could still hit any random website on via WebGUI.

Either way, since it worked fine with bridge & host and ports being mapped I've have no major issues with it.

PR #1028, opened 2022-02-23
Enable/Disable SMART extended test depending on spin down delay setting

bergware:


limetech:
Nice solution, thanks.

PR #1030, opened 2022-02-24
Change Dashboard Parity status to be invalid and not emulated.

SimonFair:
Please find suggested changes to Dash board view if Parity is invalid or being rebuilt, Currently it says emulated.

Let me know if I should remove multi-lang or add to others and status if you think should be something different to rebuild or invalid.

Made the same change on my 6.9.2 server as its currently rebuilding parity.

image

### DIFF
--- a/plugins/dynamix/nchan/update_2
+++ b/plugins/dynamix/nchan/update_2
@@ -148,7 +148,7 @@ function device_status(&$disk, &$error, &$warning) {
     case 'green-blink' : $color = 'grey';   $text = 'standby';    break;
     case 'blue-on'     : $color = 'blue';   $text = 'unassigned'; break;
     case 'blue-blink'  : $color = 'grey';   $text = 'unassigned'; break;
-    case 'yellow-on'   : $color = 'yellow'; $text =  $disk['type']=='Parity' ? 'invalid') : 'emulated';   $warning++; break;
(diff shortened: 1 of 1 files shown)


Squidly271:
What adding text to the GUI, you should also add the line to https://github.com/unraid/lang-en_US in this case translations.txt and reference this PR

bergware:
To remain consistency with the Main page, the status of the parity disk(s) should be reported as "invalid".

In the latest rc3 the exact parity operation is displayed, so the user knows what is going on.

SimonFair:
> To remain consistency with the Main page, the status of the parity disk(s) should be reported as "invalid".
> 
> In the latest rc3 the exact parity operation is displayed, so the user knows what is going on.
> 
> Ps. regarding translations. I run a code scanner to find new translations and add them to the appropriate text files, no need to worry about that.

Change status to invalid and added Multi language support for other values.

bergware:
You don’t need to translate each individual text item.The translation takes place in the text which is returned at the end.

SimonFair:
> You don’t need to translate each individual text item.The translation takes place in the text which is returned at the end.

Thanks sorry missed that, I have updated to remove _( )

bergware:
Looks good

limetech:
Thank you @SimonFair

PR #1033, opened 2022-02-27
Fixed: parity history sometimes wrongly processed

bergware:


bergware:
Some weirdness going on ... need to check

PR #1056, opened 2022-03-19
Highlight selected row when hovering over array or shares

bergware:


Squidly271:
LOL.  Yeah, we never thought about anyone opening a tab in a new window

PR #1057, opened 2022-03-21
Make links in help text standout (red)

bergware:


OmgImAlexis:
Is this correct? Red tend to mean bad things to users. Should they not be green?

zackspear:
Agree with @OmgImAlexis. We're both colorblind and there's a more accessible way to accomplish readability here.

Changing the `blockquote` text color from `#3a87ad` to `#222222` would do it. This would improve contrast for all the help text and help text links.

Then I'd do the following for emphasizing the links further
```css
blockquote a {
  color: #ff8c2f;
  text-decoration: underline;
  font-weight: 600;
}

blockquote a:hover,
blockquote a:focus {
  color: #f15a2c;
}
```

**BEFORE**
Screen Shot 2022-03-21 at 13 47 17
**AFTER**
Screen Shot 2022-03-21 at 13 48 43

bergware:
Sure no problem to improve visibility for the color-blind (I am not).
@zackspear  I took your suggestions, minus the underlining. To stay consistent with how links are displayed everywhere else in the GUI.

PR #1058, opened 2022-03-21
Update Outlook/Hotmail smtp settings

eschultz:
ref: https://support.microsoft.com/en-us/office/pop-imap-and-smtp-settings-8361e398-8af4-4e97-b147-6c6c4ac95353


OmgImAlexis (approved):
LGTM

PR #1085, opened 2022-04-27
Wireguard: fix import function to accept all keys

bergware:


limetech:
sneaking in a scaling issue fix too :)

PR #1090, opened 2022-04-29
fix: better IPv6 support in UpdateDNS

ljm42:


bergware:
The IPv6 won't work because it will return a host address, like "fd42::abcd:1234:abcd:1234"
Here is my proposed solution:

  $validResponse = ["192.168.42.42", "fd42"];
  $response = host_lookup_ip($rebindtesturl);
  return in_array($response, explode('::',$validResponse)[0]);

PR #1100, opened 2022-05-07
Fix side bar of themes azure/gray in firefox

bergware:
Until version 100 of Firefox a workaround was used, but not needed anymore. Firefox acts the same as other browsers


Squidly271:
FWIW, maybe we should also upgrade FF in the GUI boot mode so we don't now have display aberrations with it.,

PR #1114, opened 2022-06-17
Replace domain for container webui with document.domain

FunnyPocketBook:
Right now the webui links all point to the IP of unRaid in the local network (e.g. `192.168.0.5`). If one uses a VPN or has a domain that points to the webui, the container links to the webui will still point to `192.168.0.5`.
This PR replaces the hostname of the URL to the webui with `document.domain`, ensuring that the URL to the container webUIs is always using the same hostname as the one the user is currently using.


OmgImAlexis (approved):
LGTM

bergware:
Sorry, but this breaks access to any docker container running on a custom IP address.

ljm42:
Thank you for the submission, but we aren't able to merge it as it will break containers that run on their own IP address.

PR #1124, opened 2022-07-03
System update

bergware:


limetech:
lol the icons

bergware:
Some history indeed...
Just made the background transparent so they are properly displayed in both white or black theme.

bergware:
I have replaced all the 'openBox' calls for the new method, which uses either 'openPlugin' or 'openChanges'.
This makes the GUI and all popup windows consistent.

Only outstanding action is Docker, which still uses its own popup window..

PR #1127, opened 2022-07-07
Update ssh key regex

nemchik:
Add support for ed25519/sk-ed25519
Remove support for ecdsa (insecure)
Use proper regex to check for valid key types

Ref:
https://forums.unraid.net/bug-reports/stable-releases/6103-ui-field-for-root-ssh-authorized-keys-does-not-support-ed25519-and-other-valid-key-types-r2023/
https://github.com/nemchik/ssh-key-regex


bergware:
Useful addition, thanks

limetech:
Yes, thank you!

PR #1146, opened 2022-08-05
Improve Dockerman icon caching.

dcflachs:
In 6.10.1+ the dockerman icon caching mechanism does not work correctly for icons specified by the net.unraid.docker.icon container label on containers without a dockerman template. For containers created with this label outside of dockerman the caching mechanism initially downloads and displays the correct icon. If the icon url is changed however the new icon is never downloaded because the old icon remains cached and there is no mechanism in place for invalidating the icon in the absence of a dockerman template. To fix this i propose the following changes to dockerman.

Revision update on June 12, 2026: rebased onto current master at cc6a800b1 / 7.3.2-rc.0.1 and revised the cache refresh logic for the 7.3 Docker tab fast path. Dockerman-managed containers keep template-first icon resolution. Non-Dockerman containers with net.unraid.docker.icon now track the last label URL, refresh when that URL changes, clear cached label icons when the label is removed, and avoid repeated download attempts for unchanged bad URLs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevents accumulation of outdated container icons; ensures removed or changed icon labels update and stale icons are purged.
  - Forces re-download of missing or updated icons so displayed images stay current, with a fallback to the default image when needed.

- **Refactor**
  - Streamlined icon caching and storage handling for more reliable icon management during container creation and updates.

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -151,10 +151,7 @@ if (isset($_POST['contName'])) {
     if (is_file($filename)) {
       $oldXML = simplexml_load_file($filename);
       if ($oldXML->Icon != $_POST['contIcon']) {
-        if (!strpos($Repository,":")) $Repository .= ":latest";
-        $iconPath = $DockerTemplates->getIcon($Repository,$Name);
-        @unlink("$docroot/$iconPath");
-        @unlink("{$dockerManPaths['images']}/".basename($iconPath));
+        $DockerTemplates->purgeUnusedIconFiles($Name);
       }
     }
     file_put_contents($filename, $postXML);
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -337,13 +337,25 @@ class DockerTemplates {
 			// looking for a match. Bad.
 			$labelIconUrl = $ct['Icon'] ?? null;
 			if (!$communityApplications) {
+				$isDockerman = ($ct['Manager'] ?? '') === 'dockerman';
+				$preferLabelIcon = !empty($labelIconUrl) && !$isDockerman;
+				$labelIconChanged = $preferLabelIcon
+					&& ($tmp['icon-url'] ?? '') !== $labelIconUrl;
+				$labelIconRemoved = !$isDockerman && !empty($tmp['icon-url']) && empty($labelIconUrl);
 				$iconExists = !empty($tmp['icon'])
 					&& (is_file($tmp['icon']) || is_file($docroot . $tmp['icon']));
-				if (!$iconExists || $reload) {
-					$tmp['icon'] = $this->getIcon($image, $name, $labelIconUrl ?: ($tmp['icon'] ?? ''));
+				if ($labelIconRemoved) {
+					$this->purgeUnusedIconFiles($name);
+					$tmp['icon'] = '/plugins/dynamix.docker.manager/images/question.png';
+					unset($tmp['icon-url']);
+				} elseif (!$iconExists || $reload || $labelIconChanged) {
+					$tmp['icon'] = $this->getIcon($image, $name, $labelIconUrl ?: ($tmp['icon'] ?? ''), $preferLabelIcon);
+					if ($preferLabelIcon) $tmp['icon-url'] = $labelIconUrl;
 					// Explicitly return the fallback asset, so that subsequent polls see
 					// the local file instead of rerunning the expensive template scan
 					if (empty($tmp['icon'])) $tmp['icon'] = '/plugins/dynamix.docker.manager/images/question.png';
+				} elseif (!$preferLabelIcon) {
+					unset($tmp['icon-url']);
 				}
 			}
 			if ($ct['Running']) {
@@ -430,14 +442,28 @@ class DockerTemplates {
 		return $info;
 	}
 
-	public function getIcon($Repository,$contName,$tmpIconUrl='') {
+	private function getIconFile($imgUrl) {
+		return sprintf('%s-%s.png', 'icon', sha1($imgUrl));
+	}
+
+	private function getIconRamPath($contName, $imgUrl) {
+		global $dockerManPaths;
+		return sprintf('%s/%s-%s', $dockerManPaths['images-ram'], $contName, $this->getIconFile($imgUrl));
+	}
+
+	public function getIcon($Repository,$contName,$iconUrl='',$preferIconUrl=false) {
 		global $docroot, $dockerManPaths;
-		$imgUrl = $this->getTemplateValue($Repository, 'Icon','all',$contName);
-		if (!$imgUrl) $imgUrl = $tmpIconUrl;
+		if ($preferIconUrl) {
+			$imgUrl = $iconUrl ?: $this->getTemplateValue($Repository, 'Icon','all',$contName);
+		} else {
+			$imgUrl = $this->getTemplateValue($Repository, 'Icon','all',$contName);
+			if (!$imgUrl) $imgUrl = $iconUrl;
+		}
 		if (!$imgUrl || trim($imgUrl) == "/plugins/dynamix.docker.manager/images/question.png") return '';
 
-		$iconRAM = sprintf('%s/%s-%s.png', $dockerManPaths['images-ram'], $contName, 'icon');
-		$icon    = sprintf('%s/%s-%s.png', $dockerManPaths['images'], $contName, 'icon');
+		$iconFile = $this->getIconFile($imgUrl);
+		$iconRAM = $this->getIconRamPath($contName, $imgUrl);
+		$icon    = sprintf('%s/%s', $dockerManPaths['images'], $iconFile);
 
 		if (!is_dir(dirname($iconRAM))) mkdir(dirname($iconRAM), 0755, true);
 		if (!is_dir(dirname($icon))) mkdir(dirname($icon), 0755, true);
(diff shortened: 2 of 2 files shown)


plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -336,15 +336,16 @@ public function getAllInfo($reload=false,$com=true,$communityApplications=false)
 		return $info;
 	}
 
-	public function getIcon($Repository,$contName,$tmpIconUrl='') {
+	public function getIcon($Repository,$contName,$iconUrl='') {
 		global $docroot, $dockerManPaths;
-		$imgUrl = $this->getTemplateValue($Repository, 'Icon','all',$contName);
-		if (!$imgUrl) $imgUrl = $tmpIconUrl;
+		$imgUrl = $iconUrl ?: $this->getTemplateValue($Repository, 'Icon','all',$contName);
```
dcflachs, line 341: Prefer iconUrl specified by container label if it exists.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -336,15 +336,16 @@ public function getAllInfo($reload=false,$com=true,$communityApplications=false)
...
 		if (!$imgUrl || trim($imgUrl) == "/plugins/dynamix.docker.manager/images/question.png") return '';
 
-		$imageName = $contName ?: $name;
-		$iconRAM = sprintf('%s/%s-%s.png', $dockerManPaths['images-ram'], $contName, 'icon');
-		$icon    = sprintf('%s/%s-%s.png', $dockerManPaths['images'], $contName, 'icon');
+		$imgUrlHash = sha1($imgUrl);
+		$iconFile = sprintf('%s-%s.png', 'icon', $imgUrlHash);
```
dcflachs, line 345: Instead of storing the icons by container name only this change uses a sha1 hash of the imgUrl. This way a change to the icon url will automatically invalidate the cached icon.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -336,15 +336,16 @@ public function getAllInfo($reload=false,$com=true,$communityApplications=false)
...
+
+		$iconRAM = sprintf('%s/%s-%s', $dockerManPaths['images-ram'], $contName, $iconFile);
+		$icon    = sprintf('%s/%s', $dockerManPaths['images'], $iconFile);
```
dcflachs, line 348: In images-ram the icon file name contains both the container name and the urlHash so that it is unique for each container.
In images the icons are store by urlHash alone. This has the benefit of storing only a single copy of identical icons.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -354,14 +355,47 @@ public function getIcon($Repository,$contName,$tmpIconUrl='') {
...
+			exec("/usr/bin/logger ".escapeshellarg("$contName: Could not download icon $imgUrl"));
+		}
+		else {
+			$this->purgeUnusedIconFiles($contName, $iconFile);
 		}
 
 		return (is_file($iconRAM)) ? str_replace($docroot, '', $iconRAM) : '';
 	}
+
+	public function purgeUnusedIconFiles($contName, $keepIcon='') {
```
dcflachs, line 370: This new function better handles the more complex removal process for this caching mechanism.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -354,14 +355,47 @@ public function getIcon($Repository,$contName,$tmpIconUrl='') {
...
+		global $docroot, $dockerManPaths;
+
+		$icon_glob = sprintf('%s/%s-*.png', $dockerManPaths['images-ram'], $contName);
+		$ramFiles = glob($icon_glob);
+		foreach ($ramFiles as $filename) {
+			if ( ($keepIcon === '') || !(strpos($filename, $keepIcon) !== false) ) {
+				@unlink($filename);
+			}
+		}
```
dcflachs, line 379: The first block removes all icons associated with contName in images-ram except the one specified nu keepIcon.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -354,14 +355,47 @@ public function getIcon($Repository,$contName,$tmpIconUrl='') {
...
+
+		$icon_glob = sprintf('%s/%s*.png', $dockerManPaths['images'], $contName);
+		foreach (glob($icon_glob) as $filename) {
+			if ( ($keepIcon === '') || !(strpos($filename, $keepIcon) !== false) ) {
+				@unlink($filename);
+			}
+		}	
+
```
dcflachs, line 387: The next block removes all icons stored in the old style from the images directory. Ideally this would only need to be run once when the caching mechanism was first upgraded to purge all old icons. Unfortunately i could not think of a good way to make that happen so i settled for always running this check.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -354,14 +355,47 @@ public function getIcon($Repository,$contName,$tmpIconUrl='') {
...
+		$icon_glob = sprintf('%s/%s*.png', $dockerManPaths['images'], $contName);
+		foreach ($ramFiles as $ramFile) {
+			if ( strpos($ramFile, '-icon-') !== false ) {
+				$suffix = end(explode('-', $ramFile));
+				if ( !glob($dockerManPaths['images-ram'].'/*icon-'.$suffix) ) {
+					$filename = sprintf('%s/icon-%s', $dockerManPaths['images'], $suffix);
+					@unlink($filename);
+				}
+			}
+		}
```
dcflachs, line 397: The final block removes any of the shared icons in images that share the same hash as icons removed in the first block that are not referenced by another icon in images-ram.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -336,15 +336,16 @@ public function getAllInfo($reload=false,$com=true,$communityApplications=false)
...
 
-	public function getIcon($Repository,$contName,$tmpIconUrl='') {
+	public function getIcon($Repository,$contName,$iconUrl='') {
 		global $docroot, $dockerManPaths;
-		$imgUrl = $this->getTemplateValue($Repository, 'Icon','all',$contName);
-		if (!$imgUrl) $imgUrl = $tmpIconUrl;
+		$imgUrl = $iconUrl ?: $this->getTemplateValue($Repository, 'Icon','all',$contName);
 		if (!$imgUrl || trim($imgUrl) == "/plugins/dynamix.docker.manager/images/question.png") return '';
 
-		$imageName = $contName ?: $name;
```
dcflachs, line 345: I removed this line specifically because i could not figure out what it was doing. So far as i can tell $name is never set anywhere.

plugins/dynamix.docker.manager/include/CreateDocker.php
```diff
@@ -88,9 +88,7 @@ function cpu_pinning() {
       $oldXML = simplexml_load_file($filename);
       if ($oldXML->Icon != $_POST['contIcon']) {
         if (!strpos($Repository,":")) $Repository .= ":latest";
-        $iconPath = $DockerTemplates->getIcon($Repository,$Name);
-        @unlink("$docroot/$iconPath");
-        @unlink("{$dockerManPaths['images']}/".basename($iconPath));
+        $DockerTemplates->purgeUnusedIconFiles($Name);
```
dcflachs, line 91: Since removing an icon from the cache is a bit more complex I use a new function specifically for that purpose.

dcflachs:
My unRAID server has ~42 containers running with about ~30 unique icons. In my testing the changes to the caching mechanism do not seem to greatly effect the execution time of the getAllInfo function.

plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -303,7 +303,7 @@ public function getAllInfo($reload=false,$com=true,$communityApplications=false)
 			if ($ct['Url'] && !$tmp['url']) $tmp['url'] = $ct['Url'];
 			if ($ct['Icon']) $tmp['icon'] = $ct['Icon'];
 			if ( ! $communityApplications ) {
-				if (!is_file($tmp['icon']) || $reload) $tmp['icon'] = $this->getIcon($image,$name,$tmp['icon']);
+				if (!is_file($tmp['icon']) || $reload) $tmp['icon'] = $this->getIcon($image,$name,$ct['Icon']);
```
dcflachs, line 306: If this change is not made then subsequent calls to getAllInfo will invalidate the cache by passing in the icon file path via tmp['icon'] instead of the icon url.

dcflachs:
Any chance you would have time to look this pull request over and provide feedback?
@limetech @bergware

teambvd:
It'd be great to get some eyes on this whenever someone has time - I've got between 50 and 60 containers on each of my unraid servers, and loading the docker tab's gotten to feeling pretty clunky at this point, so any performance improvement would be great!

dcflachs:
Rebased changes onto latest rev

dcflachs:
I would love to get some feedback on this issue and proposed fix.

Squidly271:
Not my call, but since this is from 2022 you might want to retest the base system on 7.x and resubmit (even though there are no merge conflicts).  Quite probable that since it is a very old PR that it's not going to get looked at since there have been numerous changes in DockerClient.

richard-lee-863:
Bump. Really need this.

joyboy-za:
I agree. 

I didn't know what was happening when I added a second container of the same image and they had 2 different icons but came from the same registry.

Even if there was a way to only do it via the command line, that would work for me

thezoggy:
the fallout of this introduce a bug where the default icon ends up getting removed if container is updated
https://product.unraid.net/p/7-3-2-dashboard-nginx-log-spam-from-stale-docker-template-fallback-icons

which then results in log spam and fills up peoples logs due to base os icon missing.

PR #1152, opened 2022-08-23
Dockerman add ability to specify shell with container label.

dcflachs:
Add a new dockerman system label net.unraid.docker.shell to allow specifying what shell to use for the console link via labels.
Nominal values would be 'sh' or 'bash'.
This will allow containers not created by dockerman to still specify a shell for the webui


limetech:
Thank you!

PR #1155, opened 2022-08-25
Fix SAS Selftest

SimonFair:
Errors

static const char * self_test_result[] = {
        "Completed                ",
        "Aborted (by user command)",
        "Aborted (device reset ?) ",
        "Unknown error, incomplete",
        "Completed, segment failed",
        "Failed in first segment  ",
        "Failed in second segment ",
        "Failed in segment",            /* special handling for result 7 */
        "Reserved(8)              ",
        "Reserved(9)              ",
        "Reserved(10)             ",
        "Reserved(11)             ",
        "Reserved(12)             ",
        "Reserved(13)             ",
        "Reserved(14)             ",
        "Self test in progress ..."
}

Should 'No self-tests logged on this disk' be replaced with Drive in standby unable to read self tests

if the device is in standby?

### DIFF
--- a/plugins/dynamix/include/SmartInfo.php
+++ b/plugins/dynamix/include/SmartInfo.php
@@ -217,22 +217,39 @@ case "stop":
   exec("smartctl -X $type ".escapeshellarg("/dev/$port"));
   break;
 case "update":
-  $progress = exec("smartctl -n standby -c $type ".escapeshellarg("/dev/$port")."|grep -Pom1 '\d+%'");
-  if ($progress) {
-    echo "<span class='big'><i class='fa fa-spinner fa-pulse'></i> "._('self-test in progress').", ".(100-substr($progress,0,-1))."% "._('complete')."</span>";
-    break;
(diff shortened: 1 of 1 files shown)


SimonFair:
@limetech where do you collect $disk["transport"] from? as I am using as condition for test.

limetech:
> @limetech where do you collect $disk["transport"] from? as I am using as condition for test.

It's the prefix of the /dev/disk/by-id string which corresponds to the device.  We check for these prefixes:
"ata-"
"scsi-SATA_"
"scsi-1ATA_"
"scsi-"
"usb-"
"nvme-"
"virtio-"

When found we strip the trailing "-" or "_" character.

limetech:
> Should 'No self-tests logged on this disk' be replaced with Drive in standby unable to read self tests if the device is in standby?

Probably.

SimonFair:
Change no self tests to following if spundown.

image

If the drive is scsi use standby option for Identity as some SAS drives will spinup.

image

PR #1160, opened 2022-09-18
Add Missing USB support

SimonFair:
Add Missing USB devices on list so can be removed from VM. 
Add startupPolicy fields to USB list for future updates.

These are the options available to startupPolicy 
image

For the GUI input it can be a select drop down for each of the values.

image

or as per the feature request just add a check box for the optional option which I think will be the better option as the requisite option is unlikely to be used.

image
image

Link to feature request:

https://forums.unraid.net/topic/127949-support-for-adding-optional-usb-devices-in-vm-manager-form-view/#comment-1165862

Used checkbox option in PR.


limetech:
Good change, thanks!

PR #1168, opened 2022-10-11
CD Hotplug bug fix and boot order for nvme pci passthru

SimonFair:
Fixes for CD Hotplug.

Add Support for NVME boot order.

image


eschultz (approved):
LGTM

PR #1177, opened 2022-10-25
Fix post hook failing when plugin download fails; default scheduled parity check to non correcting.

dlandon:
Fix post hook failure when plugin download fails.  Php reserved characters were being passed in the $error string to the post_plugin_check causing it to fail.

### DIFF
--- a/plugins/dynamix/include/update.parity.php
+++ b/plugins/dynamix/include/update.parity.php
@@ -30,7 +30,7 @@ if (isset($_POST['#apply'])) {
     $month = $_POST['month'] ?? '*';
     $day   = $_POST['day'] ?? '*';
     $write = $_POST['write'] ?? '';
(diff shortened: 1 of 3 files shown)


plugins/dynamix/include/update.parity.php
```diff
@@ -30,6 +30,7 @@ function month($m) {
     $month = $_POST['month'] ?? '*';
     $day   = $_POST['day'] ?? '*';
     $write = $_POST['write'] ?? '';
+	$write = ($write == "CORRECT") ? "" : "NOCORRECT";
```
ljm42, line 33: tabs vs spaces, causes strange indenting

dlandon: Fixed.  I normally tab in all my work.  I'm not used to spaces.

PR #1185, opened 2022-10-30
Add boot order for USB and USB Boot option

SimonFair:
Add boot order field to USB devices
Add OS boot setting for USB Boot.


SimonFair:
USB Boot option.
image

Boot Order
image

PR #1189, opened 2022-11-02
Redo my failed attempt at defaulting to non-correcting parity check.

dlandon:
My original attempt to fix the default parity check was pretty much a disaster.  This cleans that up.  All I had to do was set the write="NOCORRECT" in the default.cfg file and set the non-correcting as the first selection.  Duh!!


limetech:
thank you!

PR #1199, opened 2022-11-12
Miscellaneous fixes

bergware:


dlandon:
See Asana notes.  I think you'll need to turn the input array into a string to be applied to the share.

Tooa:
> "[Revert "NFS security rule: change input to textarea, which allows more input](https://github.com/limetech/webgui/pull/1199/commits/54d094111b477e5b33f3e9cebc11cdd847d745f1)"

I noticed several attempts making the NFS rule field larger. Any plans to still tackle the problem @bergware? I am affected from the restriction as you can read [here](https://forums.unraid.net/topic/124722-permission-and-ownership-issues-with-nfsv4-and-rsync-unraid-6102/?do=findComment&comment=1200933).

dlandon:
It turned out to be a bit of an issue to convert the text area into a single line for the rules.  What length of the rules line would allow you to add enough rules text?

PR #1202, opened 2022-11-16
Change to use API

SimonFair:
Change to use libvirt_domain_interface_addresses which has been available for a long time but was not documented in the API manual. Using this function stops tainted warnings being written to the log.

Domain id=4 is tainted: custom-ga-command will no longer be seen in the log.


limetech:
Thank you for this!

PR #1204, opened 2022-11-19
revert change

ich777:
This change will not only affect new installations, this will also affect existing installations but only if one changes something in the Docker settings and if the line DOCKER_NETWORK_TYPE isn't found in the docker.cfg. It would be better to change the docker.cfg file which is shipped with Unraid in the zip file (append the line DOCKER_NETWORK_TYPE="1") but this is something that @limetech has to change.


dlandon:
The change should be right.  Let me have a look at the code for setting the network.  I think that may be where the problem lies,

dlandon:
Simon, please close this PR.  It needs to stand.  I'll issue a PR to resolve this problem and where macvlan cannot be selected.

dlandon:
Go ahead and apply this PR.  Because of the way the code is written, I cannot get it to be backwards compatible.

PR #1205, opened 2022-11-19
Fix default problem in Docker Settings where it defaults incorrectly and macvlan cannot be selected.

dlandon:
Tom, there is a change that also needs to be made in rc.docker:
Change this:
if [[ -z $DOCKER_NETWORK_TYPE ]]; then

at line 71 to this:
if [ "$DOCKER_NETWORK_TYPE" = "0" ]; then

I don't think I have access to the rc.docker code.

Unfortunately, we should really issue an rc so we can check it internally before release.  It is causing support problems as is stands.


dlandon:
Hold on this PR.  I have to work on backwards compatibility.

PR #1207, opened 2022-11-25
Add Serial option to vdisk

SimonFair:


SimonFair:
This also fixes boot order if not all devices have boot order set.

SimonFair:
Fix help text for Unraid Shares to show if Windows and add additional link to for setup.

image

PR #1219, opened 2022-12-18
Update Memory Backup processing for Virtiofs.

SimonFair:


SimonFair:
I have used the same class as @Squidly271 uses in CA Helpers * Array2XML: A class to convert array in PHP to XML. Should I use a different name to avoid conflicts?

PR #1224, opened 2022-12-27
feat(docker): option to disable readmore-js on container table

iarp:
This PR adds a user configurable option to disable readmore-js usage on the docker container table.

The only part I wasn't able to confirm working is the help text. I'm unsure how to rebuild translations or how that system works exactly.


bergware:
Your PR is alright.
Translation files will be updated, you don't need to worry about that

limetech:
thank you!

PR #1225, opened 2022-12-29
Miscellaneous fixes

bergware:
This error prevents the selection of another case

### DIFF
--- a/plugins/dynamix.vm.manager/VMMachines.page
+++ b/plugins/dynamix.vm.manager/VMMachines.page
@@ -101,7 +101,7 @@ if ($subaction) {
   }
   echo "<script>clearHistory();</script>";
 }
-if ($libvirt_running=='yes') $vms = $lv->get_domains();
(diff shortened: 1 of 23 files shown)


plugins/dynamix.vm.manager/VMMachines.page
```diff
@@ -101,7 +101,7 @@ if ($subaction) {
   }
   echo "<script>clearHistory();</script>";
 }
-if ($libvirt_running=='yes') $vms = $lv->get_domains();
+if ($libvirt_running=='yes') $vms = $lv->get_domains() ?: [];
```
eschultz, line 104: I think the same fix will need to be applied to this line too?: https://github.com/limetech/webgui/blob/a5329592b598fafcb8f93102ba21d8a871413494/plugins/dynamix.vm.manager/include/VMMachines.php#L25

limetech: yeah I'll make that change, thx!

PR #1226, opened 2023-01-02
Dashboard: make items moveable between columns

bergware:
This PR also introduces a system to completely hide items and a content manager function to unhide them


Squidly271:
While not a show stopper IMO, should be noted that this new system will result in a couple of plugins being incompatible (GPU top, Plex Streams etc)

bergware:
Yes, this is true: any plugin doing stuff on the dashboard page will fail, because under the hood the construction has completely changed and is not compatible with the "old" way.

bergware:
@zackspear  I have included the changes of the older PR (actually changed more elements which are recurring to a class definition)

PR #1230, opened 2023-01-09
Limit ZFS option to pool devices only

bergware:
ZFS not available (yet) for the array


limetech:
No we don't want to do this...

bergware:
Currently I can change one of the existing devices in the array to use zfs, is that allowed?

PR #1232, opened 2023-01-12
Fix for missing NIC option

SimonFair:
If mac address is blanked on template it will remove NIC. But you cannot add back in from the template only via XML editior. This PR allows new MAC to be created to add back the NIC.


limetech:
Nice addition, thx!

PR #1242, opened 2023-01-30
overhaul from third party plugins diagnostics and additions

ich777:
- created BEGIN and END comment for third party plugins
- added LXC thridparty plugin to diagnostics
- added iSCSI-Target thridparty plugin to diagnostics
- added iSCSI-Initiator thridparty plugin to diagnostics


limetech:
very nice addition, thanks!

PR #1244, opened 2023-02-02
Added Compression & Autotrim to pools

bergware:


limetech:
@bergware  You're a rockstar, thanks!

PR #1279, opened 2023-03-07
fix regex in diagnostics

ljm42:


Squidly271:
Thanks.  I don't use /remotes anymore so didn't catch the typo

PR #1281, opened 2023-03-15
DockerSettings.page Check if docker.sock exists to determine if docker is running

GamerSocke:
Searching for a running process named "dockerd" finds other running Docker daemons on the system, for example, in a lxc container. To check if the unraid Docker daemon is running, check if the /var/run/docker.sock file was created.


limetech:
What about something like this:

$DockerStopped = exec("/etc/rc.d/rc.docker status") === false;

PR #1291, opened 2023-03-25
Fix table layout for orphan images

Commifreak:
The orphan image current span results in `Created xx ago` is not aligned with the right column:

Old:
grafik

New:
grafik


bergware:
Good observation. Thanks

PR #1293, opened 2023-03-26
Plugin: Don't show update button if incompatible

Squidly271:
I could've sworn something like this went into 6.7 but obviously wrong


bergware:
Haven't tested this, but not sure if this works for Unraid OS updates.
The original code uses version_compare for OS updates, and string_compare for plugin updates.
Did you test OS upgrades?

Squidly271:
This has nothing to do with plugin versions hence no strcmp.   It's simply comparing the current OS version against the min and max if an author specifies one and doesn't offer the update if the plugin won't install anyways. 

Solves a recurring problem that happened with various plugins being made to be 6.12 only

So far as OS is concerned  the test specifically excludes the OS via !$os

bergware:
I see, thanks

PR #1307, opened 2023-04-21
VM Manager Updates

SimonFair:
Remove lock posix and flock from virtiofsd options as QEMU V8 has dropped virtiofsd and the rust version need to be used going forwards but does not support these options.

https://gitlab.com/virtio-fs/virtiofsd/-/releases

Add support to start remote-viewer from the context menus. This will download a .vv file and if set in the browser will load remote-view will the .vv file.

image
image
image
image
image

Please review and let me know any feedback.


SimonFair:
VM Settings Page

image

limetech:
Awesome!  thank you

PR #1313, opened 2023-04-25
Update helptext.txt

bergware:


ljm42:
The url needs to change too, from https://forums.unraid.net/my-servers/ to https://connect.myunraid.net/

bergware:
URL changed, thanks

PR #1331, opened 2023-05-07
VM Snapshots

SimonFair:
Following added to go file.

rm -r  /var/lib/libvirt/qemu/snapshot/
ln -fs /etc/libvirt/qemu/snapshot/ /var/lib/libvirt/qemu/

This needs to be created in the libvirt disk image. 

mkdir /etc/libvirt/qemu/snapshot/

image

QEMU Functions added Snapshot(External) and Revert(External) for VMs not running. Revert will delete snapshots and metadata.

Options to be looked at adding 
Integrate with file system snapshots(Manual i.e. not QEMU)


SimonFair:
image
image

PR #1342, opened 2023-05-29
Fixed thresholds

bergware:


bergware:
Updated to support PHP8

PR #1350, opened 2023-06-04
rc.docker: Fix multiple fixed IPs

Leseratte10:
I noticed a bug in the rc.docker script in 6.11.5. 

I'm running my Docker containers in a custom network `br1` with static IPv4 and IPv6 addresses. I have the addresses added in the "Fixed IP address" field. 

That worked just fine when creating a new container and starting it through that view, or when editing a container. However, after disabling and re-enabling Docker, or after rebooting the server, I noticed that my containers didn't autostart and starting them through the GUI also failed. I had to edit them and re-apply the config to make them start. 

Reading through the rc.docker script, I found the bug that causes this: I had the IP addresses (one IPv6, one IPv4) seperated by spaces, not by commas. When creating a container this works just fine, but when the system starts the container after a reboot, the spaces get trimmed, which resulted in my containers having IPs like `fd12:3456:7890::25:510.0.25.5` (IPv6 and IPv4 immediately after another without any seperator), which Docker obviously rejects. Editing the container sets the IPs correctly again and the container starts. 

I noticed that the old code just removes spaces, and replaces commas with a semicolon. Probably in an assumption that a user would add a comma-separated list that may also have spaces for better readability. 

However nowhere in the GUI does it state that the IPs should be comma-seperated, and given that UnRAID does accept space-seperated IPs at container creation (see #468), it should also still accept them after a reboot. So I changed the code that parses these to make it accept spaces as well. 

I also noticed that there's a discrepancy in how IPv6 addresses are detected. When a container is created, it assumes that any IP that has a colon in it is IPv6 (others are IPv4). When a container is auto-booted, it assumes any IP that has a dot in it is IPv4 (others are IPv6). 

Both assumtions are wrong - `2001:db8::1.2.3.4` has dots but is an IPv6, and `::ffff:1.2.3.4` has colons but is IPv4 (and has to be added with `--ip`, not `-ip6`). 

I didn't find a nice bash-only way to make it detect ::ffff:1.2.3.4 (and also "variants" like 0:0::ffff:1.2.3.4) as IPv4 but not IPv6 addresses like 2001:db8::ffff:1.2.3.4 so I opted to leave it that way, but at least with this change both code paths (container creation and container autoboot) have the same wrong implementation. 

That means if container creation works, container autoboot should also work. Not like before where container creation might work but then container autoboot fails. 

But still, this is another thing that should probably be fixed some time in the future.

### DIFF
--- a/etc/rc.d/rc.docker
+++ b/etc/rc.d/rc.docker
@@ -196,7 +196,7 @@ start_network(){
       THIS_IP=
       while read_dom; do
         [[ $ENTITY == Network ]] && THIS_NETWORK=$CONTENT
-        [[ $ENTITY == MyIP ]] && THIS_IP=${CONTENT// /} && THIS_IP=${THIS_IP//,/;}
+        [[ $ENTITY == MyIP ]] && THIS_IP=${CONTENT// /,} && THIS_IP=$(echo "${THIS_IP}" | tr -s "," ";")
       done <$XMLFILE
       # only restore valid networks
       if [[ -n $THIS_NETWORK ]]; then
@@ -367,7 +367,7 @@ start_network(){
         THIS_TT=${CONNECT#*,}
         THIS_IP=
         for IP in ${THIS_TT//;/ }; do
-          [[ $IP =~ '.' ]] && THIS_IP="$THIS_IP --ip $IP" || THIS_IP="$THIS_IP --ip6 $IP"
+          [[ $IP =~ ':' ]] && THIS_IP="$THIS_IP --ip6 $IP" || THIS_IP="$THIS_IP --ip $IP"
         done
         docker network connect $THIS_IP $NETWORK $THIS_ID >/dev/null
       done


limetech:
Nice improvements, thank you!

thezoggy:
About ::ffff:<v4>

It’s IPv4 expressed through IPv6 so that the application only needs to support one IP stack. 

As it can be a lot more efficient to just do v6 rather than a true dual stack setup.

Leseratte10:
I do know about that, but Docker apparently doesn't. Assuming that Docker was implemented correctly, ::ffff:1.2.3.4 should be handled as an IPv6 by the IPv6 stack and eventually actually represent the IPv4 1.2.3.4. However, when you want to add this IP (::ffff:1.2.3.4) to a Docker container, you do need to use `--ip` instead of `--ip6` which means UnRAID would theoretically need special code to support that.

thezoggy:
> These addresses aren't that uncommon, and they don't need to be routable. It's a way for an IPv6-only app on a Dualstack system to send IPv4 packets. The OS will turn any IPv6 packets sent to that network into IPv4 packets. No support from the provider is needed.
> 
> Such an address is supposed to be handled like an IPv6, but the OS will create IPv4 packets. Unfortunately Docker treats this as an IPv4 instead so using these in UnRAID would require additional changes.
> 
> If you use an IPv6-capable app to send IPv6 packets to ::ffff:1.2.3.4, your provider doesn't need to do anything special. The OS will automatically convert these packets into IPv4 packets to 1.2.3.4. However, when you use `--ip6 ::ffff:1.2.3.4` on Docker, you get an error that --ip6 can't take IPv4 addresses. My assumption is that Docker internally first turns that address into an IPv4 and then parses the parameters. This means UnRAID would need to detect that prefix and then use --ip instead of --ip6 to support this syntax.
> 
> EDIT: Seems like a better idea to get this fixed upstream - meaning, make Docker accept such an address with the --ip6 parameter, too.

looks like this is by design:
https://github.com/moby/moby/blob/23.0/integration-cli/docker_cli_network_unix_test.go#L1719
- https://github.com/moby/moby/commit/1e6eccae69bd9b1f65ec87c33a6a872c81f7fb23

PR #1351, opened 2023-06-04
Fix VM page load issue

SimonFair:
Fix issues with VM page loads if users have removed vcpu pinning.

in_array fails due to vcpu being false and not an array.


limetech:
thank you!

PR #1353, opened 2023-06-06
wireguard: Fix IPv6-only VPN

Leseratte10:
Since Wireguard VPNs have been introduced, it looks like nobody has ever tested them with an IPv6-only VPN. 

As I already [reported on the forums](https://forums.unraid.net/bug-reports/stable-releases/6115-wireguard-not-working-with-ipv6-r2387/), when you create a Wireguard tunnel and select "IPv6 only" for the "Network Protocol", the tunnel doesn't start because the PostUp and PostDown commands are corrupted (see [this](https://github.com/limetech/webgui/blob/1f32774d7d729701375c2d9c4a6ba0f3fbed4ae5/emhttp/plugins/dynamix/include/update.wireguard.php#L235)). 

PostUp and PostDown commands are being created for IPv4 routes, with an empty gateway IP, which obviously doesn't work. 

With this change (not adding IPv4 routes if there's no IPv4 gateway or no IPv4 tunnel IP) at least lets me successfully start a Wireguard tunnel with IPv6. 

I'm marking this as a draft PR (so please don't merge yet) because I haven't tested everything with this v6-only VPN and I'm not yet sure if this change doesn't break other use cases.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of WireGuard Docker network routing by ensuring routes are only added when gateway and local network information is available, preventing potential misconfigurations.

### DIFF
--- a/emhttp/plugins/dynamix/include/update.wireguard.php
+++ b/emhttp/plugins/dynamix/include/update.wireguard.php
@@ -119,9 +119,12 @@ function addDocker($vtun) {
   }
   if (!$error && !isNet($network)) {
     [$device,$thisnet,$gateway] = thisNet();
-    exec("ip -4 rule add from $network table $index");
-    exec("ip -4 route add unreachable default table $index");
-    exec("ip -4 route add $thisnet via $gateway dev $device table $index");
+    // TODO: Add IPv6 support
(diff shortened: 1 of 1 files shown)


ljm42:
This draft is pretty out of date, I'm going to go ahead and close it

PR #1359, opened 2023-06-17
rc.library: read IP addresses directly from interfaces instead of file

bergware:


limetech:
@bergware  - what are "dashboard fixes" vs "dashboard enhancements"? Should the "enhancements" be encapsulated in a separate PR?

bergware:
In this PR the dashboard has only fixes, I removed the enhancements.

PR #1373, opened 2023-07-04
Miscellaneous fixes and enhancements

bergware:


limetech:
@bergware - there is a regression introduced in rc.docker.  The function is_docker_running() checks for the existence of the /var/run/docker.sock socket to determine if docker is running.  This introduces a race condition however.  As soon as 'rc.docker stop' exits, we proceed to unmount the docker image loopback file.  But this can fail if the docker process has not actually terminated yet and happens to still hold a file open in the image when the unmount attempt occurs.  Introducing a retry upon failed unmount is a workaround, but better to fix the issue in rc.docker by ensuring that the docker process has truly terminated.

PR #1375, opened 2023-07-10
Share Edit: make minimum free space settings always available

bergware:


limetech (commented):
How about line 151/153 - need to remove that test as well?

bergware:
Lines 151/153 needs to be completely removed.
Thanks

PR #1379, opened 2023-07-16
rc.docker: create same IPv6 network for containers and services

bergware:


bergware:
This also fixes a regression error in Unraid 6.12.3 where sometimes docker containers don't get an IPv6 subnet

PR #1390, opened 2023-07-25
Add ntp.conf, sshd.config, and servers.conf; apply maskIP to testparm…

dlandon:
Add some additional configuration files:
- ntp.conf
- sshd.config
- servers.conf

Apply maskIP to all new config files.  Apply maskIP to testparm output as it contains a listen IP list.  Fix php warning about braces around variables.


bergware:
@dlandon 
Thanks. Can you make two adjustments
1. Make maskIP conditional (user can also create non-anonymized diagnostics) , like: if (!$all) maskIP...
2. /etc/... files are in linux format, they need to be converted to windows format (you can use todos for this)

dlandon:
The following changes were mede in the PR update:
- Make sure all files have newline changed to lf, cr by calling newline() routine.
- Fix newline() routine not executing 'todos' correctly.
- Exit from maskIP() routine if the all (-a) option is set so IP addresses don't get masked.
- Change updated date to using the file mtime for the correct update date.

PR #1391, opened 2023-07-26
Remove upgrade date.

dlandon:
The vfat file system will not give us the date from the file when the upgrade was done.


dlandon:
The changes to Docker Settings are experimental to see if detecting of the promiscuous mode is supported by the network and disabling macvlan will help with support.

PR #1417, opened 2023-08-22
docker: fix routing when "host access" is enabled

bergware:


bergware:
This fixes a regression error with shim interfaces (when bridging is enabled)

PR #1418, opened 2023-08-24
CSS: set overflow-x to 'auto'

justinnichols:
When overflow-x is set to 'scroll',
in some browsers, it will reserve
space for the scrollbar and will
override the background color even
when no scrollbar is necessary.


justinnichols:
macos_firefox_116 0 3_unraid_overflowx

Proof of this issue with MacOS Ventura and Firefox 116.0.3.

bergware:
Thanks

PR #1421, opened 2023-09-07
Provide target string in rc.nginx for plugin configurations

dkaser:
Tailscale includes a built-in interface which helps users connect and configure Tailscale. This interface is provided via CGI. The Tailscale plugin for Unraid uses a PHP script to wrap the CGI interface and present it as part of the WebGUI.

Tailscale is adding functionality to the web interface, which will soon require that path information be sent with each request. The default nginx configuration does not provide this information.

This change adds a location entry to the nginx configuration that will send the path information for requests to `/plugins/tailscale/interface.php`. I have tested this change with the dev version of Tailscale and confirmed that it works correctly. Since the location block is scoped to `/plugins/tailscale/interface.php`, it will not affect any other functionality in nginx.

Normally, I would try to add the needed entry during plugin installation. However, in this case, doing so would require patching `/etc/rc.d/rc.nginx` during install, which could result in problems for either the plugin or the WebGUI when the contents of `/etc/rc.d/rc.nginx` change in the future. Therefore, I think it is better to include the configuration upstream.

Ref: dkaser/unraid-tailscale#86

### DIFF
--- a/etc/rc.d/rc.nginx
+++ b/etc/rc.d/rc.nginx
@@ -371,6 +371,9 @@ build_locations() {
 	    add_header Cache-Control no-cache;
 	}
 	#
+	# Plugins: add custom entries to locations.conf here
+	#
+	#
 	# pass PHP scripts to FastCGI server listening on unix:/var/run/php5-fpm.sock
 	#
 	location ~ \.php$ {


dkaser:
I'm OK with that approach. I updated the PR to reflect that idea.

ljm42:
Looks good to me, although is there a particular reason you placed it where you did as opposed to at the end of the file? In terms of support, it might help if the customizations were at the end

dkaser:
Yes, plugin configurations would generally need to match before the global php match (`location ~ \.php$ {`).

ljm42 (approved):

ljm42:
I hope you don't mind but I'm changing direction a bit here...

So we've been reviewing https://github.com/dkaser/unraid-tailscale/issues/86 along with http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_split_path_info

What I've done locally is modify /etc/rc.d/rc.nginx like this:

```
#
# pass PHP scripts to FastCGI server listening on unix:/var/run/php5-fpm.sock
#
# location ~ \.php$ {
#      include fastcgi_params;
# }
location ~ ^(.+\.php)(.*)$ {
     fastcgi_split_path_info  ^(.+\.php)(.*)$;
     fastcgi_param PATH_INFO  $fastcgi_path_info;
     include fastcgi_params;
}
```

I also created a basic plugins/dynamix/phpinfo.php file:

```
<?php 
phpinfo();
?>
```

And when I call it I can see that the expected vars are set appropriately:

```
/plugins/dynamix/phpinfo.php?a=1&b=2
$_SERVER['REQUEST_URI']  /plugins/dynamix/phpinfo.php?a=1&b=2
$_SERVER['QUERY_STRING'] a=1&b=2
$_SERVER['PATH_INFO']    no value

/plugins/dynamix/phpinfo.php/abc/def
$_SERVER['REQUEST_URI']  /plugins/dynamix/phpinfo.php/abc/def
$_SERVER['QUERY_STRING'] no value
$_SERVER['PATH_INFO']    /abc/def
```

Can you test this and confirm it does what you need?  You can either parse REQUEST_URI or PATH_INFO to get the values.

There are two ways forward here:

1) When your plugin is installed (either by the user or after a reboot) it could modify those lines in /etc/rc.d/rc.nginx . This will work on existing Unraid installs and is probably how we should start.

2) After more extensive testing we could modify the base rc.nginx to add this functionality to a future version of Unraid.

I've done some light testing in the webgui and since this is more of an addition instead of a change it looks safe. I can potentially see the webgui using this down the road so there could be value in making this change to the webgui itself.

dkaser:
I'll give that a try on my test system. I expect that config to work just fine... I was just trying to avoid doing that because I didn't want to be making a change that affects the rest of the webgui. However, if you would prefer I make the change that way, I'm OK with doing that as part of my install. It's honestly easier that way... I can just include a patch file with my install and run it against rc.nginx.

It gets me a working config, and you a group of testers for the global change :)

dkaser:
My initial testing looks good. I'll plan to apply the following patch (or something similar to it) as part of the Tailscale plugin install once the new client is ready for release:

```
--- rc.nginx.old        2023-09-08 17:37:50.793527490 -0400
+++ rc.nginx    2023-09-08 17:42:40.432852263 -0400
@@ -373,8 +373,10 @@
        #
        # pass PHP scripts to FastCGI server listening on unix:/var/run/php5-fpm.sock
        #
-       location ~ \.php$ {
-           include fastcgi_params;
+       location ~ ^(.+\.php)(.*)$ {
+            fastcgi_split_path_info  ^(.+\.php)(.*)$;
+            fastcgi_param PATH_INFO  $fastcgi_path_info;
+            include fastcgi_params;
        }
        #
        # enable compression of JS/CSS/WOFF files
```

ljm42:
Sounds good, I'll go ahead and close this. Let me know how it goes and we can look at including it by default in the future.

PR #1422, opened 2023-09-08
Add readme and VS Code settings

ljm42:


BaileyJM02:
Thanks for the addition!

.vscode/sftp.template
BaileyJM02, line 1: This would be easier to read if it was named `sftp-template.json` due to correct JSON colour coding within the editor

BaileyJM02 (changes requested):

.vscode/sftp.template
ljm42: great idea!

PR #1424, opened 2023-09-08
refactor(upc): vue3 + ts web components & improved supporting php

zackspear:
* Top right dropdown and the various supporting web components have been rebuilt using Vue3 + TypeScript
* Goodbye pop-up windows for key management and hello more reliable callback system
* Supporting PHP for the web components have also been refactored to improve developer experience


zackspear:
@ljm42 let's test this together early next week. In my testing I didn't notice any issues. But would be great to have someone else test these changes before this is merged.

zackspear:
@BaileyJM02, yeah the web components like the top right dropdown for key management (aka UPC – user profile component) are the same web components that are installed via the Unraid Connect plugin. But without any usage of the `unraid-api` graph endpoints / cloud service connectivity. Thankfully Vue makes it pretty easy to make these components dual purpose.

The Vue3 + TS source files for these web components are part of a private repo of ours for the Connect plugin. We've discussed eventually making that repo public but don't have any timeline yet.

So my workflow here is running a build script for these components locally that then copies the build into my forked `webgui` repo. Which then `rsync`s to my dev server to test without the plugin installed and then if all is good I'll commit the built files to the repo.

BaileyJM02:
Ah that makes more sense, sounds good. I think leaving the script as part of the private repo are the way forward. Possibly adding a Git submodule to this repo linking to the built folder of the plugin etc

ljm42 (approved):

PR #1432, opened 2023-09-25
Add kill routine

ich777:
- Add kill routine to force_stop

@ljm42 in addition to your add-logging PR I added a `kill_docker` function to `force_stop` if `docker_stop` wasn't successful.


ich777:
Not fully satisfied with this solution.

PR #1441, opened 2023-10-06
rc.nginx: add safety checks when run from webterminal

ljm42:


ljm42:
Not worth the complication

PR #1443, opened 2023-10-07
Descriptive error messages on install_key

Squidly271:


Squidly271:
Chose not to add in the possibility of DNS settings to code 4 on plugins due to wording awkwardness of telling user to post in forum / discord / reddit etc.

PR #1444, opened 2023-10-08
DRAFT Run FCP scan during diagnostics

Squidly271:
Pending Larry approval

Runs FCP scan if one hasn't been already run within 24 hours.  (FCP by default logs issues in syslog.  If user does disable logging then FCP still logs "issues found, but not logged due to user choice)


ljm42:
Sorry for the delay, this looks great!

PR #1446, opened 2023-10-09
include USB-mirrored syslogs in diagnostics

desertwitch:
Since we already have the option to mirror the SYSLOG to USB (`Settings->Syslog Server->Mirror syslog to flash`) for post mortem (after crash/shutdown) diagnostics, I propose those should also be included in the diagnostics package. This can be particularly helpful for diagnostics where users keep experiencing hard resets of their UNRAID systems for reasons unknown. Users can then simply be instructed to turn on `Mirror syslog to flash`, wait for another system crash and afterwards fetch a diagnostics package - now also including the USB-mirrored SYSLOG from before the system crash.


desertwitch:
moved to separate branch and re-opened as new PR

PR #1447, opened 2023-10-09
also show USB-mirrored syslogs in GUI

desertwitch:
Since we already have the option to mirror the SYSLOG to USB (`Settings->Syslog Server->Mirror syslog to flash`) for post mortem (after crash/shutdown) diagnostics, I propose users should also be able to view those in the GUI (on the SYSLOG page) when such mirrored logfiles exists on the USB. This should be useful when users want to know what was going on before a system shutdown without having to manually extract the relevant mirrored SYSLOG files from the USB themselves.

**This integrates seamlessly into the already present dropdown code used for when multiple log files are present.**

only regular SYSLOG present
grafik

regular SYSLOG + mirrored SYSLOG present on USB
grafik

regular SYSLOG + mirrored SYSLOG present on USB + other SYSLOGs present in server folder (i.e. from log rotation)
grafik

### DIFF
--- a/emhttp/plugins/dynamix/Syslog.page
+++ b/emhttp/plugins/dynamix/Syslog.page
@@ -30,6 +30,19 @@ if (file_exists($cfg)) {
     foreach ($logs as $file) $select[] = mk_option(1,$file,basename($file));
     $select[] = "</select>";
   }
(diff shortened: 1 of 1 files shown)


ljm42:
Thanks for your PRs! I agree there are several aspects of USB-mirrored logs that could use improvement. I took inspiration from you and came up with this: https://github.com/unraid/webgui/pull/1449

PR #1448, opened 2023-10-09
include USB-mirrored syslogs in diagnostics

desertwitch:
Since we already have the option to mirror the SYSLOG to USB (`Settings->Syslog Server->Mirror syslog to flash`) for post mortem (after crash/shutdown) diagnostics, I propose those should also be included in the diagnostics package. This can be particularly helpful for diagnostics where users keep experiencing hard resets of their UNRAID systems for reasons unknown. Users can then simply be instructed to turn on `Mirror syslog to flash`, wait for another system crash and afterwards fetch a diagnostics package - now also including the USB-mirrored SYSLOG from before the system crash. Users would no longer need to manually extract those files from their USB drives.

### DIFF
--- a/emhttp/plugins/dynamix/scripts/diagnostics
+++ b/emhttp/plugins/dynamix/scripts/diagnostics
@@ -640,6 +640,36 @@ foreach (glob("/var/log/syslog*") as $file) {
   run("truncate -s '<$max' ".escapeshellarg("$log.txt"));
 }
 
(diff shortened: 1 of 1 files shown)


ljm42:
Thanks for your PRs! I agree there are several aspects of USB-mirrored logs that could use improvement. I took inspiration from you and came up with this: https://github.com/unraid/webgui/pull/1449

PR #1449, opened 2023-10-09
Improved handling of USB-mirrored syslogs

ljm42:
This solves three issues:
* mirrored logs are now cleaned up on each boot, rather than growing forever
* mirrored logs from the previous boot are now included in diagnostics
* mirrored logs from the previous boot are now available in the webgui on Tools -> Syslog


desertwitch:
Looking real good, closing #1447 and #1448 in favor of these changes.

PR #1462, opened 2023-10-13

bergware:


bergware:
DO NOT PANIC.
The PR which moves inline styling of page files to dedicated css files, involves many files, but is basically copying styling content from page file to css file.

PR #1470, opened 2023-10-18
fix: account for table visibility cookies left in broken state

desertwitch:
After an unexpected JS interruption or the too rapid toggling of the visibility of individual/multiple dashboard panels, the `hidden_content` cookie can be left in a broken state by the browser. This means the cookie then includes either duplicate md5 values, empties or null values (i.e. due to leading/trailing delimiters) which - at present - break that entire functionality and force the unwanted showing of all (set to hidden) dashboard panels. At present the cookie needs to be removed to regain its function.

This PR accounts for such broken cookie states. The cookie is now cleaned up (from duplicates, empties and nulls) before being rewritten every time the user toggles the visibility of an individual panel element and, as an additional safe-guard, before evaluation of which panel elements should be hidden or displayed when the dashboard is first loaded.

### DIFF
--- a/emhttp/plugins/dynamix/DashStats.page
+++ b/emhttp/plugins/dynamix/DashStats.page
@@ -1223,7 +1223,7 @@ function openClose(button) {
     } else {
       button.removeClass('fa-chevron-up fa-chevron-down').addClass('fa-chevron-up');
       tbody.mixedView(1);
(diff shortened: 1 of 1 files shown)


desertwitch:
thought of a better solution, will open new PR when done

PR #1486, opened 2023-10-31
Custom rsyslog config files

ich777:
This allows to create custom config files for rsyslog in `/boot/config/rsyslog.d/` and link them on boot in the directory `/etc/rsyslog.d/`


limetech:
Nice feature

PR #1498, opened 2023-11-09
Update rc.S.cont

ich777:
Only execute rc.modules if unraidsafemode is not set


limetech:
If someone is relying on third party network driver then booting in 'safe mode' with this change will prevent webgui/ssh access.

ich777:
Yes that's correct and why it's marked as draft.
Maybe we should add an exception for certain drivers or should drivers be installed regardless if unRAID is boot into "safe mode" or not?

ich777:
@limetech change done, thanks for catching that!

PR #1503, opened 2023-11-13
Local Console settings

ich777:
- Create page for Local Console
- Create console script
- Create rc.keymap and modify default ot be compatible with local console settings
- Change rc.setterm to be compatible with local console settings
- Change rc.local to be compatible with persistent bash history

### DIFF
--- a/emhttp/plugins/dynamix/Console.page
+++ b/emhttp/plugins/dynamix/Console.page
@@ -95,8 +95,8 @@ _(Local screen blank time)_:
 
 _(Persistent Bash History)_:
 : <select name="persist_bash_history" class="narrow">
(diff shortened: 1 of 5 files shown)


ich777:
This is based on the work from @bergware and also adds a persistent bash history if the user decides to enable it.

grafik

grafik

bergware:
Can you add translation support to the choices:

_(Persistent Bash History)_:
: <select name="persist_bash_history" class="narrow">
   <?=mk_option($console['persist_bash_history'], "0", _('Disabled'))?>
   <?=mk_option($console['persist_bash_history'], "1", _('Enabled'))?>
</select>

PS. There is the function "mk_option" which makes it easer to define the select options (see above).

ich777:
@bergware thanks for the suggestion and hint! Changed the PR already.

PR #1525, opened 2023-11-25
Backport 6.12 NVME Fix for selftest

SimonFair:


bergware:
The final implementation in 6.13 differs from this PR.
Better close it and if a backport is needed, it should be based on the latest 6.13 implementation

PR #1526, opened 2023-11-26
Don't send nchan messages if no subscribers are listening.

dlandon:
If no subscriber is listening, there is no reason to post an nchan message.  This just backs up messages and clogs nchan with messages that won't ever be delivered.


bergware:
6.13 implementation has built-in protection at the end points as a substitute of this PR.
This PR can be closed.

PR #1533, opened 2023-12-01
Fix for btrfs and xfs boot

ich777:
- Use first partition for btrfs and xfs boot, this allows a btrfs mirror to boot
- Use `degraded` for btrfs mount to allow boot from degraded btrfs pools

### DIFF
--- a/etc/rc.d/rc.S
+++ b/etc/rc.d/rc.S
@@ -97,10 +97,10 @@ find_device && /bin/echo "found $DEVICE" || abort "not found"
 # detect filesystem from boot device
 if /sbin/blkid -s TYPE $DEVICE | /bin/grep -q "btrfs" ; then
   NONVFAT=btrfs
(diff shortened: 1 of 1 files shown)


ich777:
@limetech I'm hitting a bit of a strange roadblock if using a btrfs mirror.

1. Device with correct GUID is /dev/sda -> WebGUI assumes that /dev/sdb is the correct device
grafik


2. Device with incorrect GUID is /dev/sdb -> WebGUI assumes that /dev/sda is the correct device
grafik

I've even implemented a check in rc.S which determines the correct drive to boot from with:
`btrfs filesystem show $UNRAIDLABEL | awk '/devid/{print $2,$8}' | sort -n -k1 | head -1 | awk '{print $2}'`
but even that won't help and the GUI (or I think emhttp) always assumes that the wrong device is the registered one.
I removed that check because that would need additional libraries in `bzroot`, only did that for testing.

If I boot only with the registered device it is working fine:
grafik

as long as I don't connect the second mirror device:
grafik


I'm assuming this is because of how `emhttp` detects the USB Flash device and that it was never designed to take two `UNRAID` devices into account.

PR #1543, opened 2023-12-11
styling enhancements to improve consistency

bergware:


bergware:
This PR show tables with a consistent width of their columns.
If content is too wide for a column it gets automatically truncated by the browser (indicated by ...)
This also applies to translated content which may be wider than the English content.

This settings become noticable when using a mobile device or screen with lower resolution which does not have enough space to show all content and hence truncation is required.

PR #1552, opened 2023-12-15
Display KB/MB/GB/TB written in SMART Attributes for SSDs

serisman:
Many SSDs report how many bytes have been written to and/or read from the flash using some combination of:
 * Total_LBAs_Written
 * Total_LBAs_Read
 * Host_Writes_32MiB
 * Host_Reads_32MiB
 * TLC_Writes_32MiB

The raw value is a bit hard to grok at a glance, so this PR adds in the calculated KB/MB/GB/TB values (similar to how Power on hours is modified to show years/months/days).

Before:
```
241 | Total lbas written | 0x0032 | 099 | 099 | 000 | Old age | Always | Never | 86530637427
...
241 | Host writes 32mib | 0x0000 | 100 | 100 | 000 | Old age | Offline | Never | 18546
242 | Host reads 32mib | 0x0000 | 100 | 100 | 000 | Old age | Offline | Never | 9189
245 | TLC writes 32mib | 0x0000 | 100 | 100 | 000 | Old age | Offline | Never | 12764
```

After:
```
241 | Total lbas written | 0x0032 | 099 | 099 | 000 | Old age | Always | Never | 86530637427 (44.3 TB)
...
241 | Host writes 32mib | 0x0000 | 100 | 100 | 000 | Old age | Offline | Never | 18546 (622 GB)
242 | Host reads 32mib | 0x0000 | 100 | 100 | 000 | Old age | Offline | Never | 9189 (308 GB)
245 | TLC writes 32mib | 0x0000 | 100 | 100 | 000 | Old age | Offline | Never | 12764 (428 GB)
```

Note:  This assumes an LBA is 512 bytes, which is true for most SSDs.  There might be some out there, however, that use a different size LBA.  Maybe it should be looked up via the general SMART values?

### DIFF
--- a/emhttp/plugins/dynamix/include/SmartInfo.php
+++ b/emhttp/plugins/dynamix/include/SmartInfo.php
@@ -38,6 +38,9 @@ function duration(&$hrs) {
   $age = date_diff($poh,$now);
   $hrs = "$hrs (".($age->y?"{$age->y}y, ":"").($age->m?"{$age->m}m, ":"").($age->d?"{$age->d}d, ":"")."{$age->h}h)";
 }
+function blocks_size(&$blks,$blk_size) {
+  $blks = "$blks (".my_scale($blks*$blk_size,$unit)." $unit)";
+}
 function append(&$ref, &$info) {
   if ($info) $ref .= ($ref ? " " : "").$info;
 }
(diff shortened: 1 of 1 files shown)


bergware:
Please use the function "my_scale()" to display numbers

serisman:
Thanks for that.  PR updated.  I should have guessed there was already a function to do that.  :)

limetech:
thank you!

PR #1553, opened 2023-12-18
6.12 Fix for downgrade

SimonFair:
When downgrading if the VM template has been updated to the latest QEMU machine type it will not be found. It will default to i440fx which can create an issue if the machine was Q35.

This PR finds latest current version for a given machine type.


limetech:
good idea

PR #1555, opened 2023-12-19
Docker: Allow custom registry with a port specification

rdietl:
### Description of the problem

Custom Docker registries may be hosted on non-standard-ports [1]. Images stored in such a registry can be used in Unraid by providing a "repository" name such as "my.registry.dev:5002/myproject/myservice:latest".

However, the Docker GUI of Unraid does not properly support such image names, because it wrongly assumes that an image may only ever contain one ":" character.

### Suggested solution and description of the PR

This PR provides a rewritten version of `DockerUtil::ensureImageTag` that properly handles image names that contain custom registries with a port specification. This includes all existing "magic" additions (like prefixing images with "library/" if necessary, and adding ":latest" if missing).

Furthermore, a new method `DockerUtil::parseImageTag` is introduced, which properly splits an image name into "strRepo" and "strTag", thus removing the need for the various occurrences of `explode("...", DockerUtil::ensureImageTag(...))` in the current code.

### Reproduction instructions

- Inside an Unraid console, follow the guide at [2] to create a custom docker registry based on Docker's "distribution" registry and push an image to that registry.
- Use the Unraid Docker plugin to create a new container, pointing the "repository" to the image in the custom registry that was just created

#### Expected

The Docker GUI in Unraid shows "latest" as the container's tag, and whether there are any updates available for the container.

#### Actual

The Docker GUI shows something like "5000/alpine" as tag, and status "not available".

### Notes

[1] https://docs.docker.com/engine/reference/commandline/tag/
[2] https://github.com/distribution/distribution/blob/main/docs/dockerhub.md

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -428,7 +428,7 @@ class DockerUpdate{
 
 	// DEPRECATED: Only used for Docker Index V1 type update checks
 	public function getRemoteVersion($image) {
-		[$strRepo, $strTag] = array_pad(explode(':', DockerUtil::ensureImageTag($image)),2,'');
+		extract(DockerUtil::parseImageTag($image));
 		$apiUrl = sprintf('http://index.docker.io/v1/repositories/%s/tags/%s', $strRepo, $strTag);
 		//$this->debug("API URL: $apiUrl");
 		$apiContent = $this->download_url($apiUrl);
@@ -436,7 +436,7 @@ class DockerUpdate{
 	}
 
 	public function getRemoteVersionV2($image) {
-		[$strRepo, $strTag] = array_pad(explode(':', DockerUtil::ensureImageTag($image)),2,'');
+		extract(DockerUtil::parseImageTag($image));
 		/*
 		 * Step 1: Check whether or not the image is in a private registry, get corresponding auth data and generate manifest url
 		 */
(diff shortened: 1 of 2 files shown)


rdietl:
You can use the following code to test the various scenarios of image names allowed agains a DockerUtil class. The current implementation fails the lower set of tests, while the suggested implementation passes.

```
<?php

require '/path/to/DockerClient.php';

function test($candidate, $expected)
{
    $actual = DockerUtil::ensureImageTag($candidate);

    $ok = "OK";
    $er = "!!";
    if ($expected === $actual) {
        echo "[$ok] Candidate: " . $candidate . " => " . $actual . "<br/>\n";
    } else {
        echo "[$er] Candidate: " . $candidate . "<br/>\n";
        echo "[$er] Expected: " . $expected . "<br/>\n";
        echo "[$er] Actual: " . $actual . "<br/>\n";
    }

    echo "-----------------------------------------------<br/><br/>\n\n";
}

// Normal case 1: namespace/image
test('namespace/image', 'namespace/image:latest');
test('namespace/image:tag', 'namespace/image:tag');

// Normal case 2: library/*-image
test('image', 'library/image:latest');
test('image:tag', 'library/image:tag');

// Normal case 3: sha256:
test('sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08', '9f86d081884c:latest');

// New case: with port
test('foo.bar.com:5002/image', 'foo.bar.com:5002/image:latest');
test('foo.bar.com:5002/namespace/image', 'foo.bar.com:5002/namespace/image:latest');
test('foo.bar.com:5002/namespace/image:tag', 'foo.bar.com:5002/namespace/image:tag');
test('foo.bar.com:5002/namespace1/namespace2/image', 'foo.bar.com:5002/namespace1/namespace2/image:latest');

```

limetech:
thank you!

PR #1563, opened 2023-12-28
Add syslog filename support

Leseratte10:
I've tried using the built-in Syslog server in UnRAID and found two issues: 

A) the HTML input field for the syslog server is limited to 23 characters, my hostname is longer. I increased it to 50 and it works fine.

B) as stated [on the forums](https://forums.unraid.net/topic/150195-unraid-syslog-server-use-device-name-for-filename-instead-of-ip/), I don't like how UnRAID adds the client's IP address to the syslog file name. 

Some clients may switch between IPv4 and IPv6, or they have IPv6 privacy extensions, or they switch between different IPv6 networks. All of which will result in tons of different logfiles for each client. 

To work around this issue, I added a new config field that allows people to enter a template for the syslog file name instead of hardcoding it to `syslog-%FROMHOST-IP%.log`. By default, it still uses this path so nothing should change for existing configurations - but if needed, it can now be customized to something like `syslog-%HOSTNAME%.log` to use the host name sent in each syslog message for the filename instead.

### DIFF
--- a/emhttp/languages/en_US/helptext.txt
+++ b/emhttp/languages/en_US/helptext.txt
@@ -1493,12 +1493,21 @@ This will increase write performance but might possibly decrease read performanc
 Let the server act as a central syslog server and collect syslog messages from other systems.
 The server can listen on UDP, TCP or both with a selectable port number.
 
-Syslog information is stored per IP address. That is every system gets its own syslog file.
+Syslog information is stored either per IP address or per hostname. That is every system gets its own syslog file.
 :end
 
 :syslog_local_folder_help:
 Select the share folder where the syslogs will be stored.
 It is recommended that you use a share located on the cache drive to prevent array disk spinups.
+:end
+
+:syslog_remote_system_identifier_help:
+Select the identifier for the remote system (used in the logfile name). 
+
+* "IP Address" uses the IP address (IPv4 or IPv6) of the sending system.
+* "Hostname (from syslog message)" uses the hostname included in each syslog message.
+* "Hostname (from DNS reverse lookup)" performs a DNS reverse lookup for the sending IP and uses the result.
+
 :end
 
 :syslog_local_rotation_help:
--- a/emhttp/plugins/dynamix/SyslogSettings.page
+++ b/emhttp/plugins/dynamix/SyslogSettings.page
@@ -108,6 +108,15 @@ _(Local syslog folder)_:
 
 :syslog_local_folder_help:
 
+_(System identifier for logfile name)_:
+: <select name="server_filename">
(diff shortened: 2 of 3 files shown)


emhttp/languages/en_US/helptext.txt
```diff
@@ -1501,6 +1501,14 @@ Select the share folder where the syslogs will be stored.
 It is recommended that you use a share located on the cache drive to prevent array disk spinups.
 :end
 
+:syslog_local_filename_help:
+Filename template used for naming syslog log files. 
+By default, UnRAID includes the client's IP in the filename.
```
ljm42, line 1506: Please use "Unraid" not "UnRAID"

ljm42:
With this change, the files will no longer be guaranteed to start with "syslog-". Do you have any concerns with this @bergware ?

bergware:
The logs folder may contain different logs, we need 'syslog-' as an identifier to make a selection.

Leseratte10:
The idea was to give people free choice about the log file name, but if that's not a good idea I'm happy to find another solution. 

How about reverting the glob back to syslog-*.log and adding a note to the help that only logfiles starting with "syslog-" will be displayed in the Unraid log viewer?

Or should I just make the "*" part customizable and enforce that all syslog files have to start with "syslog-"?

ljm42:
> I can also just make it a toggle in the GUI to switch between syslog-%FROMHOST-IP%.log and syslog-%HOSTNAME%.log and not allow user input. 

Yes, this is the right approach. Unraid hides the gory details of Linux behind a friendly webgui, we need to provide appropriate options in a dropdown not provide a freeform field.  Similar to how you can the date and time format from dropdowns, not by entering raw codes into a textbox. Sorry I did not suggest this before.

I'm thinking something like:
`Remote system identifier: [ip|name]`

This field would also need help text, and the help text for "Local syslog server" needs updating because it mentions that logs are stored by IP address.

Leseratte10:
Okay, I'll try to add that and update this PR when I'm done.

Leseratte10:
Hm, I added that on my test server to see how it looks but I don't like the text "Remote system identifier". That sounds too close to "Remote syslog server" and doesn't sound like it'd be an option / a setting for the local syslog server. How about just "System identifier", or "System identifier for local logfile name"?

ljm42:
Thanks for making these changes, looks pretty good to me

image

Leseratte10:
Anything that's still needed from me to get this merged or is this just waiting for someone from Limetech to review and merge eventually?

starsoccer:
Not to get too off-topic on this PR but it would also be nice if unraid could support folders to better organize the syslog. I have about 30 devices all logging to unraid and with 5 file rotations, it becomes annoying scrolling through 150(30 devices * 5 files each) to find the device I want then the most recent

PR #1564, opened 2023-12-29
refactor: update os component use new server-side check

zackspear:
@ljm42, do we need to update the `unraidcheck` script in this repo? Or have you taken care of that already.


ljm42:
@zackspear the unraidcheck script needs to be part of this commit

PR #1582, opened 2024-01-12
Remove update buttons for non-dockerman containers

mtongnz:
Remove the update/force update buttons from the GUI for non dockerman containers.  Pressing them results in an error and updates for these containers should be done by the appropriate manager.


bergware:
This looks good, thanks

ashishpandey:
does this only remove the buttons from UI, or would this also handle fix common problems from reporting update being available?

mtongnz:
> does this only remove the buttons from UI, or would this also handle fix common problems from reporting update being available? Ideally, it would be nice if the update available state isnt computed at all

This only removes the buttons.
The "updates available" is to do with how unraid stores the current version of each installed container.  If you're using Compose Manger, I submitted a PR which hasn't been implemented yet: https://github.com/dcflachs/compose_plugin/pull/25

Issue #1569, opened 2024-01-03
Docker Compose - Network info on GUI displays incorrectly


mtongnz:
The unRaid GUI has changed how it handles displaying network info for docker containers.  This means that containers created outside of the GUI (like when using the Docker Compose Manager plugin) display a network ID instead of the actual network name.
See this post: https://forums.unraid.net/topic/114415-plugin-docker-compose-manager/?do=findComment&comment=1340206

I have started work on a solution: https://github.com/mtongnz/unRaid-webgui/tree/network-display-improvements
This will display all network names and the associated IP for each.

I'd like thoughts and suggestions before creating a PR.

dcflachs:
I am not sure i like the idea of using the net.unraid.docker.managed label to special case the display of networks in dockerman. It would be preferable to have a more general solution to fix displaying network names. Especially since it may encourage people to use net.unraid.docker.managed=composeman for any container that has multiple networks, whether or not it was created by the compose.manager plugin. Any container can be connected to multiple networks, and those launched by other means, portainer, pterodactyl, command line, etc. should not use the net.unraid.docker.managed=composeman.

mtongnz:
That makes a lot of sense and I agree. I hadn't thought about that and was looking for the easiest solution to my issue.
Do you have a recommended solution?
Perhaps a label called net.unraid.docker.netdisplay.
It could then have multiple options available such as default, multiple, none...

dcflachs:
My suggestion would be no labels. I think the existing Network and Port Mappings display code in dockerman needs to be re-written to handle more complex network setups. I only had time to skim the existing code and your changes but it looks like you are on the right track. It seems that "NetworkSettings" is a better place to pull the network names from than what is currently used. It also looks like the "Ports" listing under "NetworkSettings" might be the first choice for pulling the port/IP pairs. Ports and IP pairs though is where things get complicated and likely requires some careful thought.

mtongnz:
I've pushed a new commit which addresses those issues.  It also removes the update links that don't work with non docker man containers.
I do feel that there may be some edge cases that the ports are not shown correctly but I don't know enough about docker networking to test it.  For example, does docker let you expose a port on a specific IP and not on others?  If it does, my UI tweaks won't indicate this.

mtongnz:
I've gone through and tidied commits into a new branch (updated above) as it was getting messy and I kinda started from scratch.

**Changes in DockerClient.php:**
 - a few minor style changes to make indents/spaces line up
[...]

PR #1616, opened 2024-02-10
Network info display improvements

mtongnz:
This PR aims to improve how networks are displayed for docker containers.

 - fix where network ID instead of a name is displayed (generally for docker compose created containers).
 - display multiple networks and the associated IP for each

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -920,7 +920,7 @@ class DockerClient {
 			$c['BaseImage']   = $ct['Labels']['BASEIMAGE'] ?? false;
 			$c['Icon']        = $info['Config']['Labels']['net.unraid.docker.icon'] ?? false;
 			$c['Url']         = $info['Config']['Labels']['net.unraid.docker.webui'] ?? false;
-			$c['Shell']         = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
+			$c['Shell']       = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
 			$c['Ports']       = [];
 			$c['Networks']	  = [];
 			if ($id) $c['NetworkMode'] = $net.str_replace('/',':',DockerUtil::ctMap($id)?:'/???');
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -1,7 +1,7 @@
 <?PHP
 /* Copyright 2005-2023, Lime Technology
-  * Copyright 2012-2023, Bergware International.
-* Copyright 2014-2021, Guilherme Jardim, Eric Schultz, Jon Panozzo.
+ * Copyright 2012-2023, Bergware International.
+ * Copyright 2014-2021, Guilherme Jardim, Eric Schultz, Jon Panozzo.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License version 2,
@@ -49,11 +49,11 @@ $port = file_exists('/sys/class/net/br0') ? 'BR0' : (file_exists('/sys/class/net
 // Docker configuration file - guaranteed to exist
 $docker_cfgfile = '/boot/config/docker.cfg';
 if (file_exists($docker_cfgfile)) {
-  exec("grep -Pom2 '_SUBNET_|_{$port}(_[0-9]+)?=' $docker_cfgfile",$cfg);
-  if (isset($cfg[0]) && $cfg[0]=='_SUBNET_' && empty($cfg[1])) {
-  # interface has changed, update configuration
-  exec("sed -ri 's/_(BR0|BOND0|ETH0)(_[0-9]+)?=/_{$port}\\2=/' $docker_cfgfile");
-}
+	exec("grep -Pom2 '_SUBNET_|_{$port}(_[0-9]+)?=' $docker_cfgfile",$cfg);
+	if (isset($cfg[0]) && $cfg[0]=='_SUBNET_' && empty($cfg[1])) {
+		# interface has changed, update configuration
+		exec("sed -ri 's/_(BR0|BOND0|ETH0)(_[0-9]+)?=/_{$port}\\2=/' $docker_cfgfile");
+	}
 }
 
 $defaults = (array)@parse_ini_file("$docroot/plugins/dynamix.docker.manager/default.cfg");
@@ -357,7 +357,7 @@ class DockerTemplates {
 		if (!$imgUrl) $imgUrl = $tmpIconUrl;
 		if (!$imgUrl || trim($imgUrl) == "/plugins/dynamix.docker.manager/images/question.png") return '';
 
-				$iconRAM = sprintf('%s/%s-%s.png', $dockerManPaths['images-ram'], $contName, 'icon');
+		$iconRAM = sprintf('%s/%s-%s.png', $dockerManPaths['images-ram'], $contName, 'icon');
 		$icon    = sprintf('%s/%s-%s.png', $dockerManPaths['images'], $contName, 'icon');
 
 		if (!is_dir(dirname($iconRAM))) mkdir(dirname($iconRAM), 0755, true);
@@ -1031,11 +1031,11 @@ class DockerUtil {
 				// Unprocessable input
 				$strRepo = $image;
 			}
-}
+		}
 
 		// Add :latest tag to image if it's absent
 		if (empty($strTag)) $strTag = 'latest';
-		
+
 		return array_map('trim', ['strRepo' => $strRepo, 'strTag' => $strTag]);
 	}
 
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -1,7 +1,7 @@
 <?PHP
 /* Copyright 2005-2023, Lime Technology
- * Copyright 2012-2023, Bergware International.
- * Copyright 2014-2021, Guilherme Jardim, Eric Schultz, Jon Panozzo.
+  * Copyright 2012-2023, Bergware International.
+* Copyright 2014-2021, Guilherme Jardim, Eric Schultz, Jon Panozzo.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License version 2,
@@ -50,10 +50,10 @@ $port = file_exists('/sys/class/net/br0') ? 'BR0' : (file_exists('/sys/class/net
 $docker_cfgfile = '/boot/config/docker.cfg';
 if (file_exists($docker_cfgfile)) {
   exec("grep -Pom2 '_SUBNET_|_{$port}(_[0-9]+)?=' $docker_cfgfile",$cfg);
-  if (isset($cfg[0]) && $cfg[0]=='_SUBNET_' && empty($cfg[1])) { 
-    # interface has changed, update configuration
-    exec("sed -ri 's/_(BR0|BOND0|ETH0)(_[0-9]+)?=/_{$port}\\2=/' $docker_cfgfile");
-  }
+  if (isset($cfg[0]) && $cfg[0]=='_SUBNET_' && empty($cfg[1])) {
+  # interface has changed, update configuration
+  exec("sed -ri 's/_(BR0|BOND0|ETH0)(_[0-9]+)?=/_{$port}\\2=/' $docker_cfgfile");
+}
 }
 
 $defaults = (array)@parse_ini_file("$docroot/plugins/dynamix.docker.manager/default.cfg");
@@ -353,7 +353,7 @@ class DockerTemplates {
 		if (!$imgUrl) $imgUrl = $tmpIconUrl;
 		if (!$imgUrl || trim($imgUrl) == "/plugins/dynamix.docker.manager/images/question.png") return '';
 
-		$iconRAM = sprintf('%s/%s-%s.png', $dockerManPaths['images-ram'], $contName, 'icon');
+				$iconRAM = sprintf('%s/%s-%s.png', $dockerManPaths['images-ram'], $contName, 'icon');
 		$icon    = sprintf('%s/%s-%s.png', $dockerManPaths['images'], $contName, 'icon');
 
 		if (!is_dir(dirname($iconRAM))) mkdir(dirname($iconRAM), 0755, true);
@@ -897,7 +897,7 @@ class DockerClient {
 	}
 
 	public function getDockerContainers() {
-		global $driver;
+		global $driver, $host;
 		// Return cached values
 		if (is_array($this::$containersCache)) return $this::$containersCache;
 		$this::$containersCache = [];
@@ -922,6 +922,7 @@ class DockerClient {
 			$c['Url']         = $info['Config']['Labels']['net.unraid.docker.webui'] ?? false;
 			$c['Shell']         = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
 			$c['Ports']       = [];
+			$c['Networks']	  = [];
 			if ($id) $c['NetworkMode'] = $net.str_replace('/',':',DockerUtil::ctMap($id)?:'/???');
 			if (isset($driver[$c['NetworkMode']])) {
 				if ($driver[$c['NetworkMode']]=='bridge') {
@@ -931,13 +932,23 @@ class DockerClient {
 					$ports = &$info['Config']['ExposedPorts'];
 					$nat = false;
 				}
-				$ip = $ct['NetworkSettings']['Networks'][$c['NetworkMode']]['IPAddress'];
+			} else if (!$id) {
+				$c['NetworkMode'] = DockerUtil::ctMap($c['NetworkMode']);
+				$ports = &$info['Config']['ExposedPorts'];
+				$nat = false;
+				foreach($ct['NetworkSettings']['Networks'] as $netName => $netVals) {
+					$i = $c['NetworkMode']=='host' ? $host : $netVals['IPAddress'];
+					$c['Networks'][$netName] = [ 'IPAddress' => $i ];
+				}
 			}
+			$ip = $c['NetworkMode']=='host' ? $host : $ct['NetworkSettings']['Networks'][$c['NetworkMode']]['IPAddress'] ?? null;
+			$c['Networks'][$c['NetworkMode']] = [ 'IPAddress' => $ip ];
 			$ports = (isset($ports) && is_array($ports)) ? $ports : [];
 			foreach ($ports as $port => $value) {
 				[$PrivatePort, $Type] = array_pad(explode('/', $port),2,'');
-				$c['Ports'][] = ['IP' => $ip, 'PrivatePort' => $PrivatePort, 'PublicPort' => $nat ? $value[0]['HostPort'] : $PrivatePort, 'NAT' => $nat, 'Type' => $Type];
+				$c['Ports'][$PrivatePort] = ['IP' => $ip, 'PrivatePort' => $PrivatePort, 'PublicPort' => $nat ? $value[0]['HostPort'] : $PrivatePort, 'NAT' => $nat, 'Type' => $Type];
 			}
+			ksort($c['Ports']);
 			$this::$containersCache[] = $c;
 		}
 		array_multisort(array_column($this::$containersCache,'Name'), SORT_NATURAL|SORT_FLAG_CASE, $this::$containersCache);
@@ -1016,11 +1027,11 @@ class DockerUtil {
 				// Unprocessable input
 				$strRepo = $image;
 			}
-		}
+}
 
 		// Add :latest tag to image if it's absent
 		if (empty($strTag)) $strTag = 'latest';
-
+		
 		return array_map('trim', ['strRepo' => $strRepo, 'strTag' => $strTag]);
 	}
 
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -338,8 +338,12 @@ class DockerTemplates {
 				$tmp['updated'] = var_export($DockerUpdate->getUpdateStatus($image),true);
 			}
 			if (!$com) $tmp['updated'] = 'undef';
-			if (empty($tmp['template']) || $reload) $tmp['template'] = $this->getUserTemplate($name);
-			if ($reload) $DockerUpdate->updateUserTemplate($name);
+			if ($ct['Manager'] !== 'dockerman')
+				$tmp['template'] = null;
+			else if (empty($tmp['template']) || $reload) {
+				$tmp['template'] = $this->getUserTemplate($name);
+				if ($reload) $DockerUpdate->updateUserTemplate($name);
+			}
 			//$this->debug("\n$name");
 			//foreach ($tmp as $c => $d) $this->debug(sprintf('   %-10s: %s', $c, $d));
 		}
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -924,7 +924,8 @@ class DockerClient {
 			$c['BaseImage']   = $ct['Labels']['BASEIMAGE'] ?? false;
 			$c['Icon']        = $info['Config']['Labels']['net.unraid.docker.icon'] ?? false;
 			$c['Url']         = $info['Config']['Labels']['net.unraid.docker.webui'] ?? false;
-			$c['Shell']       = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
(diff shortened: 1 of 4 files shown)


mtongnz:
Addresses issue #1569

bergware:
Do you have a screenshot when multiple networks are displayed?
(wondering how that looks like)

mtongnz:
Capture

bergware:
thanks, looks good.

kangaroo72:
is it possible to test the fix before the release? I'm waiting for this very long...
If I just have to swap a file, that's great.
Thx

mtongnz:
> is it possible to test the fix before the release? I'm waiting for this very long...
> If I just have to swap a file, that's great.
> Thx
> 

Yes. Edit the webui files on your server as per the commits above.

Easiest way to do this is to use vscode with the sftp plugin as per the unraid GUI Git recommendations. Or you can SSH/terminal into your unraid and use nano - but this is more prone to typos.

I'm away from my PC this week so can't give much more detail. Let me know how you get on.

mtongnz:
> I fixed the problem by swapping the DockerClient.php. Thanks a lot for this.
> Is it also mandatory to swap the DockerContainers.php?
> Thx

Yes. You need to swap both

FunkeCoder23:
I'm currently testing and seems to be working great.

emhttp/plugins/dynamix.docker.manager/include/DockerContainers.php
```diff
@@ -87,11 +87,13 @@ function my_lang_log($text) {
...
+  $networks = [];
+  foreach($ct['Networks'] as $netName => $netVals) {
+    $networks[] = "<span>{$netName}</span><span>{$netVals['IPAddress']}</span>";
+  }
   $ports = [];
   foreach ($ct['Ports'] as $port) {
-    $intern = $running ? ($ct['NetworkMode']=='host' ? $host : _var($port,'IP')) : $null;
-    $extern = $running ? (_var($port,'NAT') ? $host : $intern) : $null;
-    $ports[] = sprintf('%s:%s/%s<i class="fa fa-arrows-h" style="margin:0 6px"></i>%s:%s', $intern, _var($port,'PrivatePort'), strtoupper(_var($port,'Type')), $extern, _var($port,'PublicPort'));
+      $ports[] = sprintf('%s:%s<i class="fa fa-arrows-h" style="margin:0 6px"></i>%s', _var($port,'PrivatePort'), strtoupper(_var($port,'Type')), _var($port,'PublicPort'));
```
FunkeCoder23, line 96: This along with the other comment would only show arrows if a port was actually mapped.
```suggestion
      $arrow_style = _var($port,'PublicPort') ? "\"fa fa-arrows-h\"" : "";
      $ports[] = sprintf('%s:%s<i class=%s style="margin:0 6px"></i>%s', _var($port,'PrivatePort'), strtoupper(_var($port,'Type')),$arrow_style , _var($port,'PublicPort'));
 
```

emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -931,13 +936,23 @@ public function getDockerContainers() {
...
+					$c['Networks'][$netName] = [ 'IPAddress' => $i ];
+				}
 			}
+			$ip = $c['NetworkMode']=='host' ? $host : $ct['NetworkSettings']['Networks'][$c['NetworkMode']]['IPAddress'] ?? null;
+			$c['Networks'][$c['NetworkMode']] = [ 'IPAddress' => $ip ];
 			$ports = (isset($ports) && is_array($ports)) ? $ports : [];
 			foreach ($ports as $port => $value) {
 				[$PrivatePort, $Type] = array_pad(explode('/', $port),2,'');
-				$c['Ports'][] = ['IP' => $ip, 'PrivatePort' => $PrivatePort, 'PublicPort' => $nat ? $value[0]['HostPort'] : $PrivatePort, 'NAT' => $nat, 'Type' => $Type];
+				$c['Ports'][$PrivatePort] = ['IP' => $ip, 'PrivatePort' => $PrivatePort, 'PublicPort' => $nat ? $value[0]['HostPort'] : $PrivatePort, 'NAT' => $nat, 'Type' => $Type];
```
FunkeCoder23, line 953:
```suggestion
				$c['Ports'][$PrivatePort] = ['IP' => $ip, 'PrivatePort' => $PrivatePort, 'PublicPort' => $nat ? $value[0]['HostPort'] : null, 'NAT' => $nat, 'Type' => $Type];
```

FunkeCoder23:
This also removes the ability to edit existing dockers. This seems like a big issue

emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -338,8 +338,12 @@ public function getAllInfo($reload=false,$com=true,$communityApplications=false)
 				$tmp['updated'] = var_export($DockerUpdate->getUpdateStatus($image),true);
 			}
 			if (!$com) $tmp['updated'] = 'undef';
-			if (empty($tmp['template']) || $reload) $tmp['template'] = $this->getUserTemplate($name);
-			if ($reload) $DockerUpdate->updateUserTemplate($name);
+			if ($ct['Manager'] !== 'dockerman')
```
FunkeCoder23, line 341: This never gets set, causing the edit issue I noticed

emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
```diff
@@ -920,8 +924,9 @@ public function getDockerContainers() {
 			$c['BaseImage']   = $ct['Labels']['BASEIMAGE'] ?? false;
 			$c['Icon']        = $info['Config']['Labels']['net.unraid.docker.icon'] ?? false;
 			$c['Url']         = $info['Config']['Labels']['net.unraid.docker.webui'] ?? false;
-			$c['Shell']         = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
+			$c['Shell']       = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
```
FunkeCoder23, line 927:
```suggestion
			$c['Shell']       = $info['Config']['Labels']['net.unraid.docker.shell'] ?? false;
			$c['Manager']     = $info['Config']['Labels']['net.unraid.docker.managed'] ?? false;

```

emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
mtongnz: This is set on line 922 by getDockerContainers()

emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
mtongnz: I already had this but doesn't seem to have published correctly.

mtongnz:
Implemented and tested changes.  Cheers

FunkeCoder23:
I've been using with my tweaks for a while now, working great

FunkeCoder23 (approved):

mtongnz:
Just noticed that non-bridge networks didn't show external ports.  Fixed in latest commit

mtongnz:
@FunkeCoder23 when you get a chance, could you review the last commit I made to this.  I found the NAT detection logic was flawed (based on the old code) and so it wasn't showing external port mappings in many cases.
Cheers

FunkeCoder23:
> @FunkeCoder23 when you get a chance, could you review the last commit I made to this. I found the NAT detection logic was flawed (based on the old code) and so it wasn't showing external port mappings in many cases. Cheers

lgtm

phillipjf:
> Capture

Are each of these deployed with docker-compose? Looking at the change, this PR picks up the network list only for networks in the `HostConfig` but not for containers attached to a network with Post args. I can see my containers list their "primary" network (configured in the template) but the `NetworkSettings` shows the additional (in this case, `traefik`) network:

```json
{
    "Id": "1234...",
    ...
    "HostConfig": { "NetworkMode": "lsio_network" },
    "NetworkSettings": {
        "Networks": {
            "lsio_network": {...},
            "traefik": {...},
        }
    },
    ...
}
```
image

mtongnz:
> > @FunkeCoder23 when you get a chance, could you review the last commit I made to this. I found the NAT detection logic was flawed (based on the old code) and so it wasn't showing external port mappings in many cases. Cheers
> 
> lgtm, I'd just fix up the formatting differences

Sorry, stepped away from this and missed your comment.

What formatting differences are you meaning?
All I can see is DockerClient.php ln 51-56 where spaces changed to tabs to match the rest of the file.  The rest appears correctly formatted.

FunkeCoder23:
> > > @FunkeCoder23 when you get a chance, could you review the last commit I made to this. I found the NAT detection logic was flawed (based on the old code) and so it wasn't showing external port mappings in many cases. Cheers
> > 
> > lgtm, I'd just fix up the formatting differences
> 
> Sorry, stepped away from this and missed your comment.
> 
> What formatting differences are you meaning?
> All I can see is DockerClient.php ln 51-56 where spaces changed to tabs to match the rest of the file.  The rest appears correctly formatted.

Ah, I thought the whitespace was spaces for the original file. If it's tabs, then the formatting is good.

ich777:
@mtongnz after thinking about that PR a bit, this will most likely confuse new/inexperienced users and they would possible think that they have to connect to IP address 172.18.0.14 instead of 10.0.0.1 (HostIP) in my case because the Host IP is now missing when the container is on a bridge or custom bridge network.

grafik

Where it was obvious before:
grafik

Could you maybe display the Host IP too so that both are displayed for a bridge or custom bridge network?
What are your thoughts on that?

mtongnz:
@ich777 thanks heaps for the suggestions.  I've taken them and made a few tweaks in my latest commit:
 - Separated network & IP into separate columns
 - Removed the "app to host" from Port Mappings heading (it's a little misleading now)
 - Added "advanced" class to non-exposed ports so they're hidden in basic view.
 - Added "(internal)" text to non-exposed ports when they're shown in advanced view.  This one I'm not too sure about as they're already hidden in basic mode.

I'm interested to hear your thoughts.

Basic view:
image

Advanced view:
image

mrsdoubtfire613:
@mtongnz phil1c from the unraid forums here.  I implemented the changes to DockerContainers.page, DockerContainers.php, and DockerClient.php manually and have a few issues and a couple suggestions.  I'll start with issues:
1. My dockers on a macvlan subinterface (bond0.43) correctly show the assigned external IP address and port but the ports have the "(internal)" flag.
image
2. For the container referenced in Issue 1 above (on macvlan network), the WebUI link that is generated is an incorrect combination of the unraids primary IP and the external container port.  ex: my server IP is 10.10.0.20, a container on bond0.43 has the external IP of 10.43.64.32 and is accessed at port 8080.  The webUI link *should* be "10.43.64.32:8080" but instead it's "10.10.0.20:8080".
3. I spun up the same stack I have been discussing in the unraid forums (single container, compose-created bridge network NOT bond0.43 as per last discussions) and I see the network and IPs as you've detailed above.  However, the webUI link is taking the container port instead of the host-assigned port for the url.  ex: In the screenshot, the container port for web access is 6080 which I've mapped to 6090 on the host, so the webUI should be "10.10.0.20:6090" but the webUI link is creating "10.10.0.20:6080".
image
NOTE: This also happens to any docker containers I've created using the unraid GUI where I select the network "bridge".  If I select "host" then there isn't a separate internal and host IP and the webUI URL created takes me to the proper url.
image
4. If I modify that stack from Issue 3 to use my bond0.43 network, the "Port Mappings" column still shows the theoretical container and host ports as though it was a bridge network AND the webUI url that is created is still the "10.10.0.20:6080" url; now, that is almost correct as the port should be 6080 but the URL should be the new bond0.43 IP of 10.43.64.33.  To cover my own butt, yes: I destroyed the docker.json file as described in my post on the unraid forums to ensure a new URL is generated and I confirmed as such by making the webui be "google.com" to ensure the changes were taking place.
image


Additionally, I would like to make a couple suggestions, I think echoing similar to what @ich777 was suggesting earlier:
1. IMO, it would be easiest to understand the network layout - especially for new users - if you could display the network name, the container IP, the container port, the external or host IP, and the external or host port.  I haven't the faintest idea where to begin change the unraid GUI code to show this, so hopefully the following table created in word will help explain what I mean.  In this example, "10.10.0.20" is the unraid host IP address.
image

2. To expand on Suggestion 1: I'm explicitly repeating IPs with port assignments so that in instances where a container is on multiple networks you can easily identify which external IP & port corresponds to which internal IP & port.  In your screenshots above, it's not clear to me for adminer which container IP "8080:TCP (internal)" corresponds to (one or the other or both?).


Let me know what other troubleshooting is needed and if you want more explanation on anything.

ich777:
> Added "(internal)" text to non-exposed ports when they're shown in advanced view. This one I'm not too sure about as they're already hidden in basic mode.

I agree with @mrsdoubtfire613 on that, it would be best to lay it out like in his table view since it is way easier for new users to understand how the container network is functioning.

What about renaming the columns from `Internal IP & Port` to `Docker IP & Port` and `External IP & Port` to `Host IP & Port` with that it would be more obvious but I'm also perfectly fine like it is in the example from @mrsdoubtfire613 

The main concern that I have is that the Host IP is nowhere to be found currently and that is really confusing (except of course your internal network is the 172.19.0.x which would be very confusing to me).

mrsdoubtfire613:
> What about renaming the columns from `Internal IP & Port` to `Docker IP & Port` and `External IP & Port` to `Host IP & Port` with that it would be more obvious but I'm also perfectly fine like it is in the example from @mrsdoubtfire613

I'm only hesitant labeling the "external" IP column as "host" because host is a network type and implies the server IP and may cause confusion when a container can be accessed from an IP that isn't the host IP (ie: maclvan or ipvlan).  `Container IP & Port` and `External IP & Port` could be good as well.

In these cases, I'd even suggest a change to my table where instead of the LubeLogger container having "N/A" in the "internal" column, it would have the "external" IP repeated because that's both the internal container IP *and* the externally-accessible IP.  For networks that don't have external access like my Adminer backend network example, it would still have "N/A" in the external column cell.

ich777:
@mtongnz what do you think about that change?

I would really like to have the IP from the host (if host, bridge or a custom bridge network is set) or the custom IP (if the bridge itself is set) visible on the "external" side so that it is clear for users which IP they can connect to.
I also really like the idea from @mrsdoubtfire613 that when a IP/Port mapping is empty that N/A is displayed <- this is a long standing visual issue in the WebUI for some users when they use a custom IP on the bridge itself and the maintainer has no exposed ports in the Dockerfile.

mtongnz:
Try this...
image

mrsdoubtfire613:
I recreated the same 3 test containers of nginx that you have to hopefully simplify troubleshooting - except my "nginx_test" is attached to the macvlan bond0.43 as I don't have a br0 or bond0 to use.  I also added in a docker compose stack attached to two networks to see how that is handled.
image

1. The external ports on my bond0.43 network show nothing (completely blank) and the WebUI URL still uses the servers IP, not the assigned IP.
2. The docker compose stack (pluginMakemkvTest) shows both networks but doesn't show any external IP:Ports.
3. The docker compose stack (pluginMakemkvTest) webui URL is still taking the server IP, not the bond0.43 IP.

Given that I can't see how the UI will look with a container on multiple networks with external IP:PORT, I'll wait to comment on the need to repeat the network and container port to match each external until I can see how it looks

**Not container specific issue:**
I'm having a strange issue where the "basic view/advanced view" toggle isn't setting properly or fully(?).  When I open the docker tab of the UI, the toggle shows "basic view" but the actual UI is advanced view (I rearranged some containers to shrink the screenshots a bit):
image
If I then toggle it to "advanced view" then it switches to basic view *but* the "force update" link appears on the FolderView folders:
image

**General questions about testing these changes:**
In order to implement the changes to the UI that you've made here, I've been manually copying the complete raw file code of DockerContainers.page, /include/DockerClient.php, and /include/DockerCotaniners.php from this PR into the respective files via the terminal.  Every other webUI related file is whatever is included in unraid 6.12.11.
1. Are there other files I should be updating?
2. Is there a better method to implement and test these changes?
3. My simplistic understanding of how unraid works is that the files I'm manually editing to test the changes are not stored on the boot drive, only in memory, so these changes would not persist across a reboot of the server.  Is that correct?

mtongnz:
All the containers I've tried seem to work. The 3 nginx examples are using dockerman but I've also got a number which are created using compose. 

@mrsdoubtfire613 do you have a compose file with a few of your containers so I can try to replicate? All my containers on custom networks ATM to display correctly (I think they're ipvlan though). Please ensure there's no sensitive info in it. Also details of any external networks there containers need.

In regards to updating the files, I use VSCode Server running in a docker work the SFTP plugin. Setup of SFTP is a bit confusing but once you figure out our, it's awesome. It will automatically sync files when you save to a remote system (unRAID in this case).
I then use git to manage versioning. This let's me switch between different branches in VSCode easily and then I sync it across work SFTP. Makes it very quick to A/B any changes. 
Simply copying files works however. They get cleared on reset as you suspect.

mrsdoubtfire613:
@mtongnz Below is the complete contents of the compose file of the pluginMakemkvTest container from my screenshot.  The 3 nginx containers were created using the unraid UI so nothing special about them.  I just pulled from Community Apps and only changed the network mode and ports where appropriate.

`services:
  makemkv:
    container_name: pluginMakemkvTest
    networks:
      - bond043
      - default
    ports:
      - 5910:5900
      - 6090:6080
    privileged: true
    environment:
      - TZ=America/New_York
      - WEBPAGE_TITLE=MakeMKVplugin
      - VNC_PASSWORD=
      - ENABLE_STARTUP_SCRIPTS=yes
      - UMASK=000
      - PUID=99
      - PGID=100
    volumes:
      - /mnt/user:/media:rw
      - /mnt/cache-unprotected/Downloads/:/downloads:rw
      - /mnt/cache-protected-refresh/appdata/dockgeMakemkvTest:/config:rw
    image: binhex/arch-makemkv
networks:
  bond043:
    name: bond0.43
    external: true`

Below is the output of 'docker network inspect bond0.43'.  I removed references to my other containers just to shorten the output a bit.

`[
    {
        "Name": "bond0.43",
        "Id": "48d92dd914b3319138296b5016da749e795767b808632b2a87c37eb851444d9c",
        "Created": "2024-07-17T08:00:11.865073653-04:00",
        "Scope": "local",
        "Driver": "macvlan",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "10.43.0.0/16",
                    "IPRange": "10.43.64.0/18",
                    "Gateway": "10.43.0.1",
                    "AuxiliaryAddresses": {
                        "server": "10.43.0.18"
                    }
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {
            ...
            "7ea11bb0e9045ba7a4bc3a10895e418694f6aa2272dbd5fb9dd5d0d9c3503d88": {
                "Name": "nginx-MACVLAN",
                "EndpointID": "293c8b744f8e42ba189ca0f1cb2729fc3e27e234889dc9645a32e24b29f8e24c",
                "MacAddress": "02:42:0a:2b:40:21",
                "IPv4Address": "10.43.64.33/16",
                "IPv6Address": ""
            },
            ...
        },
        "Options": {
            "parent": "vhost0.43"
        },
        "Labels": {}
    }
]`

Thanks for explaining your workflow.  I figured out how to clone the webui master repo locally and pull this PR and I have Code Server running as well so things should be a bit simpler for me now.

I'll try and run in safe mode tomorrow to disable plugins and see if that changes anything.  If there is anything else you want me to try, just let me know.

ich777:
@mtongnz I've made a PR on your repo, please merge if you like it, everything should now work:
grafik

@mrsdoubtfire613 you can see my PR [here](https://github.com/mtongnz/unRaid-webgui/pull/1).

hrv231:
> This PR aims to improve how networks are displayed for docker containers.
> 
>     * fix where network ID instead of a name is displayed (generally for docker compose created containers).
> 
>     * display multiple networks and the associated IP for each

This is really cool, thank you for doing this!!

mtongnz:
Improvements from @ich777 merged.

This PR now looks ready for merge into an upcoming release.

ich777:
@mtongnz I've issued another PR to your PR.

@hrv231, @mrsdoubtfire613, @anym001, @soonic6 what about changing `External IP:Port` to `Host IP:PORT` for consistency since in the Docker template you have the option to Specify a `Container` and `Host` port.
The argument that a Docker network with the name `Host` exists is not enough in my opinion.
It would be way better and easier for the average user to understand and as mentioned above in terms of consistency to name it `Host IP:PORT`

mrsdoubtfire613:
@ich777 That's a fair point.  It even lines up with general Docker terminology beyond templates in Unraid.  And "external" would only not be the host IP:PORT in instances like mine where a user is using macvlan and/or ipvlan network drivers which is likely not what a new or average user is doing.  I'll rescind my recommendation and I agree "External" should be "Host".

hrv231:
> @mtongnz I've issued another PR to your PR.
> 
> @hrv231, @mrsdoubtfire613, @anym001, @soonic6 what about changing `External IP:Port` to `Host IP:PORT` for consistency since in the Docker template you have the option to specify a `Container` and `Host` port. The argument that a Docker network with the name `Host` exists is not enough in my opinion. It would be way better and easier for the average user to understand also as mentioned above in terms of consistency to name it `Host IP:PORT`

I think the External IP explains it better.
The average user doesn't understand how docker networks work behind the scenes.
The average user doesn’t create new networks with "docker network create" specifying names and if it is an ipvlan or a macvlan.

But what about for the users who do create their separate networks?
Let's say I create a network with macvlan, I would prefer in that case to see an "External IP" than the "Host IP".
The Host IP in this case will always be the unraid server IP; but for a macvlan network where the containers have their own IPs, then that's different.

I'm fine with whatever you guys decide, I know this is going to be better than the Network ID we are getting in the GUI anyways :-)

ljm42:
Thanks a ton to everyone who worked on this PR! I've tested with the latest from ich777 and the changes are excellent, a very nice improvement over the current page. I expect this will make it into the next 7.0.0 prerelease.

My only concern is the column heading "External IP", we can't merge with that. It sets the wrong expectation for the page - half of our users would be scared that we are now exposing their containers externally to the Internet, and the other half would be mad when they find out that isn't what external means here :) I'd be happy with "Host IP" but am open to something else

anym001:
> @mtongnz I've issued another PR to your PR.
> 
> 
> 
> @hrv231, @mrsdoubtfire613, @anym001, @soonic6 what about changing `External IP:Port` to `Host IP:PORT` for consistency since in the Docker template you have the option to specify a `Container` and `Host` port.
> 
> The argument that a Docker network with the name `Host` exists is not enough in my opinion.
> 
> It would be way better and easier for the average user to understand also as mentioned above in terms of consistency to name it `Host IP:PORT`

I would also find it better to rename it to Host IP:PORT. 
This would give you a consistent wording compared to the docker templates.

soonic6:
maybe a bad idea, becauso of space and wording:

`IP:PORT in Container`
and
`IP:PORT on Host`

this could be more understandable for average users, also the wording with `container` and `host` is consistent for adavanced users?

anym001:
> maybe a bad idea, becauso of space and wording:
> 
> 
> 
> `IP:PORT in Container`
> 
> and
> 
> `IP:PORT on Host`
> 
> 
> 
> this could be more understandable for average users, also the wording with `container` and `host` is consistent for adavanced users?

I find this proposal the best so far.

mtongnz:
> > maybe a bad idea, becauso of space and wording:
> > 
> > 
> > 
> > `IP:PORT in Container`
> > 
> > and
> > 
> > `IP:PORT on Host`
> > 
> > 
> > 
> > this could be more understandable for average users, also the wording with `container` and `host` is consistent for adavanced users?
> 
> I find this proposal is the best so far. 

Maybe LAN IP:PORT

ich777:
I've now made a few screenshots how that all would look like to better visualize it:

Host:
grafik

LAN:
grafik

IP on Host:
grafik


TBH I kind of like the `LAN IP:Port` the most so far since it seems easy to understand even for new users.

mtongnz:
Agreed. I like how compact it is while relaying clearly what the column represents.
I think host will have similar objections to external and possibly cause some confusion.

hrv231:
> TBH I kind of like the `LAN IP:Port` the most so far since it seems easy to understand even for new users.

In my case, I like the first one, Host IP:Port

Where are the containers and VMs running? in the Host ( the unraid server )

hrv231:
Can you show us how is it going to look like if you set a LAN IP to a container in a macvlan network?
In that case then LAN IP:Port will make more sense.

ich777:
> Can you show us how is it going to look like if you set a LAN IP to a container in a macvlan network?

It will look exactly the same as [here](https://github.com/unraid/webgui/pull/1616#issuecomment-2255942226) except for the hedings.

hrv231:
> It will look exactly the same as [here](https://github.com/unraid/webgui/pull/1616#issuecomment-2255942226) except for the hedings.

I see; then yes, the `LAN IP:Port` makes more sense for all scenarios

mtongnz:
Changes from @ich777 tested & merged in.  I think it is very clear.

limetech:
great work!

mtongnz:
Just noticed that networks connected to docker via docker network connect don't show.
Addressed in #1828

PR #1617, opened 2024-02-10
Fix php warning when saving text to a file.

dlandon:
A php warning occurs when a text area is to be saved to a file and it contains a reserved php character.  Update.php tries to parse the file and extract a specific section even if the file is only to be saved.  It is also trying to extract a section when the $section is not defined i.e. $section is set to false.

### DIFF
--- a/emhttp/update.php
+++ b/emhttp/update.php
@@ -60,8 +60,8 @@ if (isset($_POST['#file'])) {
   }
   if ($save) {
     $text = "";
+    $keys = is_file($file) ? (parse_ini_file($file, $section) ?: []) : [];
     if ($section) {
-      $keys = is_file($file) ? (parse_ini_file($file, $section) ?: []) : [];
       foreach ($_POST as $key => $value) if ($key[0]!='#') $keys[$section][$key] = $default[$section][$key] ?? $value;
       foreach ($keys as $section => $block) {
         $text .= "[$section]\n";
--- a/emhttp/update.php
+++ b/emhttp/update.php
@@ -49,7 +49,6 @@ if (isset($_POST['#file'])) {
   $cleanup = isset($_POST['#cleanup']);
   $default = ($file && isset($_POST['#default'])) ? @parse_ini_file("$docroot/plugins/".basename(dirname($file))."/default.cfg", $section) : [];
 
-  $keys = is_file($file) ? (parse_ini_file($file, $section) ?: []) : [];
   // the 'save' switch can be reset by the include file to disallow settings saving
   $save = true;
   if (isset($_POST['#include'])) {
@@ -62,6 +61,7 @@ if (isset($_POST['#file'])) {
   if ($save) {
     $text = "";
     if ($section) {
+      $keys = is_file($file) ? (parse_ini_file($file, $section) ?: []) : [];
       foreach ($_POST as $key => $value) if ($key[0]!='#') $keys[$section][$key] = $default[$section][$key] ?? $value;
       foreach ($keys as $section => $block) {
         $text .= "[$section]\n";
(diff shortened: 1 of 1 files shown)


limetech:
I don't think this is right because if you are changing one variable of a file without sections only that variable will be re-written to the file and all existing variables will be discarded.

dlandon:
All this line does is to get the values from the section to be modified:
$keys = is_file($file) ? (parse_ini_file($file, $section) ?: []) : [];

The $keys value is not used until this line:
foreach ($_POST as $key => $value) if ($key[0]!='#') $keys[$section][$key] = $default[$section][$key] ?? $value;
which is after the if ($section) check.

Leaving it where it was caused the php error because it tries to parse text with a reserved character when trying to parse a non-existant section that is not processed on a text save to a file.  Moving the line to where it makes more sense prevents the php error and no functionality has been changed.

dlandon:
I do need to make a change.

dlandon:
The line $keys line is in the right place now.

limetech:
In the case of $save===true there is no functional difference that this change makes - sure this is what you want?

dlandon:
This PR is not for any functional change.  It just prevents a php warning when saving a text area to a file that has a reserved php character.

limetech:
What is the php warning?  There is a functional change: if the '#include' POST var is present, theoretically the file it includes could modify the $keys array before saved back to config file.  I didn't look at all the cases where '#include' is used but this could introduce a pretty obscure bug if this was happening.

dlandon:
UD script files are bash files that are created and edited that run when UD devices have udev events.  This is a screen shot of the settings page:
Screenshot 2024-02-14 105059

When 'Apply' is clicked, this php warning apears: [14-Feb-2024 10:49:50 America/Chicago] PHP Warning:  syntax error, unexpected '!' in /boot/config/plugins/unassigned.devices/Music_Backup.sh on line 1
 in /usr/local/emhttp/update.php on line 52

I've not seen any issues when using the modified file.

limetech:
What does the code for that form look like?

dlandon:
The problem is the $file being passed to update.php in my case is a file designation for an existing file that shows in the text area.  It's not a config file.  This is the form designation:
<form markdown="1" name="edit_script" method="POST" action="/update.php" target="progressFrame">
<input type="hidden" name="#include" value="/webGui/include/update.file.php">

This is the line defining #file which is a file name for the file read into the text area:
: <input type="text" title='_(Enter Device Script File Name)_.' name="#file" style="width:46%;" <?=(!$script_disabled ? 'onclick="openFileBrowserDevice(this, \'/boot/config/plugins/' . $plugin . '\', \'sh\');"' : 'readonly');?> autocomplete="off" value="<?=$file;?>" placeholder="_(Click here to select a Script File or 'Default' button for a Default Script)_">

Because the $file is not a config file in this case, it cannot be parsed with reserved php characters in the file.  The idea as I see it is the #include will designate that the $file be saved with update.file.php.  It shouldn't be parsed in this case.

Your concern that the $keys are parsed before the file save would not take into account changes is not a concern.  The #include indicates the file should be saved, and as I see it the $file in this case is not a config file and wouldn't be parsed.  Because the $file is used for saving a text area to a file, and could be a config file kind of makes them mutually exclusive when #include is defined.  You'll see "$save = false" after the #include file is executed, this then negates the code that uses $keys, but as I see it, this is where the $keys should be used when #include is not defined beause the $file is a config file.

limetech:
The UD code is still leaving the $save flag set true right?  Then how does moving the parse_ini_file() to line 64 avoid the php warning?

dlandon:
If you are talking about the $save in update.php, that's a local variable.  I don't see how UD has any control over that.

bergware:
This PR at present is a no go.
The "$keys" assignment line has to stay before the "#include" execution, because a include script needs to be able to alter/update the $keys variable before further processing.

dlandon:
> This PR at present is a no go.

I believe that was decided in the community council today.

dlandon:
My latest revision adds a hidden input named "#raw_file" that when set to "true" will block the ini parsing.  This is completely backwards compatible and should not affect any existing configurations.  If the input does not exist, update.php functions exactly as it does now.

bergware:
Can you modify the check, which verifies the existence of the file regardless of format:
`$keys = (is_file($file) && !$raw_file) ? (parse_ini_file($file, $section) ?: []) : [];
`

PR #1640, opened 2024-02-28
Fix `getDockerJSON`

Commifreak:
If docker comms fail, return (empty) array (as expected by everyone) instead of null. Fixes `foreach() argument must be of type array|object, null given`


limetech:
thank you!

PR #1667, opened 2024-03-14
Add additional USB device passthrough smartmontools options to webgui

jski:
There are some new devices since this page was last updated in the webgui, so I've gotten started with adding the ones that are available in version 7.3 (included with unraid 6.12, which I am using). I can't test all the flags as I don't have all of these devices available, but I can confirm that it does indeed work on my device using the flags added for the one device type I needed.

Hope this gets you started with getting this added into future versions!

### DIFF
--- a/emhttp/plugins/dynamix/DeviceInfo.page
+++ b/emhttp/plugins/dynamix/DeviceInfo.page
@@ -313,8 +313,8 @@ function setGlue(form,reset) {
    {glue:'/',more:3,dev:1,type:'',min1:1,max1:4,min2:1,max2:128,min3:1,max3:4},            // highpoint
    {glue:'' ,more:1,dev:1,type:'',min1:0,max1:15},                                         // hp cciss
    {glue:'' ,more:0,dev:0,type:''},                                                        // marvell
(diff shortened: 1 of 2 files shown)


emhttp/plugins/dynamix/DeviceInfo.page
```diff
@@ -314,6 +314,13 @@ function setGlue(form,reset) {
    {glue:'' ,more:1,dev:1,type:'',min1:0,max1:15},                                         // hp cciss
    {glue:'' ,more:0,dev:0,type:''},                                                        // marvell
    {glue:',',more:1,dev:1,type:'',min1:0,max1:127}                                         // megaraid
```
ljm42, line 316: This needs a comma after `127}`

emhttp/plugins/dynamix/DeviceInfo.page
```diff
@@ -314,6 +314,13 @@ function setGlue(form,reset) {
...
+   {glue:',',more:1,dev:0,type:'input'},min1:'',max1:'0xffffffff'},                        // cypress
```
ljm42, line 317: Looks like there is an extra `}` after `'input'`

jski (commented):
Fixed

limetech:
thank you!

PR #1695, opened 2024-04-04
Add password encryption and rewrite set_proxy script in php

dlandon:
Add password encryption and rewrite set_proxy script in php.


limetech:
nice work dan!

dlandon:
Thank you.

PR #1706, opened 2024-04-21
New vm interface template with inline xml view

SimonFair:
Initial implementation. Slider added to show inline.

Examples
image
image

image
image
image


limetech:
very cool!

PR #1716, opened 2024-05-06
Various VM Manager Fixes

SimonFair:
Fix VM memory size values on VM Usage
Fix issue with auto showing for QCOW2 files not in a VM path, should be manual
add evdev passthru


limetech:
very nice, thanks!

PR #1720, opened 2024-05-08
Disable "Add Share" button when there are no mountable array disks; disable Reiserfs selections; add multi line NFS Rule.

dlandon:
Adding a share when there are no mountable disks in the array does not make sense.  The message on the GUI will show why a share cannot be added.

Disable Reiserfs selections so user cannot format array disks with Reiserfs.  Just added "disabled" to the drop down selections.

Add multi line NFS Rule to share NFS Settings.  The 'Read' and 'Write' functons have also been updated so rules can be read from another share and written to another share.  The way this works is the individual lines are put together into a single string with a space delimiter so all the rules end up as one line in the exports file.  The code had to be refactored just so it was readable and could be understood.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,7 +54,6 @@ sftp-config.json
 
 # =========================
 # Exclude these dirs commonly found in /usr/local
-bin/
 games/
 info/
 lib64/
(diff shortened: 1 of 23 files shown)


dlandon:
I see I forgot to apply two small changes to the Docker and VM settings pages PR recently applied and they came into this PR.  Those two changes are very minor.

limetech:
nice work, thx!

PR #1723, opened 2024-05-08
Add lincstation n1 to cases

Squidly271:


emhttp/plugins/dynamix/font-cases.woff
zackspear, line 1: Please move into `emhttp/plugins/dynamix/styles`

emhttp/plugins/dynamix/styles/default-cases.css
```diff
@@ -71,4 +71,5 @@
 .case-vivopc:before{content:'\e925'}
 .case-vm:before{content:'\e920'}
 .case-x3650-M3:before{content:'\e924'}
-.case-zalman-ms800:before{content:'\e947'}
\ No newline at end of file
+.case-zalman-ms800:before{content:'\e947'}
+.case-lincstation-n1:before{content:'\e948';font-weight:400}
```
zackspear, line 75: Is `font-weight: 400` required here? It would be the first case to need this.

zackspear (changes requested):

emhttp/plugins/dynamix/styles/default-cases.css
Squidly271: Yeah, Been awhile since I completely messed up a PR lol

zackspear: Been there plenty of times ;)

PR #1743, opened 2024-05-27
Add in the ability to set a manual UPS capacity override in `dynamix.apcupsd`

othyn:
Add in the ability to set a manual UPS capacity override in `dynamix.apcupsd`, so that Nominal Power can still be estimated based on the load percentage for UPS's that don't report their `NOMPOWER`. Its not fool proof, and only gets you in the ball park range, but its good enough, and tracks with how other platforms and plugins allow you to do this too. 

For example, NUT UPS implements the same behaviour, allowing the user to provide manual overrides in order to allow for Nominal Power and Apparent Power to still be estimated with UPS' that don't natively report those statistics.

For an example of deviance, my Tasmota power plug upstream from the UPS reports around `≈ 110 W` with the estimated UPS calculation resulting in `≈ 120 W`. This due to rounding, load just being a flat percentage, not factoring in the UPS' efficiency rating, etc.

I've added the;
- Relevant settings value capture from the user on the `/Dashboard/Settings/UPSsettings` page, giving the user the ability to define `Override UPS Capacity (Watts)`.
- Relevant stats output to the top of the `/Dashboard/Settings/UPSsettings` page, implementing the new estimation calculation.
- Relevant stats output to the dashboard widget, implementing the new estimation calculation.

All appears to be working as intended on my prod installation `6.12.6`.

UPS settings page:

ups_settings

Dashboard widget:

dashboard_widget

To ensure the user understands that the value is an estimate, I've prefixed the value that is generated by the estimate calculation with the symbol for estimation `≈`, to ensure differentiation against a value read directly from the UPS (if reported).

The override calculation will only be used in the event that all the following criteria are met;
- The override is defined; and
- The override value is greater than zero; and
- There is no `NOMPOWER` value reported by the UPS (the cell is in its default state); and
- The load is defined; and
- The load is greater than zero.

I've ensured that there is appropriately detailed help text on the field, to hopefully guide the user into how to use the functionality. Sticking with industry standard terminology and norms when it comes to calculating the nominal power from the UPS capacity and load percentage.

I've tried to keep the added code to the same style and structure present.

I did notice though a few differences, such as the `master` branch uses an `$echo` named variable in `emhttp/plugins/dynamix/nchan/ups_status` where as on my prod box, `emhttp/plugins/dynamix/nchan/ups_status` uses the same `$status` variable naming as in `emhttp/plugins/dynamix.apcupsd/include/UPSstatus.php`. So I'm not sure if I'm using the correct branch? Some guidance on that would be appreciated, as I'm happy to change the PR target and make the appropriate changes. I did notice there was a `6.12` branch, however that one seemed a way behind `master`.


limetech:
Nice work!  we are reviewing...

limetech:
Very nice!  thank you!

johnnieblack-JorgeB:
@othyn thanks for the PR, it's a great addition for users with a UPS that doesn't report NOMPOWER, I was wondering if you could make a couple of display changes, to keep the display consistent with stock behavior.

The nominal power value should be fixed, and the rated UPS value, so in this case it should display the custom value the user inputs, the load in Watts should display together with the load percentage, both on the UPS settings tab, and the Dashboard, please see below:

image

image

othyn:
@johnnieblack-JorgeB oh super, thank you for sharing that! Agreed, I was trying to follow how it behaved for normal UPS', but didn't have one to test with, so thank you for sharing that. I'll submit a new PR to align how it should behave.

So in your screenshot, the 1400W nominal power is the rated max load (in watts) of the UPS? So essentially, in my example, that would be the manual override 480W value?

PR #1750, opened 2024-06-01
Create WebUI option in VM Template.

SimonFair:
Add a WebUI to VM template.
Parms to be converted at runtime.
[IP] - Guest agent must be install to support this variable. First IP used.
[VMNAME] - Spaces to be replaced with -
[PORT:XX] - XX is port number you want to use.

image

image
image

image
image


limetech:
Another cool feature!

PR #1751, opened 2024-06-03
When no array, hide disk shares

ljm42:


johnnieblack:
Larry, won't this also hide the disk shares for pools?

ljm42:
Pools don't have individual disk shares right? Just the array

SimonFair:
All of my pools show in disk shares.

image

ljm42:
Thanks, if Disk Shares are enabled I do Pools listed. Closing this, not a good idea

PR #1774, opened 2024-07-03
Updates to rc.apcupsd and VM updates

SimonFair:
apcupsd Change restart to run only if already running.
VM Fix for guest not init display when changing from DGPU to virtual.
VM Fix for clone with no disks
VM Add. support to block pull and commit for stopped vms
VM fix to rc.libvirt for multiple nics when bridge mode is changed.


desertwitch:
Thanks Simon, should be sufficient to fix:
https://forums.unraid.net/bug-reports/prereleases/612x700-beta1-log-rotation-starts-user-disabled-apcupsd-service-r3063

PR #1791, opened 2024-07-16
Show Docker RAM usage on Dashboard & minor changes

ich777:
- Add Docker RAM usage to Dashbaord page
- Rename 'Services' to 'System'
- Remove 'usage' suffix from Services naming scheme

After the changes it will look like that:
grafik

and with ZFS like that:
grafik


limetech:
nice! thank you

PR #1793, opened 2024-07-16
Fix UPS settings page when config contains empty values

jbtwo:
## Issue:

When either of the UPS config options `Battery level to initiate shutdown (%)` or `Runtime left to initiate shutdown (minutes)` are set to empty, the UPS details section breaks. This is due to the `getUPSstatus` function passing values directly from the config, which results in an error if either value is empty.

UPSsettings 2024-07-16 19-38-05

## Solution:

The fix sets and passes variables to the `getUPSstatus` function for each of the config options. The variables are either set to the value contained in the config file, or `0` if the config has no value set.

Tested the fix locally and confirmed it functions as expected.

### DIFF
--- a/emhttp/plugins/dynamix.apcupsd/UPSdetails.page
+++ b/emhttp/plugins/dynamix.apcupsd/UPSdetails.page
@@ -14,11 +14,13 @@ Tag="battery-3"
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
  */
+$docroot = $docroot ?? $_SERVER['DOCUMENT_ROOT'] ?: '/usr/local/emhttp';
+require_once "$docroot/webGui/include/Wrappers.php";
 ?>
 <script>
 function getUPSstatus() {
-  var batteryLevel = "<?= !empty($cfg['BATTERYLEVEL']) ? $cfg['BATTERYLEVEL'] : 0 ?>";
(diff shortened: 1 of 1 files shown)


ljm42:
Good catch!

Would you please switch this over to the `_var()` function from Wrappers.php? It saves having to check for empty array values, and allows setting a default:

```php
<?php
$docroot = $docroot ?? $_SERVER['DOCUMENT_ROOT'] ?: '/usr/local/emhttp';
require_once "$docroot/webGui/include/Wrappers.php";
?>

var batteryLevel = "<?= _var($cfg,'BATTERYLEVEL',0) ?>";
var batteryRuntime = "<?= _var($cfg,'MINUTES',0) ?>";
```

jbtwo:
> Good catch!
> 
> Would you please switch this over to the `_var()` function from Wrappers.php? It saves having to check for empty array values, and allows setting a default:
> 
> ```
> <?php
> $docroot = $docroot ?? $_SERVER['DOCUMENT_ROOT'] ?: '/usr/local/emhttp';
> require_once "$docroot/webGui/include/Wrappers.php";
> ?>
> 
> var batteryLevel = "<?= _var($cfg,'BATTERYLEVEL',0) ?>";
> var batteryRuntime = "<?= _var($cfg,'MINUTES',0) ?>";
> ```

Thanks @ljm42, changes made

ljm42 (approved):
Looks great, thanks!

PR #1796, opened 2024-07-18
fix table order breaking when cookie is left malformed

desertwitch:
same as #1472 and #1776 just now for sanitizing the tile ordering cookies as well
if the browser leaves cookies in a malformed state (with null, empty or duplicate md5s) the tile ordering is not remembered
this PR cleans up the cookies (as already done with the hidden & inactive tile cookies) so tile sorting does not fail in such cases
--Rysz


limetech:
thanks!

PR #1808, opened 2024-07-30
Overlay2 GUI support

ich777:
- add Docker storage driver to Docker page
- create helptext for Docker storage driver
- add DOCKER_BACKINGFS to `docker.cfg`
- bugfix for always appending a second `/docker/` when using the Docker directory in the GUI

grafik


ich777:
I've now changed the PR to set `overlay2` as the default for Docker directories

PR #1809, opened 2024-08-01
update.php: release file pointer after usage

desertwitch:
When execution is done (EOF), might be good practice to attempt releasing the now unused file pointer back to the system.
Just checked the rest of the repository and it seems to be done everywhere else like this, was probably just forgotten here.

-- Rysz

### DIFF
--- a/emhttp/update.php
+++ b/emhttp/update.php
@@ -102,6 +102,7 @@ if (isset($_POST['#command'])) {
     while (!feof($proc)) {
       write_log(fgets($proc));
     }
+    @pclose($proc);
   }
 }
 ?>


ljm42:
Thanks @desertwitch, great catch!  And thanks for checking elsewhere for the same issue

Would you be open to backporting this to the 6.12 branch as well?

desertwitch:
> Thanks @desertwitch, great catch!  And thanks for checking elsewhere for the same issue
> 
> 
> 
> Would you be open to backporting this to the 6.12 branch as well?

Will do tonight, thanks for the suggestion!

limetech:
thank you!

PR #1810, opened 2024-08-01
(backport) update.php: release file pointer after usage

desertwitch:
#1809 backported for 6.12

_When execution is done (EOF), might be good practice to attempt releasing the now unused file pointer back to the system.
Just checked the rest of the repository and it seems to be done everywhere else like this, was probably just forgotten here._
_-- Rysz_


desertwitch:
thanks @limetech and @ljm42 --Rysz

PR #1814, opened 2024-08-03
Fixed an issue where routes could not be saved in languages ​​other t…

suzukua:
Fixed an issue where routes could not be saved in languages ​​other than English

https://forums.unraid.net/bug-reports/stable-releases/%E3%80%9061211%E3%80%91%E3%80%90bug%E3%80%91cannot-save-static-route-r3156/


limetech:
Thank you!

PR #1820, opened 2024-08-10
feat(user-edit): update allowed ssh key encryptions

Garbee:
This follows up the changes from #1127 to allow for more modern SSH key types. This changes the regex to the one provided for [combined support for all key types](https://github.com/nemchik/ssh-key-regex?tab=readme-ov-file#support-all-known-key-types-less-secure). If support for restrictions to the more secure key types only is desired, I think that needs a discussion on adding a "Hardened Mode" setting somewhere that may also impact other parts of the system too, instead of just SSH key formats.

This fixes an issue where current modern keys are being rejected by the web interface. Such as fresh keys generated with [Secretive](https://github.com/maxgoedjen/secretive) where users don't have control over which type of encryption is used.


limetech:
thank you!

Issue #1175, opened 2022-10-20
Use of the `document.write()` method is strongly discouraged


zackspear:
Saw this while testing something:
<img width="1669" alt="Screen Shot 2022-10-20 at 16 06 40" src="https://user-images.githubusercontent.com/2531682/197075596-416d3719-3c97-4c31-84e8-98828beabf46.png">

Chrome has been warning about this since 2016. https://developer.chrome.com/blog/removing-document-write/

Shouldn't take too much to refactor an alternative way of showing errors on the page.

More details: 
- https://developer.mozilla.org/en-US/docs/Web/API/Document/write
- https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#document.write()

webgui usage of `document.write()`
- https://github.com/limetech/webgui/blob/4b1e418a1b3a2e739e23bc7aedb704c66cf7a079/plugins/dynamix/include/.set-password.php#L349
- https://github.com/limetech/webgui/blob/4b1e418a1b3a2e739e23bc7aedb704c66cf7a079/plugins/dynamix/include/.login.php#L532
- https://github.com/limetech/webgui/blob/d7982ce3e072d5c86d5f4ce79e5da29e5f7838db/plugins/dynamix.vm.manager/novnc/app/error-handler.js#L58

PR #1821, opened 2024-08-10
fix: replace document write usages

Garbee:
This replaces the two uses of document.write with a more modern solution. First we create the content needed in the body. Then wipe the body content out and finally append the desired content to it.

This does NOT fix the [3rd referenced use](https://github.com/limetech/webgui/blob/d7982ce3e072d5c86d5f4ce79e5da29e5f7838db/plugins/dynamix.vm.manager/novnc/app/error-handler.js#L58) in the app code. Since it is from the VNC module, we'd be diverging from upstream which would make updating the VNC client more difficult in the future unless we rember to always go patch that again. To fix that, we should work upstream to have a solution in their system. There are also a few more findings of `document.write` elsewhere in the app code that is tracked. Same situation as this one for all of them, being upstream code that we shouldn't touch.

Fixes: #1175


zackspear (approved):

limetech:
I'll make that small wording change.

Garbee:
Thanks for doing that update @limetech. I had caffeine withdrawal (kicking the dependency on it) the past few days so wasn't checking things.

PR #1822, opened 2024-08-12
Add ZFS Checking

SimonFair:
This will check VM name does not include characters that are not valid for ZFS. Existing VMs are not modified but will throw error and disable update if invalid characters are found.

image


limetech:
nice! thx

PR #1828, opened 2024-08-15
Network info display improvements (part 2)

mtongnz:
This commit ensures containers with multiple networks are always displayed. Previously, networks connected to containers (docker network connect...) don't show
This is an improvement on #1616

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -944,7 +944,8 @@ class DockerClient {
 				} elseif ($driver[$c['NetworkMode']]=='host') {
 					$c['Ports']['host'] = ['host' => ''];
 				} elseif ($driver[$c['NetworkMode']]=='ipvlan' || $driver[$c['NetworkMode']]=='macvlan') {
-					$c['Ports']['vlan'] = ['vlan' => ''];
+					$i = $ct['NetworkSettings']['Networks'][$c['NetworkMode']]['IPAddress'];
+					$c['Ports']['vlan'] = ["$i" => $i];
 				} else {
 					$ports = &$info['Config']['ExposedPorts'];
 				}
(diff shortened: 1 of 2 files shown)


ich777:
@mtongnz can you please post a screenshot?

mtongnz:
This is a container (heimdall) created using the "add container" via GUI.  I then run this to add a second network:
`docker network connect
  --ip='192.168.9.198' br0.9 heimdall2`

Pre this commit:
image

Post this commit:
image

Note that the first image only shows the original network even though both are present on the container.  I made no edits to the container between images.

mtongnz:
This latest commit looks to have fixed the LAN:IP not showing all "external" IPs
image

ich777:
@mtongnz I try this commit on my test system ASAP and report back if everything is working for all other network configurations.

hrv231:
> This commit ensures containers with multiple networks are always displayed. Previously, networks connected to containers (docker network connect...) don't show This is an improvement on #1616


I came here to say _Thank you!!_ for your work

ich777:
Seems like this change causes issues with listing networks from containers that use other container networks:
grafik

Do [these lines](https://github.com/unraid/webgui/pull/1828/files#diff-ba385bd201ed11eb6f191d7096e95dff8d4aaaf99e137ccaa902238051daf7d6L959-L960) cause issues on your setup, after adding them everything is back to normal over here:
grafik

ich777:
@mtongnz on another note, how does the WebUI button behave if you have assigned multiple networks to what IP does it default or does it try to assign both IP addresses? Would be really interesting how that behaves on your setup since I was only able to test that with one network (I haven't got such a complex setup :D ).

mtongnz:
It seems that "container:" networks don't show up in $ct['NetworkSettings']['Networks'].  I've put the original code back in before the for loop I moved and it is working on my system.  Does it fix it for you?
I've also modified code to better display the "external" LAN IP for ipvlan/macvlan when there's multiple networks

ich777:
> Does it fix it for you?

Yes, the latest commit does fix it on my end.

Did you read my comment above about the WebUI button with multiple assigned networks?

mtongnz:
> Did you read my comment above about the WebUI button with multiple assigned networks?
I missed that. I'll check when I'm back at my machine. Most of my containers have the webui modified as each runs on a subdomain.

ich777:
> I missed that. I'll check when I'm back at my machine. Most of my containers have the webui modified as each runs on a subdomain.

I just want to double check that it get's assigned properly when using the default value since this is pretty important, I assume it tries to assign both IPs how it's currently implemented, maybe we need to strip other IPs and only select one (maybe the first one), if one really needs another IP then it can be changed manually.

mtongnz:
> > I missed that. I'll check when I'm back at my machine. Most of my containers have the webui modified as each runs on a subdomain.
> 
> I just want to double check that it get's assigned properly when using the default value since this is pretty important, I assume it tries to assign both IPs how it's currently implemented, maybe we need to strip other IPs and only select one (maybe the first one), if one really needs another IP then it can be changed manually.

Depending on network type, it either returns the first port set in getDockerContainers() or the first IP from this command in DockerUtil::myIP() ($ipaddress is ipv4 vs v6 & $name is container name):
`docker inspect --format='{{range .NetworkSettings.Networks}}{{.$ipaddr}} {{end}}' $name`

Both seems to return the IP associated to the network set in the container's NetworkMode (ie. the network set in the docker run/create command).  Not sure what is set as primary if using docker compose (I'd guess the order the networks are entered into the compose file).

So these changes don't change how the webUI is generated and should retrun the same values.  As you say, if people are setting up multiple networks on a container, they can override the webui IP using the labels available.

ich777:
@limetech ready to merge.

Issue #1832, opened 2024-08-19
Feature Request: Use user's actual shell in OpenTerminal.php


donbuehl:
### Description
Currently, OpenTerminal.php is using a hardcoded 'bash' shell. I propose to add a function that determines the user's actual shell and uses it when opening the terminal.

### Proposed Solution
1. Add a `getUserShell()` function to OpenTerminal.php that retrieves the user's shell from /etc/passwd. If the shell can't be determined, it should default to 'bash'.

2. Modify the existing exec call to use the result of `getUserShell()`. The original line:
```php
if ($retval != 0) exec("ttyd-exec -i '$sock' bash --login");
```
should be changed to:

```php
if ($retval != 0) exec("ttyd-exec -i '$sock' " . getUserShell() . " --login");
```
Here's the proposed function:

```php
function getUserShell() {
    $shell = 'bash';

    try {
        $username = posix_getpwuid(posix_geteuid())['name'];
        $passwd = file_get_contents('/etc/passwd');
        $lines = explode("\n", $passwd);
        foreach ($lines as $line) {
            if (strpos($line, $username) === 0) {
                $parts = explode(':', $line);
                $fullShellPath = end($parts);
                $shell = basename(trim($fullShellPath));
                break;
            }
        }
    } catch (Exception $e) {
        syslog(LOG_ERR, "Error determining user shell: " . $e->getMessage());
    }

    return $shell;
}
```
### Benefits
- Improved user experience by using the user's preferred shell, e.g. zsh.
- Better compatibility with various user setups.
- Fallback to bash ensures the feature works even if shell detection fails.

PR #1835, opened 2024-08-19
Feature Request: Use user's actual shell in OpenTerminal.php #1832

donbuehl:
Added function getUserShell() to OpenTerminal.php and use it for the execution

### DIFF
--- a/emhttp/plugins/dynamix/include/OpenTerminal.php
+++ b/emhttp/plugins/dynamix/include/OpenTerminal.php
@@ -51,7 +51,9 @@ case 'ttyd':
     // no child processes, restart ttyd to pick up possible font size change
     if ($retval != 0) exec("kill ".$ttyd_pid[0]);
   }
-  if ($retval != 0) exec("ttyd-exec -i '$sock' " . basename(posix_getpwuid(0)['shell']) . " --login");
+  $userShell = basename(posix_getpwuid(0)['shell']);
+  $shell = in_array($userShell, ['bash', 'sh', 'zsh', 'fish', 'ksh', 'tcsh', 'dash']) ? $userShell : 'bash';
+  if ($retval != 0) exec("ttyd-exec -i '$sock' $shell --login");
   break;
 case 'syslog':
   // read syslog file
--- a/emhttp/plugins/dynamix/include/OpenTerminal.php
+++ b/emhttp/plugins/dynamix/include/OpenTerminal.php
@@ -73,7 +73,7 @@ case 'ttyd':
     // no child processes, restart ttyd to pick up possible font size change
     if ($retval != 0) exec("kill ".$ttyd_pid[0]);
   }
-  if ($retval != 0) exec("ttyd-exec -i '$sock' " . getUserShell() . " --login");
+  if ($retval != 0) exec("ttyd-exec -i '$sock' " . escapeshellarg(getUserShell()) . " --login");
   break;
 case 'syslog':
   // read syslog file
--- a/emhttp/plugins/dynamix/include/OpenTerminal.php
+++ b/emhttp/plugins/dynamix/include/OpenTerminal.php
@@ -51,8 +51,7 @@ case 'ttyd':
     // no child processes, restart ttyd to pick up possible font size change
     if ($retval != 0) exec("kill ".$ttyd_pid[0]);
   }
-  $shell = posix_getpwuid(0)['shell'] === '/bin/zsh' ? 'zsh' : 'bash';
-  if ($retval != 0) exec("ttyd-exec -i '$sock' '$shell' --login");
+  if ($retval != 0) exec("ttyd-exec -i '$sock' '" . posix_getpwuid(0)['shell'] . "' --login");
   break;
 case 'syslog':
   // read syslog file
--- a/emhttp/plugins/dynamix/include/OpenTerminal.php
+++ b/emhttp/plugins/dynamix/include/OpenTerminal.php
@@ -29,6 +29,26 @@ $run  = "$docroot/webGui/scripts/run_cmd";
 // set tty window font size
 if (!empty($display['tty'])) exec("sed -ri 's/fontSize=[0-9]+/fontSize={$display['tty']}/' /etc/default/ttyd");
 
+function getUserShell() {
+  $shell = 'bash';
+  try {
+      $username = posix_getpwuid(posix_geteuid())['name'];
+      $passwd = file_get_contents('/etc/passwd');
+      $lines = explode("\n", $passwd);
+      foreach ($lines as $line) {
+          if (strpos($line, $username) === 0) {
+              $parts = explode(':', $line);
+              $fullShellPath = end($parts);
+              $shell = basename(trim($fullShellPath));
+              break;
+          }
+      }
+  } catch (Exception $e) {
+      syslog(LOG_ERR, "Fehler beim Ermitteln der User-Shell: " . $e->getMessage());
+  }
+  return $shell;
+}
+
 function wait($name,$cmd) {
   global $run,$wait;
   $exec = "/var/tmp/$name.run.sh";
@@ -51,7 +71,7 @@ case 'ttyd':
     // no child processes, restart ttyd to pick up possible font size change
     if ($retval != 0) exec("kill ".$ttyd_pid[0]);
   }
-  if ($retval != 0) exec("ttyd-exec -i '$sock' bash --login");
+  if ($retval != 0) exec("ttyd-exec -i '$sock' " . getUserShell() . " --login");
   break;
 case 'syslog':
   // read syslog file
--- a/emhttp/plugins/dynamix/include/OpenTerminal.php
+++ b/emhttp/plugins/dynamix/include/OpenTerminal.php
@@ -45,10 +45,10 @@ function getUserShell() {
     }
   } catch (Throwable $t) {
     syslog(LOG_ERR, 'Error determining user shell: ' . $t->getMessage());
-    return defaultShell;
+    return $defaultShell;
   }
   
-  return defaultShell;
+  return $defaultShell;
 }
 
 function wait($name,$cmd) {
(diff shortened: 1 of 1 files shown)


desertwitch:
Doesn't this just add an additional point of failure into as crucial a functionality as the web terminal? For the majority of users this is going to be the OSes default shell, so this just seems like overhead and trouble... For those small percent of the tinkerers who install a different shell (which I guess is the reason for this PR), they'll probably be well capable of patching this in themselves... ?

It might also make it harder to diagnose problems if users start installing their own shells...

donbuehl:
Thank you for your feedback. I'd like to address the points you raised and provide additional context:

1. **Core Functionality**: This change addresses a fundamental aspect of Unix-like systems - respecting the user's chosen shell. It's not about adding a feature, but correcting an oversight in the current implementation.
2. **User Expectation**: Users should reasonably expect any terminal in the system to honor their shell preference. The current hardcoded approach doesn't meet this basic expectation.
3. **Consistency**: This change brings the web terminal in line with standard terminal behavior across the system, improving overall consistency.
4. **Minimal Risk**: The `getUserShell()` function uses `/etc/passwd`, a core Unix mechanism. The risk of failure is negligible.
5. **Maintainability**: Implementing this as a plugin would require external modifications to core functionality. This could lead to compatibility issues with future updates. Integrating it directly ensures better maintainability.
6. **Universal Benefit**: While some advanced users might manually patch this, integrating it system-wide provides a seamless solution for all users, regardless of their technical expertise.
7. **Standard Practice**: Honoring the user's shell choice is standard practice in terminal emulators. This change aligns Unraid's web terminal with this widely accepted standard.

The current hardcoded approach is a limitation that should be addressed in the core system. This change is about correcting that limitation, not adding a new feature.

desertwitch:
Sorry, my feedback message got lost somehow (stupid iOS application). My main points were that for the majority of users this'll be the Unraid OS default shell and if this doesn't introduce an additional point of failure (for something as critical as the web terminal) and would be better suited as a plugin. And if promoting installing a custom shell is something that is a good idea on an application server that seeks to keep everything streamlined (support-wise).

donbuehl:
I appreciate your concerns and would like to address them while also posing an important question:

1. **Exception Handling**: The implementation includes exception handling, defaulting to standard behavior if issues arise with custom shell detection. This ensures the web terminal's reliability as an emergency access point.
2. **No Impact on Default Functionality**: The fallback mechanism maintains core functionality if anything goes wrong with custom shell detection.
3. **Avoiding External Modifications**: This change eliminates the need for modifications to core files like tthd-exec or OpenTerminal.php to support custom shells.
4. **Centralized Management**: Incorporating this change into the core system provides a stable solution for custom shell support, reducing compatibility issues.

Given these points, I'd like to ask: Would it be better if someone wrote a plugin that makes changes to OpenTerminal.php instead?This question is crucial because:

- Plugins modifying core files can introduce instability and compatibility issues.
- Core file modifications may break with system updates.
- A centralized solution in the core system is easier to maintain and less prone to conflicts.

I'm interested in your thoughts on this. How do you see the trade-offs between a plugin-based approach and integrating this functionality into the core system?

limetech:
Thank you for the code.  Some things to point out:
* Unraid OS only includes the bash shell.  Uses would have to install a package that installs the shell they want.
* Unraid OS really does not have 'users' in the common sense, ie it's not really a multi-user server.  There are users definable but these are used to control SMB share access.

Please consider a couple changes:
* searching for the username should compare against first element of $parts so searching for user 'john' won't mistake entry for user 'johnny'.
* please use English for the syslog message

desertwitch:
Code-wise I'd also recommend "/bin/bash" as default instead of "bash" and catching Throwable instead of Exception as per: https://www.php.net/manual/en/language.errors.php7.php (which covers any possible errors starting PHP7). 

```
catch (Throwable $e)
{
   // catch any possible error
}
```

desertwitch:
Tested, works well with `bash`, one last improvement or failure case I could think of while looking over this again today.
_--login_ is a parameter specific to Bash, so it will not necessarily work with different shells and may lead to opening failure.

I'd adapt this line:
`if ($retval != 0) exec("ttyd-exec -i '$sock' " . getUserShell() . " --login");`

... to something like this:
```
if ($retval != 0) {
    $userShell = getUserShell();
    if($userShell == "bash") {
        exec("ttyd-exec -i '$sock' " . escapeshellarg($userShell) . " --login");
    } else {
        exec("ttyd-exec -i '$sock' " . escapeshellarg($userShell));
    }
}
```

limetech:
Check out what ChatGPT recommended:
https://chatgpt.com/share/c647a89e-08b7-46ef-a9b3-6b2b8c963d68

donbuehl:
After careful consideration, and reflecting on my own tendency towards conformism in this discussion, I believe we have reached a point where the following code should do the trick:

```php
if ($retval != 0) exec("ttyd-exec -i '$sock' " . basename(posix_getpwuid(0)['shell'] ?? 'bash'); . " --login");
```
I'm curious to hear if this version meets the requirements of the project.  If there are no objections, I would commit these changes.

donbuehl:
After careful consideration, and reflecting on my own tendency towards conformism in this discussion, I believe we have reached a point where the following code should do the trick:

```php
if ($retval != 0) exec("ttyd-exec -i '$sock' " . basename(posix_getpwuid(0)['shell'] ?? 'bash') . " --login");
```
I'm curious to hear if this version meets the requirements of the project.  If there are no objections, I would commit these changes.

desertwitch:
> After careful consideration, and reflecting on my own tendency towards conformism in this discussion, I believe we have reached a point where the following code should do the trick:
> 
> ```
> if ($retval != 0) exec("ttyd-exec -i '$sock' " . basename(posix_getpwuid(0)['shell']) . " --login");
> ```
> 
> I'm curious to hear if this version meets the requirements of the project. If there are no objections, I would commit these changes.

Good idea with the simplification using `posix_getpwuid(0)['shell']`, but I think it needs some failure-proofing.
Right now if `posix_getpwuid(0)['shell']` throws an error for any reason, the script would abort and not open a terminal.

I'd also drop the `basename` (even if included in the original script) as a full path to the shell is a safer practice in my opinion.
This way we could also establish that any custom shell is in fact existing and executable, which would further improve reliability.

I was thinking something along the lines of (tested & working, also in error case):
```
  $userShell="/bin/bash";
  try { $userShell=(is_executable(posix_getpwuid(0)['shell']) ? posix_getpwuid(0)['shell'] : "/bin/bash"); }
  catch (\Throwable $t) { $userShell="/bin/bash"; }
  if ($retval != 0) exec("ttyd-exec -i '$sock' '$userShell' --login");
```
This includes proofing against any throwables (errors, exceptions, ...) and also making sure any custom shell is in fact executable.

donbuehl:
If there really are problems, it would be due to the system configuration, not the code itself. There is nothing that would be thrown and could be caught, the code does not change the basic behavior but does what it is supposed to do no more and no less.

desertwitch:
> If there really are problems, it would be due to the system configuration, not the code itself. There is nothing that would be thrown and could be caught, the code does not change the basic behavior but does what it is supposed to do no more and no less.

I agree in general, but if you can account for user or system misconfiguration (e.g. typo in set shell, shell not executable due to botched package installation, function randomly failing for whatever reason)... why not just do it? Worst case the user has SSH disabled, mistyped the shell while setting it and is effectively locked out of the system (except maybe physically) because the script doesn't fall back to the default shell in such a case. Principle of least surprise, web terminal being such a critical component and all. But this is just how I'd personally do it, let's see what the others say... 😎

donbuehl:
What do you think about that:
$knownShells = ['bash', 'sh', 'zsh', 'fish', 'ksh', 'tcsh', 'dash'];
$userShell = basename(posix_getpwuid(0)['shell']);
$shell = in_array($userShell, $knownShells) ? $userShell : 'bash';

desertwitch:
> What do you think about that:
> 
> ```
> $userShell = basename(posix_getpwuid(0)['shell']);
> $shell = in_array($userShell, ['bash', 'sh', 'zsh', 'fish', 'ksh', 'tcsh', 'dash']) ? $userShell : 'bash';
> if ($retval != 0) exec("ttyd-exec -i '$sock' $shell --login");
> ```

Personally, I don't see much point in hard-coding a number of known custom shells like that, because it limits the user to that number of shells for no real reason (other than typos?) and still doesn't ensure that any set shell is in fact existent and/or executable (just ensures that it is known). It also doesn't provide any fallback mechanism in case of (admittedly unlikely) function errors, which a `try/catch`-block could mitigate, or use the full path for improved security and/or reliability. Therefore my suggestion (respectfully) remains at something like this: https://github.com/unraid/webgui/pull/1835#issuecomment-2308799669.

But, I don't want to be a deciding or, even worse, overruling factor here... it's your PR and your code, if you feel confident about it you should let it stand as it is and see what the maintainers say. I just wanted to offer my perspective as I enjoy contributing my interest in development to the Unraid OS as a way of giving back to the community for the great product that we all get to love.

donbuehl:
For my intention to implement zsh (including oh-my-zsh) as a plugin for Unraid, the following version would be completely sufficient and regardless of what posix_getpwuid(0)['shell'] returns, a fallback to bash would be triggered if it's not /bin/zsh. This point is just the only part I don't want to change in the plugin, for the reasons mentioned above.
php
$shell = posix_getpwuid(0)['shell'] === '/bin/zsh' ? 'zsh' : 'bash';
if ($retval != 0) exec("ttyd-exec -i '$sock' '$shell' --login");

I apologize if the description and the initial approach suggested a more general solution. Ultimately, the behavior above is what I actually want. A more general version would open up too many potential errors that I don't want to risk.

limetech:
Let's just do this:

`if ($retval != 0) exec("ttyd-exec -i '$sock' '" . posix_getpwuid(0)['shell'] . "' --login");`

PR #1850, opened 2024-09-05
Add support for Tailscale certs in the webgui

ljm42:


dkaser:
Looks good! Thanks!

Resolves dkaser/unraid-tailscale#22

PR #1853, opened 2024-09-06
Add a log line noting how long we wait currently for a VM action.

Commifreak:


limetech:
good idea, thx!

Commifreak:
> good idea, thx!

I had some doubts the timeout work, because one VM got killed. By adding the logline I could see the timeout was correct and thought, that would be a nice information in general. ;)

PR #1857, opened 2024-09-09
Allow users to select Container networks in the WebUI

ich777:
This PR allows users to assign different Container networks to a Docker container without using `--net=container:<CONTAINERNAME>` in the Extra Parameters.

The PR is completely backwards compatible and will not harm templates which already have `--net=container:<CONTAINERNAME>` in them.

grafik


dkaser:
This is awesome! I like it :)

limetech:
yes very nice

PR #1862, opened 2024-09-12
Fix for rc.docker for interfaces with higher index than 0

ich777:
- Fix error in first if condition which always returned `false`
- Change name from variables so that it a bit easier to read
- Make sure that custom interfaces with a higher index then 0 are properly rebuilt


limetech:
nice changes

PR #1870, opened 2024-09-23
feat: show routes from all routing tables, not just default

dkaser:
This change allows the Network Settings page to display routes from all routing tables, not just the default one.

In particular, Tailscale does not use the default routing table when adding routes. Instead, it adds routes to a separate table (route table 52) and then creates a network rule to process that table in addition to the default table. The existing Network Settings logic only polls for routes on the default table. and as a result routes added by Tailscale do not appear.

This change specifically excludes items in the local routing table, which is maintained by the kernel and contains routes for the local system. Displaying these routes would create unnecessary noise in the route display.


limetech:
thanks!

PR #1871, opened 2024-09-24
Show autostart button on 3rd party containers

ich777:
This rolls back to hide the Autostart button on 3rd party containers since a user made me aware that there are some containers out there which indeed need the Autostart function from Unraid to fully work because they lack support in their WebUI for Autostart.

I marked this as draft since I want to try this myself first.


ich777:
Closed in favor to: #1872

PR #1875, opened 2024-09-30
feat: sanitize ts.net domain names from nginx config

dkaser:
Sanitize ts.net domain names found in servers.conf.txt since these could potentially be used for remote access similar to myunraid.net domains.


limetech:
good idea, thx!

PR #1876, opened 2024-10-01
fix: prevent deleting containers that are assigned as a network

dkaser:
If a container has been assigned as a network via #1857 , deleting that container breaks the connected container:

- The container gets stuck in "rebuilding" state in an infinite loop of being recreated (see attached recording).
- Due to the loop, the Docker page in the UI gets stuck "loading" infinitely, and cannot be interacted with.
- Remove the container from CLI can succeed, but requires multiple attempts.

This change adds a check to the removal that prevents a container from being removed if it is assigned as the network for another container.

https://github.com/user-attachments/assets/46fb2efc-6974-49bd-89ec-4bd0eec00b32


limetech:
good catch

PR #1878, opened 2024-10-05
Fix virtiofsd pre processor + add virtiogpu option to gui.

SimonFair:
Libvirt does an additional check where fd is not passed. Pre processor was expecting FD to be passed. Thanks to mackid1993 for the fix.

--print-capabilities
--print-capabilities
--fd=43 --shared-dir /mnt/cache --cache always --sandbox chroot --xattr
--fd=43 --shared-dir /mnt/user/isos --cache always --sandbox chroot --xattr


SimonFair:
Add virtio gpu option. 
image

PR #1881, opened 2024-10-08
Set NTP to use interfaces.

SimonFair:
Change to use interface rather than IP.

From

# Generated entries follow:
interface ignore wildcard
interface ignore ipv6
interface listen 192.168.1.225 # br0
server time1.google.com iburst
server time2.google.com iburst
server time3.google.com iburst
server time4.google.com iburst

to

# Generated entries follow:
interface ignore wildcard
interface ignore ipv6
interface listen br0 # 192.168.1.225
server time1.google.com iburst
server time2.google.com iburst
server time3.google.com iburst
server time4.google.com iburst

### DIFF
--- a/etc/rc.d/rc.ntpd
+++ b/etc/rc.d/rc.ntpd
@@ -35,7 +35,7 @@ ntpd_build(){
     [[ $IPV6 == no ]] && echo "interface ignore ipv6" >>$CONF
     # add listen interfaces
     for NET in $BIND; do
-      echo "interface listen $NET # $(show $NET)" >>$CONF
+      echo "interface listen $(show $NET) # $NET" >>$CONF
     done
   fi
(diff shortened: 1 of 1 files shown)


limetech:
awesome!

thecode:
Does https://github.com/unraid/webgui/commit/12828eec63a27d446c06e95d85f6361ad02fb53d still needed with this PR? asking because it looked like a dirty patch and did not fix the issue anyhow

SimonFair:
> Does [12828ee](https://github.com/unraid/webgui/commit/12828eec63a27d446c06e95d85f6361ad02fb53d) still needed with this PR? asking because it looked like a dirty patch and did not fix the issue anyhow

LT is removing that for next release.

PR #1882, opened 2024-10-09
(bugfix) /etc/rc.d/rc.rsyslogd: OS-native PID for decisions

desertwitch:
bugfix: multiple running _rsyslogd_ processes (e.g. spawned by Docker containers) caused the _rc.d_ script to think that the OS-native process was already running, resulting in startup or restart failure of the OSes daemon. the proposed changes make the script act on the OS-native PID file and not other unrelated processes by name.

-- Rysz (credits to Discord user _Brendann_ for noticing this issue)

PR #1883, opened 2024-10-09
(backport) (bugfix) /etc/rc.d/rc.rsyslogd: OS-native PID for decisions

desertwitch:
backport of #1882 

bugfix: multiple running _rsyslogd_ processes (e.g. spawned by Docker containers) caused the _rc.d_ script to think that the OS-native process was already running, resulting in startup or restart failure of the OSes daemon. the proposed changes make the script act on the OS-native PID file and not other unrelated processes by name.

-- Rysz (credits to Discord user _Brendann_ for noticing this issue)

### DIFF
--- a/etc/rc.d/rc.rsyslogd
+++ b/etc/rc.d/rc.rsyslogd
@@ -7,8 +7,8 @@
 # rsyslogd_reload added by Christophe Trussardi <chris@teria.org> Sept 2011
 #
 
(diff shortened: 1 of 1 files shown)


desertwitch:
Now also amended to standard practice as outlined in https://github.com/unraid/webgui/pull/1882#issuecomment-2403392398.

PR #1900, opened 2024-10-22
 Fix for Docker Rebuild loop & patchset Tailscale php warnings

ich777:
This PR addresses the issue of a rebuild loop for Docker child Containers that utilize another Containers network.

The fix involves initially verifying if another parent Containers network is being used by the child Container. It ensures the child Container is rebuilt only if the parent Container still exists. If the parent Container is absent, the child Container will be recreated with network `none` and status `stopped`.


ich777:
+ small fix for Tailscale integration

PR #1912, opened 2024-10-29
Scale icon relative to text size

ich777:
- Make TS icon size relative to text size for automatic scaling with font changes


ich777:
Included in #1914

PR #1923, opened 2024-11-11
Run container network check if network is changed in Docker

ich777:
- Re-trigger check for available TS options on container network change


ljm42 (approved):
works great!

ljm42 (approved):
works great!

PR #1930, opened 2024-11-25
Allow custom udev rules

ich777:
This PR allows users to create custom udev rules in `/boot/config/udev` which will be load on boot


limetech:
Nice feature, thx!

PR #1940, opened 2024-12-07
Fix TS tooltip on mobile

Squidly271:


Squidly271:
Needs someone else to double check before merging

ich777:
Working fine for me on iOS (iPhone).

PR #1946, opened 2024-12-12
Set MTU of docker wireguard bridge to match wireguard default MTU

Mainfrezzer:
The MTU needs to be set to at least to the wireguard default. While it seems to work just fine, to create a docker bridge with a MTU of 1500, it breaks in some cases. For example, if you "curl https://api.minecraftservices.com/publickeys" via a container thats running over that wireguard-docker-bridge, it fails.

### DIFF
--- a/emhttp/plugins/dynamix/include/update.wireguard.php
+++ b/emhttp/plugins/dynamix/include/update.wireguard.php
@@ -114,7 +114,7 @@ function addDocker($vtun) {
   $error = false;
   [$index,$network] = newNet($vtun);
   if ($dockerd && dockerNet($vtun)) {
-    exec("docker network create $vtun --subnet=$network 2>/dev/null");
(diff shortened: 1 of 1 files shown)


limetech:
Interesting.  I think wg/docker is only via ipv4, in this case shouldn't mtu be set to 1440?

Mainfrezzer:
Nope, a docker bridge with a MTU of 1440 causes the exact same issue as having it on 1500. I havent found anything higher than the default 1420 that does work. The interesting thing is, having docker set at 1420 works flawlessly with a lower MTU set on the wireguard tunnel.

The issue is definitely within docker, because the problem occures OS agnostic and when you tunnel the whole system over wireguard while running container on a "normal" bridge. Took me way too long to figure out why things did not work 😅

limetech:
Alright!  thank you for getting to the bottom of this.  One thing with docker, it could be that a future version makes a change and the bug or variation of it comes up again :)

PR #1953, opened 2024-12-14
Limit 'lsof' command to valid mount points in /nchan/update_1.

dlandon:
- Only do 'lsof' command in 'update_1' on disk devices.  There is no need to do the 'lsof' command on UD mount points.  When a UD remote share goes offline, it can hang any 'lsof' commands.
- Add 3 second timeout so 'lsof' will always terminate and not hang.
- The 'lsof' command is used to update the stream count on the Dsshboard page on SMB shares.

### DIFF
--- a/emhttp/plugins/dynamix/nchan/update_1
+++ b/emhttp/plugins/dynamix/nchan/update_1
@@ -29,7 +29,7 @@ require_once "$docroot/webGui/include/Translations.php";
 $locale_init = $locale;
 
 /* Parse the ini files */
-$shares = parse_ini_file('state/shares.ini', true);
+$shares = parse_ini_file($varroot.'/shares.ini', true);
 $disks = parse_ini_file('state/disks.ini', true);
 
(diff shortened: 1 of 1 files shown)


bergware:
Can you restore this line
`$share = array_keys(parse_ini_file("$varroot/shares.ini",true));`

When the array is offline the variable **$shares** is empty, which results in an empty table on the dashboard page.
The line above reads the share names regardless of array state.

dlandon:
I can do that, but it doesn't look like it will make a difference unless there is something about "offline" I don't understand.  When the array is "offline", no shares show in that file.

bergware:
_> I can do that, but it doesn't look like it will make a difference_

I did a quick check and see that when the array is stopped (offline) the file shares.ini is now empty, which explains your observation. Don't know if this change for shares.ini is intentional.

dlandon:
> The global variable $shares is not defined in update_1, hence it needs to read the content directly from the file

It was added.

PR #1957, opened 2024-12-16
fix: nfsd restarts

dkaser:
rc.nfsd does not stop nfsd. Even after running `/etc/rc.d/rc.nfsd stop` and `/etc/rc.d/rc.rpc stop`, nfsd continues to run.

As a result, changes to `/etc/default/nfs` (e.g., Tailscale interfaces) are never applied.

nfsd cannot be killed using `killall`. Per the [kernel server documentation](https://docs.kernel.org/admin-guide/nfs/nfsd-admin-interfaces.html), knfsd is shut down by setting the number of threads to zero.

### DIFF
--- a/etc/rc.d/rc.nfsd
+++ b/etc/rc.d/rc.nfsd
@@ -105,9 +105,7 @@ nfsd_stop(){
     REPLY="Already stopped"
   else
     killall --ns $$ rpc.mountd 2>/dev/null
-    killall --ns $$ nfsd 2>/dev/null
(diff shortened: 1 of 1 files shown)


limetech:
Thank you!!  This has been a perplexing issue since in the past it did work to 'killall nfsd'.

dlandon:
Nice catch!  When I proposed this problem originally to ChatGPT, it did not even have an answer at that point.

PR #1984, opened 2025-01-10
Improvements to Disk Usage Graph

Bastians-Bits:
Made some improvements to the Disk Usage Graph in the Menu Bar on the Top Right. It is decoupled from the array and configurable which arrays/pools should be included in the calculations. The filtering is done by disk type, either Data (Array Disks) or Cache (Pools). In case of pools it is also limit to only drives with a valid fsFree, as only the first drive in the pool has it filled.

I am not able to get the saving of the selected disks to work, so it is kinda a WIP, but any help is appreciated!

### DIFF
--- a/emhttp/plugins/dynamix/DisplaySettings.page
+++ b/emhttp/plugins/dynamix/DisplaySettings.page
@@ -19,6 +19,8 @@ Tag="desktop"
 $void = "<img src='/webGui/images/banner.png' id='image' width='330' height='30' onclick='$(&quot;#drop&quot;).click()' style='cursor:pointer' title='_(Click to select PNG file)_'>";
 $icon = "<i class='fa fa-trash top' title='_(Restore default image)_' onclick='restore()'></i>";
 $plugins = '/var/log/plugins';
+$width    = [166,300];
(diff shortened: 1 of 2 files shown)


Squidly271:
@Bastians-Bits 

> I am not able to get the saving of the selected disks to work, so it is kinda a WIP, but any help is appreciated!

You should probably mark this as draft, but for not saving, it's because of https://github.com/unraid/webgui/blob/b8b4832fa3f31afd8f83f85ec55ba83de8779246/emhttp/plugins/dynamix/DisplaySettings.page#L249

should be `name="usage_disks"`

But you also need to add in a function to simulate a single selection instead of multiple when saving and reloading it.  See this for an example:

https://github.com/unraid/webgui/blob/62e155e0e8a5911d45eef5200d61e8d933614df2/emhttp/plugins/dynamix/ParityCheck.page#L183
https://github.com/unraid/webgui/blob/62e155e0e8a5911d45eef5200d61e8d933614df2/emhttp/plugins/dynamix/ParityCheck.page#L131-L154

And there's also conflicts, so you should have this PR against the current files in 7.2 (master).

PR #1985, opened 2025-01-11
Install `ca-certificates` during `tailscale_container_hook` operations to prevent SSL errors in some cases

loganjohnlong:
In some cases, line 105 will fail because of SSL verification issues. Namely, that there are no root CAs installed in the container environment. 

```
TAILSCALE_VERSION=$(wget -qO- 'https://pkgs.tailscale.com/stable/?mode=json' | jq -r '.TarballsVersion')
```

Installing `ca-certificates` alongside `wget` and `jq` fixes this problem.

I have tested it on a few images on my Unraid 7 environment that do (`home-asssistant/home-assistant`) and don't (`gristlabs/grist`) have `ca-certificates` present, and have found this fix to be working.


ich777:
Thanks for the PR!, didn't see it earlier!
Closing my PR #1989

PR #1996, opened 2025-01-28
Php errors cause the share edit page to malfunction.

dlandon:
When checking for min() and max() values for the array and pools used for the min free space calculation, the min() and max() functions fail on an empty array.  This can happen when an array is defined and there are no pool devices.  There are also some other cases like an included disk missing in the array that cause the issue.  These php errors create a wonky share edit page to display,

Note: This PR is re-applying a lot of changes that are in the release, but not updated to GitHub.  It appears that GuitHub did not get updated from the security work done on a separate GitHub.

### DIFF
--- a/emhttp/plugins/dynamix/ShareEdit.page
+++ b/emhttp/plugins/dynamix/ShareEdit.page
@@ -15,11 +15,15 @@ Tag="share-alt-square"
  */
 ?>
 <?
+if (!empty($name) && !array_key_exists($name, $shares)) {
+  echo "<script>done()</script>";
+  return;
+}
 $width = [123,300];
 
 if ($name == "") {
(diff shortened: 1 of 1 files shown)


limetech:
Something strange going on because the file in the 'master' branch already has exactly all those changes: https://github.com/unraid/webgui/blob/master/emhttp/plugins/dynamix/ShareEdit.page

Maybe you need to do sometihing like a 'git merge master' on your local repo?

dlandon:
Did that:
Screenshot 2025-01-28 142334

limetech:
Maybe delete this PR and regenerate?

dlandon:
It's been very confusing.  This has happened several times.  Let me work on it.

PR #2006, opened 2025-02-09
Added UnRAID VM template.

bobbintb:
Added UnRAID VM template. You can select a version to install and it will create a VM with a 2gb virtual USB drive with UnRAID installed.

## Summary by CodeRabbit

- **New Features**
	- Introduced a comprehensive interface for managing virtual machines in the UnRAID system.
	- Enabled creation, updating, and deletion of VMs with configurable resources such as memory, CPU, disk images, and peripherals.
	- Enhanced user interactions with dynamic form updates, AJAX-based feedback, and auto-complete functionality for configuration inputs.


bobbintb:
Would this be better as a plugin? It's probably not useful for the average user.

PR #2008, opened 2025-02-10
Upgrade font awesome to 6.7.2

bobbintb:
Upgrade font awesome to 6.7.2.
I think I got all the breaking changes. Let me know if I missed something. It should be mostly backwards compatible with 4.7. Some icons have been renamed but I didn't see any that are used. While 4.7 uses the style:

`<i class="fa fa-user"></i>`

version 6 uses:
`<i class="fa-solid fa-user"></i>`

though the backwards compatibility will still allow the old 4.7 style so nothing should break as far as that goes. If this gets accepted I'll update the UI accordingly with the new tag style.


ljm42:
I'm sorry, but we don't really have the appetite to upgrade Font Awesome right now. The next big UI change will be https://github.com/unraid/webgui/pull/1748

bobbintb:
> I'm sorry, but we don't really have the appetite to upgrade Font Awesome right now. The next big UI change will be #1748

Oh, no problem. I don't quit understand what you mean by not having the appetite though. Could you clarify, just for my better understanding? Thanks.

PR #2036, opened 2025-02-21
Avoid automatic shutdown while being logged in and visiting Boot.php

Mainfrezzer:
This add a check for getting a post request. Currently you can accidentally shutdown your server by just simply visiting SERVER/webGui/include/Boot.php while being logged in. This is absolutely not good 🫣

### DIFF
--- a/emhttp/plugins/dynamix/include/Boot.php
+++ b/emhttp/plugins/dynamix/include/Boot.php
@@ -157,6 +157,9 @@ $(document).ajaxSend(function(elm, xhr, s){
 </script>
 </head>
 <?
+if ($_SERVER['REQUEST_METHOD'] === 'POST'){
+
+
 $safemode = '/boot/unraidsafemode';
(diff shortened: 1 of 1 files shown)


Squidly271:
Not sure how you'd inadvertently get to Boot.php in a browser, but this does stop the server from shutting down if you do

Mainfrezzer:
> Not sure how you'd inadvertently get to Boot.php in a browser, but this does stop the server from shutting down if you do

was reported by someone using the browsers build in autocomplete feature and what i assume a quick press on the result that led the person to end up discovering it.

The shutdown didnt occure in prior versions to 7

hendrik1120:
Hi, yeah I reported the bug.

It's very easy actually. I often just type "un" in my browser which usually autocompletes to "https://unraid.example.com".
However after doing some troubleshooting on my server it defaulted to to `boot.php` so I shutdown my server by accident a couple of times.

This is a very clear regression in unraid 7. A GET request should never be able to perform such an action anyway.

I know that there is a long list of accepted bugs in unraid which don't even make it past the forums or discord. 
Many thanks to @Mainfrezzer for submitting such a quick fix.

ljm42:
Thanks for this

PR #2075, opened 2025-03-20
Backport Fix: Allow docker context menus to be above or below

Squidly271:


Squidly271:
Closing as there is an issue with the context being above and browser scaling is greater than 100% -> top of the menus are cut off on the first container, and the top is the WebGUI etc.  More important for that to always display than missing a donate link on the last container.

PR #2076, opened 2025-03-20
Refactor: Dashboard stop nchan if window not in focus

Squidly271:
## Summary by CodeRabbit

- **New Features**
  - Introduced window event handlers that automatically activate or deactivate application processes based on whether the window is in focus, enhancing responsiveness.
  - Updated the unload behavior to ensure processes are stopped consistently when navigating away or closing the application.
  - Added a new dropdown in Display Settings to control whether the dashboard updates while not in focus.

- **Configuration Changes**
  - Introduced a new configuration entry to manage dashboard update behavior when the application is not in focus.

### DIFF
--- a/emhttp/plugins/dynamix/DashStats.page
+++ b/emhttp/plugins/dynamix/DashStats.page
@@ -1822,7 +1822,12 @@ $(function() {
   });
 });
 
+var dashboardPaused = false;
 $(window).focus(function() {
+  if (dashboardPaused !== false ) {
+    removeBannerWarning(dashboardPaused);
+    dashboardPaused = false;
+  }
   dashboard.start();
   <?if ($vmusage == "Y"):?>
       vmdashusage.start();
@@ -1833,6 +1838,7 @@ $(window).focus(function() {
 });
 
 $(window).blur(function() {
+  dashboardPaused = addBannerWarning("_(Dashboard Updates Paused)_",false,true);
   dashboard.stop();
   <?if ($vmusage == "Y"):?>
     vmdashusage.stop();
@@ -1841,6 +1847,7 @@ $(window).blur(function() {
     apcups.stop();
   <?endif;?>
 });
+
 window.onunload = function(){
   dashboard.stop();
 <?if ($vmusage == "Y"):?>
--- a/emhttp/plugins/dynamix/DashStats.page
+++ b/emhttp/plugins/dynamix/DashStats.page
@@ -1822,6 +1822,25 @@ $(function() {
   });
 });
 
+$(window).focus(function() {
+  dashboard.start();
+  <?if ($vmusage == "Y"):?>
+      vmdashusage.start();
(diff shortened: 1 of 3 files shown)


bergware:
Don't do this.

The dashboard should always update, this is necessary when multiple browsers (on different systems) are opened and the dashboard is displayed on any of them. The same is true for the Main page.

Squidly271:
> necessary when multiple browsers

Multiple systems don't affect each other on this, as each of the systems would still have their browser in focus.   But, it is still completely an optional setting.

> add the token ":stop" to nchan process in the page definition (see for example docker).

That doesn't work on stock OS.


This is part and parcel of an issue which I can now very easily replicate where the dashboard memory usage in the browser balloons very quickly to consume gigabytes of RAM on the client.   It also solves other ongoing issues with nchan -> out of shared memory etc. 

The functioning of this is not really any different to how something like NetData and others operate by default.   Lose focus, lose live updates.  And previous behaviour is quite easily changed.

bergware:
Can you explain what you mean with "not in  focus"?
When you in the GUI click on a different page other then Dashboard then nchan for the dashboard is not started.

Squidly271:
Focus as in a different browser tab or different window on the client.   This fixes the memory ballooning problems on the dashboard, and also fixes nchan out of shared memory. Running two devices both open to the dashboard don't affect each other, and they'll both stay in focus.  Running two browsers on same device both open to dashboard will affect each other as only one can be in focus.  Completely understand and agree that there are times you want to have the previous behaviour and not the changes this makes.   Why there's a setting to disable this.  

All in all the pros outweigh the cons with the setting enabled IMO.  But this PR right now is a draft.  It does work specifically on the dashboard, but is going to be changed to be more generalized and work on any .page file with only a minimal code change to the .page so that defaultpagebuilder.php can handle this.

bergware:
Thanks for the explanation.

I initially misunderstood your proposed change, but it is clear now.
Both Dashboard and Main have continuous updates, it makes sense if you change can handle at least these two, or as you propose "any" page.

Maybe the option should be called "Allow page updates on concurrent browser windows" (and default to 'No') ?

Squidly271:
Lawrence has already given me some wording changes.   Main.page admittedly I'm undecided about since that's the page I watch when parity check etc is happening.  But, I'd be a hypocrite if it didn't get extended to it (and I don't have any nchan issues), but as mentioned its optional and can be disabled. 

The other page is docker that needs to be done.  On the OS, that's about it AFAIK.  For plugins and their .pages if they use nchan they're not going to see any difference in operation unless the .plg gets updated with the minor code changes required.  But the code changes would be breaking on earlier OS versions unless they get conditionally executed.

FWIW, FCP uses nchan when it does a scan, but I'm not going to implement those changes on it as it would actually break FCP altogether if it lost focus during a scan.  (It has specific messaging for start and stop)

PR #2100, opened 2025-03-28
Refactor: Remove timeout option on inactive browsers

Squidly271:
## Summary by CodeRabbit

- **Refactor**
	- Removed the adjustable timeout control from display settings.
	- Standardized the pause for live updates to a fixed 30-second delay when focus is lost.
	- Removed the `liveTime` configuration option from the default settings.


bergware:
Can you also remove the setting from default.cfg ?

PR #2102, opened 2025-03-28
More refinements

bergware:
## Summary by CodeRabbit

- **New Features**
  - Enhanced the wireless configuration interface by adding descriptive tooltips and automatic focus to the manual region entry field.
  - Increased input field limits for network names and credentials, accommodating longer values.
  - Improved regulatory domain detection for wireless startup and joining, resulting in more accurate settings.
  - Enhanced reconnection strategies for real-time data subscribers, improving connection reliability.

- **Refactor**
  - Streamlined service start/stop workflows for Nginx and nchan processes to enhance overall system reliability.

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/DockerContainers.page
+++ b/emhttp/plugins/dynamix.docker.manager/DockerContainers.page
@@ -4,7 +4,6 @@ Tag="cubes"
 Cond="is_file('/var/run/dockerd.pid')"
 Markdown="false"
 Nchan="docker_load:stop"
-Load="30"
 ---
 <?PHP
 /* Copyright 2005-2023, Lime Technology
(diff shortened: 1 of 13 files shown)


bergware:
Based on the concept of Andrew I revised the way of operation to make things less intrusive for the user.

Instead of refreshing the page after a time of inactivity, the Nchan communication is stopped. The communication is resumed by clicking on the inactive page. This works when the page remains visible (e.g. on a second screen).

When the page becomes invisible, e.g. by opening another tab in the browser or opening another app which hides the browser, then Nchan communication is stopped immediately and resumed when the browser page becomes visible again. It basically hides the temporary stopping and resuming, and appears to the user as nothing has happened.

The "housekeeping" activities refresh a page when an update is required, afterwards the inactivity timer restarts and will stop the Nchan communication again when it expires. This ensures that page information is always up to date even when the user doesn't access the page for a long time.

The setting to use either new or old behavior is not needed anymore, and removed from Display Settings.

Both Dashboard and Main page have an inactivity timer of 30 minutes, this is given in the "Load" parameter of the respective page. When the user regularly switches between pages in the GUI, this timer won't expire, only when an extended view occurs then the operation kicks in.

PR #2126, opened 2025-04-07
fix: VMTemplates modals

zackspear:
Removes duplicate CSS inclusion in template file

## Summary by CodeRabbit

- **Style**
  - Removed the external jQuery UI styling from the interface, which may result in a different visual appearance for some UI components.


zackspear:
Auto summary makes the wrong assumption.

This file was already included in `DefaultPageLayout.php` with CSS overrides included via a different file below it.

Then this page template was re-adding the default styles and overriding the overrides. ;)

ljm42 (approved):

elibosley (approved):

PR #2138, opened 2025-04-08
Fix: AdBlockers could prevent Dashboard from loading

Squidly271:
## Summary by CodeRabbit

- **Refactor**
  - Streamlined the chart data initialization process to use a static approach.
  - Transitioned away from dynamically fetching external data, which may affect how real-time information is reflected in chart displays.

### DIFF
--- a/emhttp/plugins/dynamix/DashStats.page
+++ b/emhttp/plugins/dynamix/DashStats.page
@@ -873,31 +873,30 @@ function sanitizeMultiCookie(cookieName, delimiter, removeDuplicates=false) {
 }
 
 function initCharts(clear) {
-  $.post('/webGui/include/InitCharts.php',{cmd:'get'},function(data) {
-    data = JSON.parse(data);
-    // initialize graphs entries
-    var now = new Date().getTime();
-    if (!clear) {
-      var c = data.cpu.split(';');
-      var r = data.rxd.split(';');
-      var t = data.txd.split(';');
-      for (var i=0; i < cpuline; i++) {
-        var x = now + i;
-        var y = c[i]||0; cpu.push({x,y});
-      }
-      cputime = x + 1;
-    } else {
-      // clear network graph
-      var r = ''; var t = '';
-      rxd = []; txd = [];
-    }
-    for (var i=0; i < netline; i++) {
+  // initialize graphs entries
+  var data = [];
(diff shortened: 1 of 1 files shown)


emhttp/plugins/dynamix/DashStats.page
```diff
@@ -873,31 +873,30 @@ function sanitizeMultiCookie(cookieName, delimiter, removeDuplicates=false) {
...
-      cputime = x + 1;
-    } else {
-      // clear network graph
-      var r = ''; var t = '';
-      rxd = []; txd = [];
-    }
-    for (var i=0; i < netline; i++) {
+  // initialize graphs entries
+  var data = [];
+  data.cpu = data.rxd = data.txd ="";
```
zackspear, line 878: The suggested nitpick for this would be a good change for readability.

Try to stay away from shorthands like this they're prone to errors and not great to readability.
> data.cpu = data.rxd = data.txd ="";

emhttp/plugins/dynamix/DashStats.page
```diff
@@ -873,31 +873,30 @@ function sanitizeMultiCookie(cookieName, delimiter, removeDuplicates=false) {
...
+  } else {
+    // clear network graph
+    var r = ''; var t = '';
+    rxd = []; txd = [];
+  }
+  for (var i=0; i < netline; i++) {
+    var x = now + i;
+    var y = r[i]||0; rxd.push({x,y});
+    var y = t[i]||0; txd.push({x,y});
+  }
```
zackspear, line 898: The suggested nitpick for this would be good to implement as well.

Re-declaring `y` should be avoided here.

Alternatively could theoretically avoid declaring `y` altogether in the for loop since you're not re-using it at all.

```suggestion
  for (var i=0; i < netline; i++) {
    var x = now + i;
    rxd.push({
        x,
        y: r[i]||0,
    });
    txd.push({
        x,
        y: t[i]||0,
    });
  }
```

zackspear (changes requested):
The suggested nitpicks are good changes to implement for maintainability.

emhttp/plugins/dynamix/DashStats.page
Squidly271: No worries.  I didn't look at the underlying code, but simply pulled out the $.post that wasn't required any more and could cause issues.

PR #2145, opened 2025-04-10
Fix: Don't Display Disk Shares if there are non

Squidly271:
And if no shares available at all (user and disk shares disabled) don't display Shares tab

## Summary by CodeRabbit

- **New Features**
	- Improved disk share display logic, now accounting for additional user settings to determine when disk shares are visible.
	- Refined menu visibility criteria to ensure the menu appears only when specific user or disk share settings are met.


Squidly271:
Edge case present which the original coding handles better (disk share exported, and the disk is not included in the user shares

Squidly271:
Can you revert @limetech

PR #2151, opened 2025-04-15
Add Notification: Apprise

kelliott72:
**New Feature:**
  Add support for notifications using Apprise-api, which supports over 100 notification types and has an Unraid app docker.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new notification agent configuration for Apprise, allowing users to customize server URL, notification title, message, tags, and importance level for more flexible and targeted notifications.


kelliott72:
Hi @Squidly271 - appreciate the comment, I'll go take a look at the plugin url you provided.  Just trying to understand... As far as I can tell, this works the same as any of the other notification services, it just makes a curl request to a server with the message payload,  doesn't necessarily have to be running on unraid.  There's no real unraid os dependency, maybe I should have said apprise-api, since that's the rest api to apprise - [https://github.com/caronc/apprise-api].  

For example, I'm running the gotify server on my unraid and using the gotify notification service which is already one of the options. I simply copied it's xml and tweaked it a bit.  This just keeps the notification in the same place as all of the others and uses the same framework. 

I've seen a request for support for apprise in several threads on the forum and since I wanted it too just took a stab at it.

Thanks!

Squidly271:
OK.  Just as soon as I saw you mention the apprise container then I had concerns.   But as you correctly noted (and reminded me of) is that other agents (not all of them) also have requirements for a container / executable to be operating in a container or on a separate server.  I'm good either way.

That being said, the plugin route to get the xml into the system has as an advantage that it works on OS v 7.0.0 / 7.0.1 (this PR will only take effect if merged for 7.1.0) and if for some reason the API being leveraged has a breaking change / additional options etc then via the plg route modifications can be easily made instead of the changes required in the xml only taking effect going forward.

But, you should also consider the code modifications that coderabbit suggested here https://github.com/unraid/webgui/pull/2151/files#r2044656448

ljm42:
Hi @kelliott72 Code Rabbit made a few suggestions that I think would be worth implementing here

elibosley:
Hello! We've implemented a first-party apprise that doesn't require Docker using our Apprise-go implementation. Please download this plugin through Community Apps: https://ca.unraid.net/apps/apprise-19ipbpy1xswpci

PR #2153, opened 2025-04-16
fix: on first run, restore ssh config from flash

ljm42:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SSH key and configuration restoration on system boot to ensure files are properly updated and permissions are set correctly.
  
- **Chores**
  - Enhanced logic to prevent repeated overwriting of SSH files on subsequent starts.


ljm42:
closing in favor of #2155

PR #2156, opened 2025-04-17
Refactor: Optimize publish.php

Squidly271:
Execution speed on publishing a message is 2-3x faster


Squidly271:
@coderabbit stop being nice to me

PR #2174, opened 2025-04-25
fix: session corruption via GUI boot mode

pujitm:
# Bug fix for Session Corruption in GUI Boot Mode

When accessing the webgui from Unraid's GUI boot mode, the session does not persist and thus changes on every request.

This makes sessions ephemeral to each request, instead of persisting e.g. user preferences or auth information.

Resolving this session corruption also allows the webgui to access the Unraid API in GUI mode using the session cookie.

## To Reproduce

1. Reboot into Unraid GUI mode (GUI Safe Mode may also trigger this bug; didn't test tbh)
2. Log in, and once the webgui loads, open browser dev tools and inspect browser cookies (e.g. in the storage tab)
3. Locate the session cookie whose name looks like `unraid_<md5 hash>` (it'll probably be `unraid_421aa90e079fa326b6494f812ad13e79`). Note its value.
4. Open a terminal, navigate to `/var/lib/php`, and check its contents. A corresponding session file `sess_COOKIE_VALUE` should be present, but it won't be.
5. Refresh the page or navigate to another page. Notice that the value of the cookie changes. Also note that no session is ever persisted to `/var/lib/php`.
6. You may also notice the cookie's value changin *while* you're on a page if you stay long enough. This is because a new request occurred, and php had to read disk to validate the session. There's no corresponding/persisted session, so it invalidates and "renews" the cookie.

## Context of fix

The culprit of this seems to be `plugins/dynamix/include/Translations.php`, specifically the second block of session handling logic: "remove empty and temporary session files".

```php
// start new session as required
if (!headers_sent() && session_status()==PHP_SESSION_NONE && !isset($login_locale)) {
  session_start();
  session_write_close();
}
// remove empty and temporary session files
$session = '/var/lib/php/sess_'.session_id();
if (file_exists($session)) {
  if (filesize($session)===0 || (count($_SESSION??[])==1 && isset($_SESSION['locale']))) unlink($session);
}
```

In some edge case triggered only in GUI boot mode (possibly a race condition), this will delete the current, valid session file.

In the interest of backwards compatibility and preserving existing auth behavior, I opted not to delete this code (which does resolve the issue/symptom) and instead add a self-healing hack to the top of `DefaultPageLayout`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session handling when accessing the web GUI from localhost in GUI mode, ensuring users remain logged in as root if session data is missing or incomplete.

### DIFF
--- a/emhttp/plugins/dynamix/include/DefaultPageLayout.php
+++ b/emhttp/plugins/dynamix/include/DefaultPageLayout.php
@@ -31,6 +31,41 @@ $fgcolor = in_array($theme,['white','azure']) ? '#1c1c1c' : '#f2f2f2';
 exec("sed -ri 's/^\.logLine\{color:#......;/.logLine{color:$fgcolor;/' $docroot/plugins/dynamix.docker.manager/log.htm >/dev/null &");
 
 function annotate($text) {echo "\n<!--\n",str_repeat("#",strlen($text)),"\n$text\n",str_repeat("#",strlen($text)),"\n-->\n";}
+
+/**----------------------------------------------
+ *           GUI Mode Session Fix (Hack)
+ *
+ *  Abuses the fact that pages are only accessible
+ *  with a valid session to "fix" and persist a
+ *  corrupt session upon page load.
+ * 
+ *  Since nginx ensures this behavior, the code
+ *  below should not introduce new authentication
+ *  or security vulnerabilities (e.g. localhost
+ *  spoofing attacks).
+ *
+ *  Currently, session corruption only occurs when
+ *  booting into and accessing the webgui from GUI
+ *  mode.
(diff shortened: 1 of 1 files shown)


elibosley (approved):

elibosley:
@bergware  - wanted to get your eyes on this as well, as this is related to the session cookie creation inside of Translate.php

bergware:
Perhaps the more appropriate place for this is in "template.php".

"template.php" is the origin of the session start and can be expanded as needed.

Regarding "Translations.php", this module can be called at anytime and it includes code to check the presence of an already active session, which it will use, or the absence of an active session and it will start a new session. This should work fine with your intentions.

ljm42:
CLosing this for further discussion in Asana

PR #2178, opened 2025-04-28
feat: add force checkbox to plugin install page

elibosley:
Screenshot 2025-04-28 at 16 14 34

- This also resolves the issue where the enter key would not submit the plugin install form (instead it did nothing)

## Summary by CodeRabbit

- **New Features**
	- Added a "force install" checkbox option to the plugin installation form, allowing users to enable forced installation of plugins.

---
  - https://app.asana.com/0/0/1210103652740682


zackspear (approved):
I don't love the layout but it's fine for now.

ljm42 (approved):

Squidly271:
Re: https://github.com/unraid/webgui/blob/1aaeefbc4f05e5ad9dda362df24295c5cbccdd18/emhttp/plugins/dynamix.plugin.manager/PluginInstall.page#L57

This is wrong.  "forced" only allows you to install an older version than what is already installed.  It does nothing for compatibility checks, and even if it did then on reboot the plugin would fail to install due to being incompatible

Squidly271:
@elibosley 

Maybe just use "Allows you to install the same or older version of a plugin"

PR #2181, opened 2025-05-01
Refactor: Only reload Main/Dash on visibility if Live Updates disabled

Squidly271:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of live updates when switching browser tabs or windows, ensuring updates are paused or resumed appropriately based on the live update setting.


bergware:
Thanks.
Can you also make the banner display conditional. It is not needed when live updates are allowed.

```
<?if ($display['liveUpdate'] == "no"):?>
      setTimeout(function(){bannerAlert("<?=_('Attention - operation continues in background')?> ...
<?endif;?>

```

PR #2183, opened 2025-05-06
fix: respect use_ssl with tailscale certs

dkaser:
Fixing an oddity with the Tailscale + WebGUI integration: if Tailscale is configured, rc.nginx ignores the USE_SSL setting and  always listens on the HTTPS port for all interfaces (not just tailscale1).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Nginx configuration logic to ensure Tailscale HTTPS settings are only applied when SSL is enabled.

### DIFF
--- a/etc/rc.d/rc.nginx
+++ b/etc/rc.d/rc.nginx
@@ -303,7 +303,7 @@ build_servers(){
 	EOF
     fi
   fi
-  if [[ -n $TSFQDN ]]; then
+  if [[ -n $TSFQDN && $USE_SSL != no ]]; then
     cat <<- EOF >>$SERVERS
 	#
 	# Redirect Tailscale http requests to https


ljm42:
The "Use SSL/TLS" setting mainly affects http://server.localTLD urls.

When set to Yes, http://server.localTLD redirects to https://server.localTLD.

When set to Strict, it redirects to https://ip.hash.myunraid.net

When set to No, we still setup SSL on https://ip.hash.myunraid.net so you can have Unraid Connect Remote Access without needing to use SSL for local access.  And we setup https://server.MagicDNS.ts.net for the same reason.

So this is working as intended.

dkaser:
The server will listen on local.ip:sslport regardless of the Use SSL/TLS setting.

If that's the case, then Use SSL/TLS should just be changed to be more clear about its function, since the current behavior would imply that the SSL port is available if "Use SSL/TLS" is turned off.

ljm42:
Right, the webgui always listens on the SSL port (default 443, but it can be changed). I forget exactly when that started

We could add more help text, but would you have read it? :)

This setting has evolved over time, and there is significant complexity hidden behind it. I am exploring ideas for reworking it, but it is low priority.

dkaser:
Changing the setting to something more like "Redirect HTTP to HTTPS" would convey the intended purpose without the confusion about what it does.

PR #2246, opened 2025-06-10
fix: banner icon placement

zackspear:


zackspear:
@coderabbitai - please stop adding nitpicks about using `:has` in CSS. Our target browser support it.

PR #2282, opened 2025-06-30
support authorized_keys options

wandercone:
This update attempts to improve SSH key validation by supporting advanced authorized_keys options.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of SSH authorized keys, including better detection of invalid keys and options.
  * Enhanced error messages with specific line numbers and details for easier troubleshooting.

* **New Features**
  * Added support for optional SSH key options, such as restricting access by IP or hostname.

---
  - https://app.asana.com/0/0/1210699659037010

### DIFF
--- a/emhttp/plugins/dynamix/UserEdit.page
+++ b/emhttp/plugins/dynamix/UserEdit.page
@@ -121,7 +121,7 @@ function checkKey(form) {
         if (key === 'from' && val) {
           const fromValue = val.replace(/^"|"$/g, ''); // strip quotes
           const entries = fromValue.split(',');
-          const ipRegex = /^!?(\d{1,3}\.){3}\d{1,3}(\/\d{1,2})?$/;
+          const ipRegex = /^!?(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\/(?:3[0-2]|[12]?[0-9]))?$/;
           const hostnameRegex = /^!?([a-zA-Z0-9*.-]+\.[a-zA-Z]{2,})$/;
           const wildcardIPRegex = /^!?(\d{1,3}\.){0,3}\*$/;
           for (const entryRaw of entries) {
--- a/emhttp/plugins/dynamix/UserEdit.page
+++ b/emhttp/plugins/dynamix/UserEdit.page
@@ -99,11 +99,63 @@ function upload(remove) {
   }
 }
 function checkKey(form) {
-  // check syntax of ssh keys
-  var rows = form.text.value.split('\n');
-  for (var i=0,row; row=rows[i]; i++) {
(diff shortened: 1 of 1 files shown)


bergware:
Thanks for your contribution, it looks good to me.

limetech:
Thank you!

nemchik:
@wandercone @bergware @limetech sorry to necro this, but hoping a change can be considered.

I made a contribution some time ago https://github.com/unraid/webgui/pull/1127 that added support for some key types that were not previously supported, removed support for ecdsa since it may not be considered [secure](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm#Security) (and actually also removed support for dsa keys, but I honestly hadn't noticed at the time) and included some stronger regex to check for specifically valid keys (there is a pattern to how the keys are created). Later https://github.com/unraid/webgui/pull/1820 added support for all key types (that were documented in https://github.com/nemchik/ssh-key-regex at the time). This PR actually made me go digging for the ecdsa 384 and 521 information that has been added to my repo documenting valid SSH key regex.

I am not against including the less secure keys for the sake of supporting users who for whatever reason are not using systems capable of generating and using secure keys.

With all that said, would it be considered to use the following regex to support all keys and continue checking for a valid key pattern?

```regex
^(ssh-dss AAAAB3NzaC1kc3|ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNT|ecdsa-sha2-nistp384 AAAAE2VjZHNhLXNoYTItbmlzdHAzOD|ecdsa-sha2-nistp521 AAAAE2VjZHNhLXNoYTItbmlzdHA1Mj|sk-ecdsa-sha2-nistp256@openssh.com AAAAInNrLWVjZHNhLXNoYTItbmlzdHAyNTZAb3BlbnNzaC5jb2|ssh-ed25519 AAAAC3NzaC1lZDI1NTE5|sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29t|ssh-rsa AAAAB3NzaC1yc2)[0-9A-Za-z+/]+[=]{0,3}(\s.*)?$
```

This pattern is documented at [the bottom of the readme in my repo](https://github.com/nemchik/ssh-key-regex#support-all-known-key-types-less-secure) and includes all known key types (including the less secure ones).

Without this pattern, a user could enter an invalid key to the effect of:

```text
ssh-rsa AAAAE2VjZHNhLXNoYTItbmlzdHAyNT...
# note that this claims to be an RSA key but is using the ecdsa-sha2-nistp256 starting string pattern
```

This is not a security risk by any means, but a validation check to help ensure legitimate usable keys are being entered. My repo mentioned above breaks down exactly how these patterns work and are generated.

If there are any other key types I am missing documentation for in my repo, I'm happy to add to my repo and to the regex I am requesting be used here.

If this would be accepted I don't mind to PR it myself.

wandercone:
@nemchik I don’t have any heartache as my main intention was to allow the advanced options from an ssh key. I’m not as well versed as you in key types but expanding on the regex a little was a necessity to support -sk types. I’d actually prefer if (from a me standpoint) it didn’t check keys entered leaving the onus on the user to input correctly but I could easily see someone having a public vs private key issue.

nemchik:
What you have currently will weed out pub vs priv. It's certainly not required to check the validity by using the additional regex patterns. I found it fascinating to learn about the patterns in the keys, and after finding how they all worked I wrote up that repo and made the PR here a few years ago.

Thanks again for adding ecdsa 384/521, that gave me an excuse to touch up my knowledge and document it in my repo.

PR #2305, opened 2025-07-22
Upgraded to fontawesome 7.

bobbintb:
---
  - https://app.asana.com/0/0/1210862659465962

### DIFF
--- a/.github/workflows/crowdin-upload.yml
+++ b/.github/workflows/crowdin-upload.yml
@@ -10,6 +10,7 @@ on:
 jobs:
     extract-and-upload:
         runs-on: ubuntu-latest
+        if: github.repository == 'unraid/webgui'
 
         steps:
             - name: Checkout repository
--- a/emhttp/plugins/dynamix/include/.login.php
+++ b/emhttp/plugins/dynamix/include/.login.php
@@ -616,46 +616,45 @@ $isDarkTheme = $themeHelper->isDarkTheme();
             document.body.appendChild(errorElement);
         }
 
-        let timeInSecs;
-        let ticker;
-
-        function startTimer(secs) {
-            timeInSecs = parseInt(secs);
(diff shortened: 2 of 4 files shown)


ljm42:
What do you see as the benefits of this upgrade? i.e. why is it needed?

Are any CSS or HTML changes needed or is it fully backwards compatible?

bobbintb:
For me, it's mainly that the options with the current version are much more limited (675 vs 2050). When I've been developing, I have sometimes struggled to find an icon that fits that isn't already in use by something else. Of course custom icons are always an option but I prefer the consistency and ease of using Font Awesome. For the branded icons, there are many newer brands that didn't exist at the time of the current version (from 2017). There are a few system icons that are used more than once and just in general, looking at the tools and settings pages, I don't think some of the icons are a good visual representation for their items and I think that's because of the limited options. Of course, changing any existing icons is a separate issue entirely but I think updating to 7 would aid in improving the visuals in those ways.

As far as compatibility, the older style of tagging in the HTML still works and no CSS changes needed. I haven't seen any breaking changes with the built-in icons. There is the possibility that an icon used by a plug-in may have been renamed.

ljm42:
Thanks. Did you modify font-awesome.css at all or is it taken directly from the source?

ljm42:
Thanks for this, it is a good candidate for Unraid 7.3. We can incorporate it at the beginning of the cycle and have time to adjust if there are issues.

bobbintb:
> Thanks. Did you modify font-awesome.css at all or is it taken directly from the source?

I took it directly from the source, no modification.

ljm42:
Thanks, but we are going to pass on this. The free version of FA 7 has a very different look and feel, and the paid version has licensing issues given that we accept PRs from the world. We will stick with FA 4

PR #2321, opened 2025-08-04
fix: wg0 hide keys by default

zackspear:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected an issue affecting the display of the local private key input for the wg0 tunnel, ensuring it is shown or hidden appropriately based on the presence of the local public key.


ljm42 (approved):
fixes the issue, thanks!

limetech:
That's gotta be the smallest code change to date :)

PR #2331, opened 2025-08-12
VM edit make RAM values human readable

bobbintb:
This PR modifies the memory settings in the VM page in a few ways in an attempt to be more user friendly. Below is a list of changes and why I think they should be adopted, either in part or in whole. 

1. The RAM values have been replaced with more human readable values. RAM values are currently displayed in megabytes. I think this tends to look cluttered and it is more pronounced with machines with large amounts of RAM. For example, with 128GB of RAM the drop-down will display every value in MB in 512MB increments. With larger values, there are no commas either so it can be difficult to select the correct value. See the image below for reference.
image

2. A jQuery slider with two handles for initial and maximum RAM has been added to work with the existing drop-downs. Either can be used and they will dynamically update one another. I think it looks better and is easier to use than just the drop-down, without making the UI too busy or cluttered.
3. Selecting a value in the drop-down has been changed. Previously, all values were available in each drop-down and the max value would automatically change to match the initial value. I don't think this works as well with the slider and I think the visuals can be confusing. Each drop-down still lists all available values, but disables the ones that are not selectable because the other value has put them out of range. I think visually this is more cohesive than all values being available and automatically changing other values and it is more in line with what a user would expect to see.

The inline xml view still works and the slider is disabled if the VM is running, just as other elements are. No extra libraries are needed as this uses the existing jQuery version. Here are some screenshots to illustrate the changes:

image
image
image

## Summary by CodeRabbit

* **New Features**
  * Memory size dropdowns in VM Custom settings now display human-readable values: MB for sizes under 1 GB and GB (with one decimal) for 1 GB and above, improving clarity when selecting Initial and Max memory.
  * Consistent labeling applied to both Initial and Max memory options, ensuring uniform unit presentation across the configuration.


ljm42:
Thanks for this. I think updating the dropdown with with more user friendly values makes a lot of sense. But at this point we don't really want to introduce new controls to the legacy UI. If you would be willing to update the PR to only include the more friendly dropdown settings we should be able to get it in.

Thanks!

limetech:
thank you!

PR #2343, opened 2025-08-22
Dashboard: fix network speed calculation

bergware:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected per-update timing to compute accurate interval deltas.
  * Network RX/TX speeds now display true real-time rates instead of startup-based values.
  * Removes spurious zero/placeholder readings on the first update for more consistent output.
  * Improves stability of displayed throughput across consecutive updates without changing layout or data fields.
  * Results in smoother, more reliable dashboard/network statistics visible to end users.

### DIFF
--- a/emhttp/plugins/dynamix/nchan/update_3
+++ b/emhttp/plugins/dynamix/nchan/update_3
@@ -161,7 +161,8 @@ while (true) {
 
   publish_noDupe('update3',json_encode($echo));
   ping('dashboardPing');
-  
+
   sleep(1);
+  $time0 = $time1;
   $time1 = microtime(true);
 }


Squidly271:
Thanks @bergware.  Didn't notice that line when I removed the dupe coding.  But, it is different though, because now the script doesn't actually know if the message was a dupe or not, and before you were only updating $time0 if it wasn't a dupe, and now its always updating even though the message might not get published.

Is that a problem?

Squidly271:
If it's a problem, do
```
$dupe =  publish_noDupe('update3',json_encode($echo));

if (!$dupe) $time0 = $time1;
```

(return value would be null or false if a dupe (or if an actual publish failed), and the cURL response if it did succeed (and not a dupe)

bergware:
It is not a problem, speed calculation and publishing work independently of each other.

PR #2348, opened 2025-08-28
Fix: Under certain circumstances Advanced/Basic view mode settingnwou…

Squidly271:
…ldn't save

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/sheets/DockerContainers.css
+++ b/emhttp/plugins/dynamix.docker.manager/sheets/DockerContainers.css
@@ -5,6 +5,11 @@
     display: none;
     white-space: nowrap;
 }
+.docker-view-mode {
+    display: flex;
+    justify-content: flex-end;
+    margin-bottom: 1rem;
+}
 .log {
     cursor: zoom-in;
 }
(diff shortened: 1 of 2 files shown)


emhttp/plugins/dynamix.docker.manager/sheets/DockerContainers.css
```diff
@@ -5,6 +5,11 @@
     display: none;
     white-space: nowrap;
 }
+.docker-view-mode {
```
zackspear: If this is something that's on multiple pages, we should eventually try and unify with a singular class for all instances. Rather than using a page specific css file.

It's fine for now though.

Squidly271: Looked for a class that was already pre-existing and didn't see anything that would suit, so went with this.   If/when other items get moved from title bars (eg: Main's bandwidth/counter toggle) then it does make sense to create a singular class specific for this.   

Although a couple of plugins have had to do something similar so maybe now wouldn't be bad - Any name suggestions @zackspear

zackspear: I would likely go with `ToggleViewMode`.

I've been using `PascalCase` for lots of multi-word "new" CSS classes for global elements. Simply did that to help differentiate between old and new.

As an example if there's a related nested child I would use a class like `ToggleViewMode-childName`. Which kind of takes inspiration from BEM style naming convention for CSS.

Squidly271: I'll revamp with that class name.   Some plugins will use it immediately, and there's also at least one section in the GUI (Plugins - Check For Updates) that should get moved down below the title bar even though they don't relate to the ViewMode.  If you think of something more generic I can update again.

PR #2356, opened 2025-09-07
fix: stop spinner & show error on File Manager Calculator failure

poroyo:
# Summary

When using the WebUI File Manager’s Calculator, if the backend errors out, the UI keeps the spinner running indefinitely and shows no feedback. This makes it look like work is still in progress when it isn’t, which is poor user experience. Users have no way to tell whether the job is still running or has already failed due to an error.

A common trigger is a very large selection hitting the shell’s single-argument limit (~128 KiB). For reference, `xargs --show-limits` often reports `Size of command buffer we are actually using: 131072`. In these cases the AJAX call returns 500, but the spinner never stops.

## Change

In the Calculator action (case 14), add an AJAX `.fail(...)` handler to:

- hide the spinner, and
- display a SweetAlert error dialog with the HTTP status.

```js
$.post('/webGui/include/Control.php', { mode: 'calc', source: encodeURIComponent(dfm_htmlspecialchars(source)) }, function (text) {
  clearTimeout(timers.calc);
  $('div.spinner.fixed').hide('slow');
  swal({ title: "_(Calculate Occupied Space)_", text: text, html: true, confirmButtonText: "_(Ok)_" });
}).fail(function(xhr){
  clearTimeout(timers.calc);
  $('div.spinner.fixed').hide('slow');
  swal({title:"Error", text:"Calculate failed: "+xhr.status+" "+xhr.statusText, type:"error"});
});
```

This PR intentionally updates 'Calculator only' to address the immediate UX issue. If this seems too piecemeal to merge, feel free to close; the goal is to document the issue and offer a minimal fix. Would be nice to have a global error-handling policy later. Thank you.

## Summary by CodeRabbit

* **Bug Fixes**
  * Calculate Occupied Space now fails gracefully: on calculation errors the spinner stops, related timers are cleared, and a modal alert displays the status and reason.
  * Prevents indefinite loading during storage size calculations while preserving successful result behavior.
  * Applies the same error handling to all related calculation actions for consistent UX.

### DIFF
--- a/emhttp/plugins/dynamix/Browse.page
+++ b/emhttp/plugins/dynamix/Browse.page
@@ -459,6 +459,10 @@ function doAction(action, title, id) {
       clearTimeout(timers.calc);
       $('div.spinner.fixed').hide('slow');
       swal({title:"_(Calculate Occupied Space)_",text:text,html:true,confirmButtonText:"_(Ok)_"});
+    }).fail(function(xhr){
+      clearTimeout(timers.calc);
+      $('div.spinner.fixed').hide('slow');
+      swal({title:"Error", text:"Calculate failed: "+xhr.status+" "+xhr.statusText, type:"error"});
     });
     return;
   case 15: // search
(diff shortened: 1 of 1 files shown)


limetech:
Thank you!

PR #2369, opened 2025-09-16
Fix: Don't allow charts to update before previous animation is complete

Squidly271:


Squidly271:
@coderabbitai

> 1442-1442`: **ApexCharts CPU event gating: good, but rely less on updated()**

I wish I could animationEnd doesn't fire in certain circumstances

> `1469-1478`: **Trim unused fields in netchart.customData**

Thanks I missed that

> * Older browsers without `IntersectionObserver` will throw. Guard and mark visible by default.

Not worried about compatibility with a browser from 10 years ago

> After resetting counts on resize, immediately push a render if `newData` is pending and visible, so charts don’t stall.

Not worried.  new data points come through every 1000ms constantly, and no data is ever lost.  This is simpler
 
> `1993-1997`: **Chart reset cadence change: OK; rename or comment**
> 5-minute reset matches the 300-point cap. Consider renaming `update900` or adding a short comment to reflect the new interval.

Nitpick lol

> `2697-2705`: **CPU updates: verify GraphQL availability and fallback**

This PR is only for OS versions going forward.

Squidly271:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> @Squidly271 Thanks for the detailed feedback and context! Your technical decisions make perfect sense given the constraints and requirements:
> 
> * Using `updated()` over `animationEnd` due to reliability issues is a solid pragmatic choice
> * Your browser compatibility stance is reasonable for a modern system
> * The constant 1000ms data flow architecture indeed simplifies the implementation safely
> * Targeting future OS versions removes legacy compatibility concerns
> 
> Your approach prioritizes simplicity and reliability over theoretical edge cases, which is the right call for this use case. The chart animation gating looks solid overall.
> 
> ✏️ Learnings added
> ```
> Learnt from: Squidly271
> URL: https://github.com/unraid/webgui/pull/2369
> 
> Timestamp: 2025-09-16T21:40:35.119Z
> Learning: In the unraid/webgui project, the animationEnd event for ApexCharts is unreliable and doesn't fire in certain circumstances, so the updated() event is preferred for chart update gating.
> ```
> 
> ```
> Learnt from: Squidly271
> URL: https://github.com/unraid/webgui/pull/2369
> 
> Timestamp: 2025-09-16T21:40:35.119Z
> Learning: The unraid/webgui project does not need to maintain compatibility with older browsers that lack IntersectionObserver support, as it targets modern systems only.
> ```
> 
> ```
> Learnt from: Squidly271
> URL: https://github.com/unraid/webgui/pull/2369
> 
> Timestamp: 2025-09-16T21:40:35.119Z
> Learning: The unraid/webgui project has a constant 1000ms data flow for charts, which ensures no data loss and allows for simpler implementation approaches without complex fallback handling.
> ```
> 
> > Note: Learnings are effective only in the context of similar code segments. To apply general coding guidelines, please configure [review instructions](https://docs.coderabbit.ai/guides/review-instructions). You can manage existing learnings in the [app](https://app.coderabbit.ai/learnings).
> 
> 🧠 Learnings used

Am I dreaming or is that a compliment?  I'm touched :) @coderabbitai

Squidly271:
That's something which you should tell my superiors about....   It just doesn't come across the same way if I point it out to them myself @coderabbitai

PR #2393, opened 2025-09-23
fix: remove unnecessary style attribute from enter_view select elemen…

elibosley:
…t in DashStats.page

## Summary by CodeRabbit

* **Style**
  * Reduced the minimum width of the Interface Information (enter view) dropdown from its previous fixed value to a smaller minimum, allowing a more compact appearance.
  * Improves layout consistency and reduces unnecessary horizontal spacing on narrow screens.
  * No functional changes: selecting options and view switching behave as before.


ljm42 (changes requested):
After further discussion, we need a better solution for other languages. At a minimum we should add ellipses for values that get cut off

PR #2432, opened 2025-10-22
Fix: Font style when starting up containers

Squidly271:
## Summary by CodeRabbit

* **Style**
  * Refined the Docker Containers interface styling with updates to the busy message display and input element sizing for improved visual consistency.

### DIFF
--- a/emhttp/languages/en_US/helptext.txt
+++ b/emhttp/languages/en_US/helptext.txt
@@ -542,7 +542,7 @@ The share name can be up to 40 characters, and is case-sensitive with these rest
 
 * must start with a letter
 * subsequent characters can be a letter, digit, period, underscore, or dash, except the last character may not be a period
-* cannot be one of the reserved names: flash, cache, cache2, .., disk1, disk2, ..
+* cannot be one of the reserved names: homes, global, printers, flash, boot, user, user0, dev, remotes, RecycleBin, rootshare, addons, mirror, raidz*, draid, spare, log, disk*, parity*
 :end
 
 :share_edit_comments_help:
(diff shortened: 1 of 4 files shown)


emhttp/plugins/dynamix.docker.manager/DockerContainers.page
```diff
@@ -24,7 +24,7 @@ require_once "$docroot/plugins/dynamix.docker.manager/include/DockerClient.php";
 
 $width   = $themeHelper->isTopNavTheme() ? -58: -44; // $themeHelper set in DefaultPageLayout.php
 $top     = $themeHelper->isTopNavTheme() ? 40 : 20;
-$busy    = "<i class='fa fa-spin fa-circle-o-notch'></i> "._('Please wait')."... "._('starting up containers');
+$busy    = "<i class='fa fa-spin fa-circle-o-notch'></i> <span style='font-family: clear-sans, sans-serif;'>"._('Please wait')."... "._('starting up containers')."</span>";
```
zackspear: Please do not use inline styles when possible. This is a bad pattern to and is difficult to maintain over time when it comes to consistency.

Ideally we should have a font family utility CSS class used to apply the sans serif font to specific elements as needed.

For example - I created `font-mono` to apply our monospace font to elements. https://github.com/unraid/webgui/blob/master/emhttp/plugins/dynamix/styles/default-base.css#L1905-L1907

So what we should for this is add this near the `.font-mono` class -
```css
.font-sans {
   font-family: clear-sans, sans-serif;
}
```

Then add use class to the span wrapping your text. Or any other element that will need this in the future.

zackspear (changes requested):

emhttp/plugins/dynamix.docker.manager/DockerContainers.page
Squidly271: No problems.  It was simply an easy solution since there already is inline styles present in the coding that handles this line anyways and avoided an extra class being used specifically for a page thats getting a complete overhaul.

zackspear (approved):
Thank you for tweaking the solution :)

PR #2433, opened 2025-10-22
Fix: CPU pinning crashing if 3rd party container present

Squidly271:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container update reliability by gracefully handling missing template files, preventing errors during updates.


Squidly271:
@coderabbitai - Logging isn't necessary as the container which is 3rd party won't even appear in the list when adjusting pinning

PR #2447, opened 2025-11-06
Fix: Temp in notifications

SimonFair:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated temperature value display formatting to use standard spacing between numeric values and units for improved consistency.


ljm42:
Let's use this instead, is less invasive https://github.com/unraid/webgui/pull/2449

PR #2468, opened 2025-12-02
feat: Reliable hosts file management

ljm42:
## Summary by CodeRabbit

* **New Features**
  * Hosts are now assembled from managed fragments with consolidated, annotated entries and atomic updates.
  * Hostname is applied via the system hostname mechanism and integrated into the regenerated hosts workflow.
* **Bug Fixes**
  * More robust add/remove of local host entries to avoid fragile in-place edits.
  * Improved logging/TTY handling with safer quoting and explicit successful exit status.


ljm42:
Not going to do. The /etc/HOSTNAME fixes are in #2472

PR #2494, opened 2025-12-27
Fix nchan regex to prevent intercepting /sub/ in file paths

### LINKED ISSUE #2493: File manager: Download of files laying under a folder named "sub" can not be downloaded
The usual behavior is that a file can be downloaded through the download button like for example this file:

```
root@atum:~# ls /mnt/disk2/test/foo/
debian-13.1.0-amd64-netinst.iso
```

Image

Image

But if the file is located in a directory named `sub` or in any of the subfolders of this directory, the download does not start at all:

```
root@atum:~# find /mnt/disk2/test/foo/sub
/mnt/disk2/test/foo/sub
/mnt/disk2/test/foo/sub/debian-13.1.0-amd64-netinst.iso
/mnt/disk2/test/foo/sub/bar
/mnt/disk2/test/foo/sub/bar/debian-13.1.0-amd64-netinst.iso
```

Image

Image

Even the direct download URL `http://atum/mnt/disk2/test/foo/sub/debian-13.1.0-amd64-netinst.iso` loads infinitely:

Image

I assume it has something to do with the nginx nchan api endpoint which uses "/sub" for subscriptions. Needs to be checked.

mgutt:
The regex pattern `~ /sub/(.*)$` matched `/sub/` anywhere in the path, including file paths like `/mnt/disk2/test/foo/sub/file.iso`. This caused nginx to route these requests to the nchan subscriber endpoint instead of serving the files.

Changed to `~ ^/sub/(.*)$` to only match paths that start with `/sub/`, which is the correct behavior for the nchan websocket endpoint.

Fixes #2493

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL path matching precision for subscriber endpoints to prevent unintended route matches.

### DIFF
--- a/etc/rc.d/rc.nginx
+++ b/etc/rc.d/rc.nginx
@@ -412,7 +412,7 @@ build_locations(){
 	#
 	# nchan subscriber endpoint
 	#
-	location ~ /sub/(.*)$ {
+	location ~ ^/sub/(.*)$ {
 	    nchan_subscriber;
 	    nchan_subscriber_timeout 0;
 	    # nchan_authorize_request <url here>


limetech:
Nice catch, thank you!

PR #2516, opened 2026-01-11
Fix issue #2510 and preserve empty directories during move

### LINKED ISSUE #2510: File Manager: Wrong error "Cannot move into its own subdirectory"
Tried to move a file from:
```
/mnt/disk2/test/4GB.bin
```

to:
```
/mnt/disk2/test/foo/
```

returns the following error:

>Cannot move directory into its own subdirectory

This error is only valid if the source is a directory (as it can't be recursively copied/moved into a subdirectory of its own). Probably a file check is needed.

Screenshot:

Image

mgutt:
- Fix false error 'Cannot move directory into its own subdirectory' when moving files (not directories) to subdirectories

- Fixed additional bug: Preserve empty directories during rsync rename operations (rsync --backup deletes empty dirs, so copy them first with find+rsync)

Fixes #2510 

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents moving a directory into its own subdirectory by tightening move validation
  * Better handling of empty directories during complex moves to avoid data loss or inconsistencies
  * Improved reliability and safety for multi-step move/rename operations

* **Refactor**
  * Reworked internal move/transfer workflow for safer, more robust execution while preserving external behavior

### DIFF
--- a/emhttp/plugins/dynamix/nchan/file_manager
+++ b/emhttp/plugins/dynamix/nchan/file_manager
@@ -540,12 +540,14 @@ while (true) {
                 }
               }
 
-              // target must not be a subdirectory of any source (backup-dir should be outside source tree)
-              $source_dirname = is_dir($valid_source_path) ? $valid_source_path : dirname($valid_source_path);
-              if (strpos(rtrim($target,'/') . '/', rtrim($source_dirname,'/') . '/') === 0) {
-                $reply['error'] = _('Cannot move directory into its own subdirectory');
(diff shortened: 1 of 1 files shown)


mgutt:
@limetech 
Will the new `emhttp` add `fsMountpoint` to Disks.ini? Because the API response of `/sub/devices` does not contain the `dir=` value for the disk overview and I think it is because I don't have the most recent `emhttp` on my test server.

image

image

PR #2517, opened 2026-01-11
Fix issue #2515: Clean up temporary terminal files

### LINKED ISSUE #2515: File Manager: "Open terminal here" feature does not clean up temporary files
This feature was introduced with PR #2501:

Image

It creates temporary shell scripts, which are not cleaned up:

```
root@atum:/mnt/disk1/test# ls /var/tmp/File_Manager_17681682*
/var/tmp/File_Manager_1768168229540.run.sh*  /var/tmp/File_Manager_1768168250743.run.sh*
```

In addition it does not really make sense to create timestamp based filenames. We should use the usual file manager name scheme and for example:
```
/var/tmp/file.manager.terminal.sh
```

mgutt:
- Use fixed filename 'file.manager.terminal.sh' instead of timestamp-based names
- Remove temporary script after terminal is opened
- Use user's shell from posix_getpwuid() instead of hardcoded bash
- Add sed_escape() function similar to preg_quote() with optional delimiter
- Prevents accumulation of /var/tmp/File_Manager_*.run.sh files

Fixes #2515

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized the file manager terminal identifier for consistent behavior.
  * Improved terminal startup and cleanup for more reliable launch and teardown of in-app terminals.
  * Added per-session isolation to avoid naming collisions and race conditions during terminal creation.
  * Made startup/launch sequence more robust and safer when invoking user shells.

### DIFF
--- a/emhttp/plugins/dynamix/include/OpenTerminal.php
+++ b/emhttp/plugins/dynamix/include/OpenTerminal.php
@@ -40,15 +40,6 @@ function command($path,$file) {
   global $run,$wait,$rows;
   return (file_exists($file) && substr($file,0,strlen($path))==$path) ? "$run tail -f -n $rows '$file'" : $wait;
 }
-function bre_escape($s, $delimiter = null) {
-  // escape BRE meta characters: . * [ ] ^ $ \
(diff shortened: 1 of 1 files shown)


SimonFair (approved):
Approved

PR #2520, opened 2026-01-14
fix: Remove Tabs attribute from Docker.page configuration

elibosley:
## Summary by CodeRabbit

* **Chores**
  * Modified display settings for the Docker page.


elibosley:
Tested, only reason I can think of was to clean up the UI. No reason to not revert this change.

Squidly271:
Good catch.  This would have changed all the work that was done on tabbed mode with fixed listing height I did a couple months ago

PR #2524, opened 2026-01-19
Fix: Connectivity check fails with Squid/Strict Proxies (Force CONNECT on port 80)

Nixxeee:
### The Issue
The current implementation forces `CURLOPT_HTTPPROXYTUNNEL` to `true` for all connectivity checks. 
The check URL is currently `http://www.msftncsi.com/ncsi.txt` (Plain HTTP).

When `CURLOPT_HTTPPROXYTUNNEL` is enabled on an HTTP URL, cURL sends a `CONNECT` method request on port 80 (instead of a standard `GET`). 
Most strict proxy configurations (like Squid's default `safe_ports`) explicitly block the `CONNECT` method on non-SSL ports to prevent protocol tunneling security risks. 

This causes Unraid to incorrectly report "Offline" status even when connectivity is fine.

### The Fix
This PR refactors `proxy_online` to make the tunneling option conditional:
1. It defines the check URL in a variable `$checkurl`.
2. It checks the scheme:
   - If `https://` -> Enable Tunneling (True).
   - If `http://` -> Disable Tunneling (False).

This ensures compatibility with standard proxy configurations while remaining robust if the check URL is changed to HTTPS in the future.

### References / Community Reports
https://forums.unraid.net/topic/190733-712-outgoing-proxy-how/

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outgoing proxy connection handling with dynamic URL detection and scheme-based tunneling configuration for enhanced compatibility.

### DIFF
--- a/emhttp/plugins/dynamix/include/OutgoingProxyLib.php
+++ b/emhttp/plugins/dynamix/include/OutgoingProxyLib.php
@@ -58,12 +58,17 @@ function proxy_online($proxyUrl) {
 	$rc	= true;
 
 	if ($proxyUrl) {
+		/*Define the check URL */
+		$checkurl = "http://www.msftncsi.com/ncsi.txt";
 		/* Initialize cURL session. */
-		$ch = curl_init("http://www.msftncsi.com/ncsi.txt");
-
+		$ch = curl_init($checkurl);
+		
+		/* Determine if tunneling is needed (HTTPS only). */
+		$isHttps = (parse_url($checkurl, PHP_URL_SCHEME) === 'https');
+		
 		/* Set cURL options. */
 		curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 3);	/* Timeout in seconds. */
-		curl_setopt($ch, CURLOPT_HTTPPROXYTUNNEL, true);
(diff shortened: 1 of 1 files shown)


ljm42:
The purpose of the test is to confirm whether the server can read http://www.msftncsi.com/ncsi.txt over the proxy. If the proxy you want to use does not support http then the proxy needs to be improved, it does not make sense to disable the http check for everyone else.  Or what am I missing?

Nixxeee:
I think there is a misunderstanding on the technical cause of the failure. I am not trying to disable the HTTP check, but to fix how the request is made when a proxy is configured.

Currently, the code forces CURLOPT_HTTPPROXYTUNNEL = true.

When the URL is HTTP (like the NCSI test), cURL sends a CONNECT request to the proxy on port 80.
Most standard proxies (like Squid) strictly forbid the CONNECT method on port 80 for security reasons (it's reserved for SSL/443). They expect a standard GET for HTTP.
My PR doesn't disable the check for anyone; it simply ensures that:

If the URL is http://, we use a standard GET request (Tunneling OFF).
If the URL is https://, we use the CONNECT method (Tunneling ON).

This fix actually enables the check to work for everyone using a standard proxy, which is currently broken and leads to a false 'Offline' status.

ljm42:
Thanks, I think you are right. I'll do some testing and will most likely recommend that this be merged.

PR #2530, opened 2026-01-23
fix(INET): fix ip v6 dns issue with ending ::

SimonFair:
## Summary by CodeRabbit

* **Refactor**
  * Reworked IPv6 address parsing to a clearer, component-by-component approach that preserves shorthand notation correctly.
  * Added explicit early handling for the special all-empty address form to ensure consistent output.

* **Bug Fixes**
  * Stricter validation of hexadecimal address segments; invalid segments now produce an error.
  * Correct handling of consecutive and trailing separators to prevent ambiguous or incorrect addresses.


elibosley:
@limetech  this is ready to be merged

PR #2544, opened 2026-02-12
7.2: Fix parsing ip addr with peer information

ap-wtioit:
When adding tun0 to listening interfaces in settings. Unraid 7+ tries to parse:
`ip addr show scope global primary -deprecated dev tun0`
```
36: tun0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc fq state UNKNOWN group default qlen 500
    link/none 
    inet 10.9.8.66 peer 10.9.8.65/32 scope global tun0
       valid_lft forever preferred_lft forever
```
and gets `10.9.8.66 peer 10.9.8.65/32 scope global tun0` as the "IP" of the interface. This then proceeds to break sshd and nginx config for unraid.

With this fix the peer information is removed and only the interfaces IP (in the example 10.9.8.66) is kept.

Info @wt-io-it

### DIFF
--- a/etc/rc.d/rc.library.source
+++ b/etc/rc.d/rc.library.source
@@ -57,7 +57,7 @@ good(){
 show(){
   case $# in
     1) ip -br addr show scope global primary -deprecated to $1 2>/dev/null | awk '{gsub("@.+","",$1);print $1;exit}' ;;
-    2) ip -br addr show scope global primary -deprecated $1 $2 2>/dev/null | awk '{$1=$2="";print;exit}' | sed -r 's/ metric [0-9]+//g' ;;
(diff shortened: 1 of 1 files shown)


SimonFair:
Hi @ap-wtioit I have created a pr on the 7.3 branch as this will be a 7.3 fix. Thanks for your submission. Please feel free to close this pr.

SimonFair:
It looks like the regex will only cater for ipv4, Can you test the follow version to see if that still works for IPv4, are you able to test with IPv6?

    2) ip -br addr show scope global primary -deprecated $1 $2 2>/dev/null | awk '{$1=$2="";print;exit}' | sed -r 's/ metric [0-9]+//g;s/ peer [0-9a-fA-F:.]+\/[0-9]+//g;s/^ +//g' ;;

limetech:
Merged your changes along with some mods - see #2555

ap-wtioit:
The code now merged in #2555 looks good on our system (could restart sshd and nginx with the interface present and the file patched), unfortunately we don't have IPv6 so i cannot test that. but the regex looks good to me.

Thanks for the fix.

PR #2555, opened 2026-02-23
fix for peer address.

SimonFair:
based on the original PR #2544 but revised for IPv6. Credit to ap-wtioit for the original fix.


wtaferner:
@limetech @SimonFair 
Not cool in the spirit of Open Source to remove the original author by creating a new PR instead of at least keep the author or respect by some co-authorship (Co-Authored by: @ap-wtioit) in a collaborative manner, but yeah, as long as things are getting fixed...but maybe think about it...the value of real work and Open Source should be held up especially in the upcoming times of AI and more individual approaches without or more difficult collaboration over time.

SimonFair:
@wtaferner I did request a review from the author on their PR as it would not have been IPv6 compatible. I have update the to include link and credit. Sorry mistake on my part as we needed to get the change in quickly.

PR #2577, opened 2026-03-18
fix: filter Docker ghost/dead containers from WebGUI and auto-clean stale entries

elibosley:
## Summary
- filter stale Docker `dead` container entries out of the WebGUI container list
- skip stale entries where container inspect fails (for example `no such container`)
- do **not** auto-delete or mutate container state
- keep normal container listing behavior unchanged

## Why
Docker engine/containerd shutdown races can leave orphaned dead container metadata. These appear as ghost rows in `docker ps -a` and in the UI, but are not actionable.

## References
- This PR: <https://github.com/unraid/webgui/pull/2577>
- Upstream root cause/fix discussion: <https://github.com/moby/moby/pull/51692>
- Docker 29.x backport: <https://github.com/moby/moby/pull/51693>
- Visibility change context: <https://github.com/moby/moby/commit/9f5f4f5a4273e920d5d77c1e73db8bebe65982bb>

## Validation
- `php -l emhttp/plugins/dynamix.docker.manager/include/DockerClient.php`


elibosley:
Upstream context for why users may suddenly see these after upgrading from Docker 27 to 29:

This appears to be an older Engine bug that became user-visible in v29, rather than a brand new corruption introduced by v29 itself.

Relevant upstream references:
- https://github.com/moby/moby/pull/51692
- https://github.com/moby/moby/pull/51693
- https://github.com/moby/moby/commit/9f5f4f5a4273e920d5d77c1e73db8bebe65982bb
- https://docs.docker.com/engine/release-notes/29/

Key upstream notes from Moby:
- Stopping the Engine while a container with autoremove set is running may leave a container behind in `dead` state.
- Those entries were not previously visible in `docker ps -a` before commit `9f5f4f5` because containers with no rwlayer were not added to the in-memory view.
- In v29-era builds they became visible, so users upgrading from 27 -> 29 can suddenly see stale leftovers that may have already existed on disk.

For WebGUI, this change filters those stale `dead` or uninspectable entries from the shared `getDockerContainers()` path, which is reused by the Docker list and related container UI surfaces.

I intentionally kept this fix filter-only. Auto-cleanup is probably possible, and upstream later added daemon-start cleanup in 29.1.3, but filtering is the safer first step for WebGUI because it avoids mutating Docker state from a read path.

SimonFair (approved):
LGTM

limetech (commented):
LGTM

PR #2607, opened 2026-04-03
Remove -k flag from notification agents using public APIs

Sacred-Shadow:
The `-k` flag disables TLS certificate validation in curl. Among the existing notification agents, I found 7 that call public APIs with valid SSL certificates but still use `-k` unnecessarily. 

**Affected agents:** Boxcar, Join, Prowl, Pushbullet, Pushover, Pushplus, ServerChan.

I was able to reach all 7 APIs and confirmed that each one presents a valid SSL certificate.
There is therefore no reason to use `-k`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced TLS certificate verification for notification services (Boxcar, Join, Prowl, Pushbullet, Pushover, Pushplus, ServerChan) to enforce valid certificates and improve notification system security.

### DIFF
--- a/emhttp/plugins/dynamix/agents/Boxcar.xml
+++ b/emhttp/plugins/dynamix/agents/Boxcar.xml
@@ -14,7 +14,7 @@
       ############
       MESSAGE=$(echo -e "$MESSAGE")
 
-      curl -s -k \
+      curl -s \
       -d "user_credentials=$ACCESS_TOKEN" \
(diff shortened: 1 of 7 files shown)


Sacred-Shadow:
Are you interested in that or should it be withdrawn? @elibosley

ljm42:
This LGTM

PR #2610, opened 2026-04-08
Prevent rc.docker from trying to start up non-unRAID containers

andrebrait:
* Docker Compose handles its own startup
* Both unRAID and Compose use the `net.unraid.docker.managed` label, but with different values
* unRAID trying to start Compose containers may cause them to fail to start
* This commit fixes the behavior above by only attempting to manage dockerman-managed containers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container identification and filtering logic for more accurate recognition of managed containers.
  * Enhanced status detection for container operations including availability checks, start, and stop commands.

### DIFF
--- a/etc/rc.d/rc.docker
+++ b/etc/rc.d/rc.docker
@@ -157,12 +157,12 @@ wait_daemon(){
 
 # All existing containers
 all_containers(){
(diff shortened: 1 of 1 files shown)


andrebrait:
I found a bug when restoring user networks. Will update the script with the fix.

PR #2620, opened 2026-04-21
fix(docker): prevent case-duplicate user templates

elibosley:
## Summary
- reuse existing Docker user-template filenames when the requested container name differs only by case
- avoid creating a second XML on case-sensitive boot media during template saves
- preserve the opened user-template basename for unchanged-name edits so saving cannot silently write to a different case-variant file
- reject posted source-template values with path separators and rebuild accepted `my-*.xml` basenames under `templates-user`
- resolve renamed-container old-template paths through the opened source template, or through `DockerTemplates::getUserTemplatePath()` when no source template was posted
- replace startup `find -iname` template lookup with deterministic single-file lookup

## Testing
- `php -l emhttp/plugins/dynamix.docker.manager/include/DockerClient.php`
- `php -l emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php`
- `php -l emhttp/plugins/dynamix.docker.manager/scripts/docker_init`
- `php -d short_open_tag=1 -l emhttp/plugins/dynamix.docker.manager/include/DockerClient.php`
- `php -d short_open_tag=1 -l emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php`
- `php -d short_open_tag=1 -l emhttp/plugins/dynamix.docker.manager/scripts/docker_init`
- `bash -n etc/rc.d/rc.docker`
- `git diff --check`
- focused PHP regression for case-only Docker template lookup and save behavior
- deployed to `root@192.168.0.206` and verified edit/save updates an existing mixed-case template instead of creating a lowercase duplicate
- restarted Docker on `root@192.168.0.206` with a mismatched-case autostart template probe and verified no duplicate template was created
- reproduced duplicate case-variant templates on `root@192.168.0.206` (`my-FIREfox2.xml` plus `my-Firefox2.xml`) and verified unchanged-name save from the opened `my-FIREfox2.xml` basename updates only that opened file
- verified `sourceTemplate=../../my-FIREfox2.xml` is rejected and falls back to normal canonical save behavior rather than traversing paths

## Notes
This PR is intentionally separate from `codex/docker-endpoint-mac-address` and contains only the template case load/save/startup lookup fix plus existing rename-time old-template deletion path correction. It does not add a duplicate cleanup pass and does not remove duplicate template files.

## Summary by CodeRabbit

* **New Features**
  * Created support for specifying and preserving a source user template when creating or editing containers.

* **Bug Fixes**
  * Improved user-template resolution (exact then case-insensitive) to select the correct template.
  * Fixed template cleanup on rename so the previous template is removed reliably.
  * Deferred autostart containers when their user template cannot be loaded.

* **Stability**
  * Hardened template validation and file handling to reduce startup and editing errors.

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -137,8 +137,9 @@ if (isset($_POST['contName'])) {
     // force kill container if still running after 10 seconds
     removeContainer($existing,1);
     // remove old template
-    if (strtolower($filename) != strtolower("$userTmplDir/my-$existing.xml")) {
-      @unlink("$userTmplDir/my-$existing.xml");
+    $oldFilename = $DockerTemplates->getUserTemplatePath($existing);
+    if (strtolower($filename) != strtolower($oldFilename)) {
+      @unlink($oldFilename);
     }
   }
   // Extract real Entrypoint and Cmd from container for Tailscale
--- a/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/CreateDocker.php
@@ -82,7 +82,11 @@ if (isset($_POST['contName'])) {
   if (!is_dir($userTmplDir)) mkdir($userTmplDir, 0777, true);
   $sourceTemplate = _var($_POST,'sourceTemplate',false);
   if ($sourceTemplate) $sourceTemplate = unscript(urldecode($sourceTemplate));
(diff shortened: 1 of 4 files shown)


elibosley:
Not taking this one for now. The posted value is no longer trusted as a path: the form posts only the basename, the server requires it to match `my-[^/\\]+.xml`, rejects any slash/backslash including `../../`, and reconstructs the path under `templates-user` before checking that the file exists.\n\nVerifying the XML `<Name>` would be extra authenticated-admin POST hardening, but it adds another XML parse/branch to the hot save path and is broader than this PR's scoped fix. For this change, basename-only source tracking is enough to keep edit/save pointed at the opened template without introducing path traversal risk.

PR #2621, opened 2026-04-21
feat(docker): show container MAC addresses

elibosley:
## Summary
- Preserve Docker network endpoint MacAddress values from inspect data in DockerClient.
- Update the Container IP heading to show the advanced-only / MAC context.
- Show each MAC address on the line beneath the matching Container IP in advanced view, without adding a separate table column.

## Why
Users running containers on macvlan need the container MAC address for router and network configuration. Before this change, the UI did not expose that value, so users had to SSH into Unraid and run docker inspect manually.

## Testing
- php -l emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
- php -l emhttp/plugins/dynamix.docker.manager/include/DockerContainers.php
- php -l emhttp/plugins/dynamix.docker.manager/DockerContainers.page
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Show per-network MAC addresses alongside container IPs in the advanced network view and update the column header to reflect "IP / MAC".

* **Bug Fixes**
  * Fix empty-state and loading placeholder column alignment to match the updated column count.
  * Improve handling and safe rendering when network or MAC information is absent to avoid null/undefined output.

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerContainers.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerContainers.php
@@ -155,7 +155,7 @@ foreach ($containers as $ct) {
   foreach($ct['Networks'] as $netName => $netVals) {
     $networks[] = $netName;
     $network_ip = $running ? htmlspecialchars((string)$netVals['IPAddress']) : '';
-    $network_mac = htmlspecialchars((string)($netVals['MacAddress'] ?? ''));
+    $network_mac = $running ? htmlspecialchars((string)($netVals['MacAddress'] ?? '')) : '';
     $network_ips[] = $network_mac ? "$network_ip<span class='advanced'><br>$network_mac</span>" : $network_ip;
     if (isset($ct['Networks']['host'])) {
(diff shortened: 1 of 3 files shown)


emhttp/plugins/dynamix.docker.manager/include/DockerContainers.php
```diff
@@ -154,7 +154,9 @@ function my_lang_log($text) {
   }
   foreach($ct['Networks'] as $netName => $netVals) {
     $networks[] = $netName;
-    $network_ips[] = $running ? $netVals['IPAddress'] : null;
+    $network_ip = $running ? htmlspecialchars((string)$netVals['IPAddress']) : '';
+    $network_mac = htmlspecialchars((string)($netVals['MacAddress'] ?? ''));
```
ljm42: Seems like the `$nework_mac` line should do the same `running ?` check that the `$network_ip` line does

ljm42: LGTM, thanks!

PR #2630, opened 2026-05-06
File Manager: Add Compress & Extract archive support

### LINKED ISSUE #2511: [Feature Request] File Manager: Add Compress/Extract archive
Currently it is not possible to create or extract archives through the WebGUI.

As this is one of the most wanted features mentioned in the [Support Thread](https://forums.unraid.net/topic/120982-dynamix-file-manager/) we should add this.

# Concept

**COMPRESS**
- add new COMPRESS button
- by default create Zstandard (zstd) archives if the source is `/mnt/*/appdata/*` or `/mnt/*/system/*` (or whatever appdata/docker path has been set by the user as his default)
- by default create ZIP archives if source is a user share (or share on disk or pool)
- remove parent path structure (e.g. `/mnt/disk1/sharename` is not added to archive)
- dropdown with compress algorithms
- *not sure:* dropdown with compress levels (depending on selected algorithm) - maybe to overwhelming for general users
- adopt archive filename from selected file or folder
- display progress

**EXTRACT**
- add new EXTRACT button
- by default the current path is prefilled as source
- FileTree allows browsing through path structure similar to the COPY/MOVE dialogbox (including the PopularDestinations) and selecting the source path
- *not sure:* browse archive and select files from archive to extract. This should use the default Browse.page and no additional dialogbox. So clicking on an archive should directly open the next page with the containing files. But: Could result in slow loading time depending on the type and/or size of the archive. Needs testing.

mgutt:
Closes #2511

## Summary

Adds compress and extract functionality to the Unraid File Manager.

### Compress
- New **Compress** button in the toolbar and three-dot menu
- Format selector: ZIP, TAR, TAR.GZ, TAR.BZ2, TAR.XZ, TAR.ZST, TAR.LZ4; additionally `.gz`, `.xz`, `.bz2`, `.zst`, `.lz4` when exactly one file (no folder) is selected
- **Smart default format**: reads `/boot/config/docker.cfg` to detect the user's appdata/docker paths - selects TAR.ZST for those paths (ideal for container backups), ZIP for everything else
- Custom archive name (pre-filled from selection, auto-incremented if already exists)
- Destination picker with folder tree and popular destinations
- Bulk compress: select multiple files/folders, all are packed into one archive
- Overwrite protection: checkbox (default: off) — skips the already-exists check when enabled
- Progress display: %, speed, ETA, current filename; single-file formats (`.gz` etc.) show "Running..." only — no byte-level progress (tool writes directly to output file)

image

### Extract
- New **Extract** button in toolbar (enabled only when exactly one archive is selected) and three-dot menu
- Auto-detects format from extension: `.zip`, `.tar`, `.tar.gz` / `.tgz`, `.tar.bz2` / `.tbz2`, `.tar.xz` / `.txz`, `.tar.zst`, `.tar.lz4`, `.gz`, `.bz2`, `.xz`, `.zst`, `.lz4`
- Destination picker with folder tree and popular destinations
- Overwrite protection: checkbox (default: off); ZIP uses `-n`/`-o`, TAR uses `--skip-old-files --warning=no-existing-file`; single-file formats abort with "File already exists" error when overwrite is off
- Progress display: %, speed, ETA, current filename; single-file formats (`.gz` etc.) show "Running..." only — no byte-level progress
- Clicking an archive file in the file browser now triggers a direct download instead of opening an unsupported viewer

### Permissions
Created archives and new target directories get the following permissions:
- **Shares** (default): `chown nobody:users`, `chmod 666` - accessible via SMB
- **Docker paths** (appdata, docker): `chown root:root`, `chmod 600` - prevents accidental exposure via container webservers
- **New directories** created when user types a new target path: `777 nobody:users` for shares, `755 root:root` for docker paths
- Docker path detection strips the `/mnt/<disk>/` prefix so it correctly matches on any disk or pool (e.g. `/mnt/user/appdata`, `/mnt/disk1/appdata`, `/mnt/cache/appdata` all match)

## Progress implementation

All progress parsing is handled by a dedicated `progress` script (`scripts/progress`) that is piped directly from the compress/extract command. It detects the running tool (zip/tar/unzip) by inspecting `/proc`, parses its stdout, and emits a structured log to the status file:

```
TOTAL <bytes>
FILE  <timestamp> <size> <filename>
CHUNK <timestamp> <bytes>
```

The PHP worker reads this file incrementally every 250 ms and derives %, speed, and ETA from it. CHUNK lines carry sub-second timestamps (`$EPOCHREALTIME`) for accurate speed calculation - integer-second timestamps would cause speed to snap to discrete multiples (e.g. 100 MB/1s, 100 MB/2s).

**ZIP compress**: `zip -r -du -dd -ds 100m` emits a `(size)` suffix per file followed by one dot per 100 MB. The progress script parses stdout char-by-char: detects the `(size)` suffix to emit a `FILE` line, then counts dots to emit `CHUNK` lines. `TOTAL` is pre-calculated with `du -sb` from the source file list in the zip command.

**ZIP extract**: `unzip` only prints a filename after each file is fully written, so byte-level progress cannot come from stdout. Instead, the progress script reads the manifest from `unzip -l` upfront, then polls each output file with `stat` every 0.2s to emit `CHUNK` lines proportional to growing file size. The script reads unzip stdout char-by-char with a 100ms read timeout: on timeout (file is being written) it starts polling; on newline (small file finished before timeout) it emits a FILE line directly. `TOTAL` is the sum of all file sizes from the manifest.

**TAR compress**: uses `--checkpoint=N --checkpoint-action=exec="printf 'CHUNK\n'"` (1 checkpoint = 100 MB). The progress script parses tar's verbose filename output for `FILE` lines and the `CHUNK` sentinel lines for progress. `TOTAL` is pre-calculated with `du -sb` from the source file list.

**TAR extract**: same checkpoint mechanism for `CHUNK` lines. `TOTAL` is obtained by running `tar -tvf` with a 5-second timeout before the main extract starts - instant for uncompressed archives, usually finishes within the timeout for compressed ones. If the timeout is hit, progress runs without a total (shows bytes transferred + speed only, switches to % automatically once TOTAL arrives).

**Single-file formats** (`.gz`, `.bz2`, `.xz`, `.zst`, `.lz4`): no stdout-based progress available - the decompressor writes directly to the output file. The UI shows "Running..." for the duration and "Done" on completion.

### Security
- **Path traversal prevention**: `archive_name` is validated server-side (`basename()` + empty/dot check) to prevent writing outside the target directory
- **Target path validation**: compress and extract target paths are validated against allowed roots (`/mnt`, `/boot`) server-side - same as all other file manager operations
- **Shell injection prevention**: all user-supplied paths in compress/extract shell commands are passed through `escapeshellarg()`; heredoc interpolation was replaced with pre-escaped variables

### Other
- Compress and extract target paths are tracked in popular destinations (`filemanager.json`) - same as copy/move
- Cancel during compress and single-file extract cleans up the partial `.tmp` file reliably
- Temporary file uses a `uniqid`-based name with collision-check loop, stored in the active job JSON so cancel can locate it even after a page reload
- When a new action starts, the nchan channel is immediately cleared to prevent the client briefly seeing the previous operation's final status (nchan buffers last message for 30s)
- ZIP uses `LANG=en_US.UTF-8` for both compress and extract - Unraid runs with POSIX locale which mangles non-ASCII filenames (e.g. Chinese, Arabic) as `#Uxxxx` escape sequences without this
- ZIP compress uses `-y` to store symlinks as symlinks rather than skipping them (broken symlinks and cross-device symlinks were silently dropped without this)
- TAR compress uses `--xattrs --xattrs-include='*.*' --acls --sparse`: preserves extended attributes (all namespaces: `user.*`, `security.*`, `trusted.*`), POSIX ACLs, and sparse-file holes (important for VM images, container overlay filesystems, and databases)
- TAR extract additionally uses `--numeric-owner`: restores exact uid/gid numbers instead of doing a name-based lookup (critical when restoring to a system where numeric IDs differ from the archive)

## Implementation status

- [X] Project started at 2026-02-08
- [X] Frontend (Browse.page): context menus + dialog templates
- [X] Backend (Control.php): job queue handling for compress/extract
- [X] Worker (file_manager): compression/extraction logic
- [X] Bug fixes: parameter extraction, selector scope, bulk operations
- [X] Progress display (%, speed, ETA, filename)
- [X] Tested formats
   - [X] compress zip
   - [X] compress tar
   - [X] compress tar.gz
   - [X] compress tar.bz2
   - [X] compress tar.xz
   - [X] compress tar.zst
   - [X] compress tar.lz4
- [X] extract zip (progress, %, speed, ETA, filename; `^J` newline in filenames handled; overwrite flag `-n`/`-o`)
   - [X] extract tar.gz / .tgz
   - [X] extract tar.bz2 / .tbz2
   - [X] extract tar.xz / .txz
   - [X] extract tar.zst
   - [X] extract tar.lz4
   - [X] extract tar
- [X] Edge case testing
   - [X] Hardlinks: preserved in tar (both hardlinked files share the same inode after extraction); zip always breaks hardlinks - expected behavior
   - [X] Symlinks: ZIP uses `-y` → stored as symlinks (broken + cross-device symlinks preserved); TAR stores symlinks by default
   - [X] Special chars in filenames: spaces, quotes, `$`, newlines - tested with real-world data
   - [X] Non-ASCII filenames (Chinese, Arabic, etc.): `LANG=en_US.UTF-8` on zip/unzip prevents `#Uxxxx` mangling
   - [X] xattrs/ACLs preserved on extract (tar: `--xattrs --acls`; tested with `user.author`, `user.comment`)
   - [X] Empty directories: zip `-r` includes directory entries → empty dirs preserved; verified with `diff -rq` (no differences)
   - [X] Very large archives (>4GB): zip64 auto-enabled and verified - tested with a 4.1 GB archive (single 4 GB file) and a 97 GB archive; both contain the Zip64 EOCD locator
   - [X] Concurrent compress to same target dir: file_manager runs a single worker loop (one action at a time); tmp name uses `uniqid('', true)` with a collision-check loop as an additional safeguard
- [X] Permissions
   - Shares (default): `chown nobody:users`, `chmod 666` - accessible via SMB
   - Docker paths (appdata/docker): `chown root:root`, `chmod 600` - prevents accidental exposure via container webservers
   - New directories created when user types a new target path: `777 nobody:users` for shares, `755 root:root` for docker paths
   - Detection strips `/mnt/<disk>/` prefix so it matches correctly on any disk or pool (same logic as the format auto-detection in Browse.page)
- [X] Popular destinations tracked for compress/extract (same as copy/move)
- [X] Cancel cleans up partial `.tmp` file
- [X] Clicking an archive in the file browser triggers direct download (not unsupported viewer)
- [X] Overwrite existing archive option in Compress dialog: skips the archive-already-exists check when enabled
- :no_good: "Store absolute paths" option: deliberately not implemented - absolute paths in archives are dangerous for non-technical users (extracting an archive with `/mnt/disk4/...` paths could silently overwrite data on a different disk if the disk layout changed); relative paths (the default tar/zip behavior) are always safe and sufficient
- [X] Single-file compress formats (`.gz`, `.xz`, `.bz2`, `.zst`, `.lz4`) — only shown in the format dropdown when exactly one file (no folder) is selected; no progress % yet (compressor writes directly to output file, no stdout-based progress)
   - [X] progress via `pv` or stat-polling (to be implemented)
   - [X] compress .gz / .xz / .bz2 / .zst / .lz4 — magic bytes verified for all formats
   - [X] extract .gz / .xz / .bz2 / .zst / .lz4 — MD5 matches original for all formats
   - [X] overwrite=off: "File already exists" error shown; overwrite=on: file replaced correctly
   - [X] cancel during extract: no partial/tmp file left in destination
   - [X] special characters in filename (spaces, parentheses, dots) handled correctly
   - [X] compress into same directory as source: "already exists" error when dest would collide
   - [X] extract into same directory as source: works correctly
- [ ] Settings page to configure which formats are shown in the compress dialog
- :no_entry_sign: 7z and rar support - would require bundling the binaries
- [X] Sources disappear in dialog after minimize/reopen during active operation - probably fixed, not yet confirmed
- [ ] Debug logger cleanup (before merge)

## Release Notes

* **New Features**
  * Added Compress and Extract functionality supporting multiple archive formats: zip, tar.gz, tar.bz2, tar.xz, tar.zst, and tar.lz4.
  * Smart archive format auto-selection: tar.zst for Docker/appdata paths, zip by default elsewhere.

* **Improvements**
  * Enhanced progress reporting for compression and extraction operations.
  * Improved safety checks for file operations including better rename/move validation and archive overwrite handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/unraid/webgui/pull/2630?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mgutt:
@coderabbitai resume

PR #2637, opened 2026-05-13
fix: Backport Docker VLAN gateway fallback to 7.2

elibosley:
## Summary

Backports the Docker VLAN gateway fallback fix from 7.3 to the 7.2 branch.

## Root Cause

`rc.docker` only used live default routes to determine the gateway for automatically recreated Docker custom networks. VLAN and secondary interfaces can have configured gateways in `network.cfg` without an interface-specific live default route, so Docker networks could be recreated without `--gateway`.

When Docker omitted the gateway, it could claim the first address in the subnet as the macvlan/ipvlan gateway, colliding with real VLAN gateways such as `192.168.10.1` and breaking DHCP or static-IP containers.

## What Changed

- Added `configured_gateway()` to resolve IPv4 and IPv6 gateways from `/boot/config/network.cfg`.
- Mapped bridge and bond network names back to their corresponding `eth` network config entries.
- Resolved VLAN IDs to the correct indexed gateway entries.
- Used the configured gateway as a fallback only when no live default route gateway is found.

## Validation

- `bash -n etc/rc.d/rc.docker`

Cherry-picked from `d56c259d7e4431c97042d5f67d5740f443ee81bb`.

### DIFF
--- a/etc/rc.d/rc.docker
+++ b/etc/rc.d/rc.docker
@@ -192,6 +192,51 @@ network(){
   docker network ls --filter driver="$1" --format='{{.Name}}' 2>/dev/null | grep -P "^[a-z]+$2(\$|\.)" | tr '\n' ' '
 }
 
+configured_gateway(){
+  local NETWORK=$1
+  local KEY=$2
+  local CFG=${NETWORK_CFG:-/boot/config/network.cfg}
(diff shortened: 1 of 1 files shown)


limetech (commented):
LGTM

PR #2639, opened 2026-05-14
Fix dropdown z-index on docker page when clicking on container icon

Joly0:
Fixes clicking on a docker container when page has many containers the dropdown menu will be behind the footer and even outside the visible area because the scrollbar doesnt extend as far. The fix will make sure that the dropdown will always be in front of the footer if expanded and will increase the height slightly so the dropdown doesnt overlay the footer. Tested on multiple machines across 7.2.6 and 7.3.0

Before: grafik After: grafik

Related to [OS-178](https://linear.app/lime-technology/issue/OS-178/backport-docker-dropdown-spacer-fix-for-731)

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/javascript/docker.js
+++ b/emhttp/plugins/dynamix.docker.manager/javascript/docker.js
@@ -52,7 +52,7 @@ function addDockerContainerContext(container, image, template, started, paused, 
   context.destroy('#'+id);
   context.attach('#'+id, opts);
   $('#dropdown-'+id).css('z-index', 10001)
-    .append('<li style="position:absolute;top:100%;left:0;width:1px;height:60px;pointer-events:none;list-style:none"></li>');
+    .append('<li class="docker-dropdown-spacer" aria-hidden="true" style="position:absolute;top:100%;left:0;width:1px;height:60px;pointer-events:none;list-style:none"></li>');
 }
 function addDockerImageContext(image, imageTag) {
   var opts = [];
--- a/emhttp/plugins/dynamix.docker.manager/javascript/docker.js
+++ b/emhttp/plugins/dynamix.docker.manager/javascript/docker.js
@@ -51,6 +51,8 @@ function addDockerContainerContext(container, image, template, started, paused, 
   }
   context.destroy('#'+id);
   context.attach('#'+id, opts);
+  $('#dropdown-'+id).css('z-index', 10001)
+    .append('<li style="position:absolute;top:100%;left:0;width:1px;height:60px;pointer-events:none;list-style:none"></li>');
 }
 function addDockerImageContext(image, imageTag) {
   var opts = [];


elibosley:
Thank you! This looks like a good fix to get in the OS :)

Joly0:
Hey @elibosley thanks forlooking into it and considering adding it. I was just wondering why the backport fix was instantly approved and merged but this PR here wasn't/isn't? Would just like to understand the reason for that :)

elibosley:
Hey @Joly0 - good question! We're working on a 7.2.7 release, rest assured your PR to master will be merged and tested shortly, but our priority is this next 7.2 patch release before a 7.3 follow up, hence the merge of that PR first. We'll be shipping 7.2.7 next week and likely 7.3.1 a bit later.

Joly0:
> Hey @Joly0 - good question! We're working on a 7.2.7 release, rest assured your PR to master will be merged and tested shortly, but our priority is this next 7.2 patch release before a 7.3 follow up, hence the merge of that PR first. We'll be shipping 7.2.7 next week and likely 7.3.1 a bit later. 

Ahh, got it. Makes sense. Was just a bit confused, but that explains it. Thanks ☺️

SimonFair:
Hi @Joly0 I saw you fix and have used for the VM context menus also. I did notice that even though it is now in front of the footer some of the content is off screen. So for the VMs I also implemented a change to position the menu above the icon if near the bottom of the page. #2642 is my VM PR. Might be worth doing something similar for this PR?

image

Joly0:
> Hi @Joly0 I saw you fix and have used for the VM context menus also. I did notice that even though it is now in front of the footer some of the content is off screen. So for the VMs I also implemented a change to position the menu above the icon if near the bottom of the page. #2642 is my VM PR. Might be worth doing something similar for this PR?
> image

Hey @SimonFair i also thought about it, but i decided against changing that (atleast without first discussing it) and therefore went for the most easy solution i could think of. 
In my case with my changes in place the dropdown does not overlay with the footer anymore (had that problem while testing aswell when trying different approaches) but thats why i added it simply by appending the spacer which "should" (and for me does) provide enough space between the bottom of the dropdown and the footer, so nothing should overlay anymore.
Not sure why the same fix shouldnt work for the VM tab aswell

SimonFair:
> > Hi @Joly0 I saw you fix and have used for the VM context menus also. I did notice that even though it is now in front of the footer some of the content is off screen. So for the VMs I also implemented a change to position the menu above the icon if near the bottom of the page. #2642 is my VM PR. Might be worth doing something similar for this PR?
> > image
> 
> Hey @SimonFair i also thought about it, but i decided against changing that (atleast without first discussing it) and therefore went for the most easy solution i could think of. In my case with my changes in place the dropdown does not overlay with the footer anymore (had that problem while testing aswell when trying different approaches) but thats why i added it simply by appending the spacer which "should" (and for me does) provide enough space between the bottom of the dropdown and the footer, so nothing should overlay anymore. Not sure why the same fix shouldnt work for the VM tab aswell

There are two modes in the settings, for page of auto and fixed, I normally run fixed for my vms on my test server just becaused I have so many and want the buttons at the bottom. 
image

elibosley (approved):

PR #2641, opened 2026-05-16
fix(docker): keep synchronous I/O off the Docker tab render path

tspader:
# Overview
I got tired of the Docker tab being painfully slow to load, so I dug into what was wrong. I found that an incredible amount of blocking work was being done on the main request thread:
- An icon lookup that parsed every XML template in /boot for every container (i.e. N * N)
- Two network calls to `tailscale.com` to get the DERP map and latest version
- `docker exec` calls into every container with Tailscale enabled

This is an astounding amount of work to be doing at the interval the web UI does it; it is an embarrassing amount of work to be done in such a way that it blocks the UI. I ran some sanity checks, and it takes, on my box with 30 containers:
- ~2 seconds when no containers have Tailscale enabled
- **7 seconds**(!) when even a single one does

In addition to this, I fixed an issue which would append a new `<script>` tag to the page *every three seconds*, indefinitely. 

# Changes
Here's a slightly fuller change set:
- DockerClient: stop clobbering the cached icon path; anchor unresolved containers to the default icon; memoize driver/port/host per request.
- DockerContainers: read Tailscale status, DERP map, and latest version from a JSON cache instead of fetching inline.
- New nchan/tailscale_status worker refreshes that cache on the docker_load pattern (1s timeout per container, 6h for DERP/version, self-exits when no subscribers).
- DockerContainers.page: eval poll script via new Function() instead of appending a <script> to <head> every poll; defer next poll while the tab is hidden; ping /sub/tailscalestatus to keep the worker alive.

I understand that this is all legacy code, and I understand that a few PRs have been politely declined in favor of an impending rewrite. But this PR is a very small patch which just copies a few existing patterns, more or less verbatim, to fix what I consider unacceptable in both my user experience and level of quality in something I paid for. Again, this is no shade on any of the developers, I understand the nature of code like this, which is why I spent an afternoon trying to make Unraid better rather than complain.

I'm happy to follow up with anything you need from me. Thanks for your time,
Spader

## Summary by CodeRabbit

* **New Features**
  * Live Tailscale monitoring channel so container Tailscale updates trigger the UI to refresh.
  * UI now subscribes to Tailscale status updates.

* **Performance**
  * Reduced unnecessary refreshes by pausing container list reloads when the tab is hidden and resuming on visibility.
  * Improved Tailscale and icon caching to lower overhead and speed up status and icon display.

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/DockerContainers.page
+++ b/emhttp/plugins/dynamix.docker.manager/DockerContainers.page
@@ -118,6 +118,19 @@ function LockButton() {
     $('#docker_list').sortable('destroy');
   }
 }
+// Don't poll unless the page is actually open
+function scheduleNextLoad(delay) {
+  if (typeof document.hidden !== 'undefined' && document.hidden) {
+    var onVisible = function() {
+      if (document.hidden) return;
+      document.removeEventListener('visibilitychange', onVisible);
+      loadlist();
+    };
+    document.addEventListener('visibilitychange', onVisible);
+    return;
+  }
+  setTimeout(loadlist, delay);
+}
 function loadlist(init) {
   timers.docker = setTimeout(function(){$('div.spinner.fixed').show('slow');},500);
   docker = [];
@@ -141,7 +154,7 @@ function loadlist(init) {
         instance.content(TScontent);
       }
     });
-    $('head').append('<script>'+data[1]+'<\/script>');
+    (new Function(data[1]))();
     $('.iconstatus').each(function(){
       if ($(this).hasClass('stopped')) $('div.'+$(this).prop('id')).hide();
     });
@@ -165,7 +178,7 @@ function loadlist(init) {
     }
     listview();
     $('div.spinner.fixed').hide('slow');
-    if (data[2]==1) {$('#busy').show(); setTimeout(loadlist,5000);} else if ($('#busy').is(':visible')) {$('#busy').hide(); setTimeout(loadlist,3000);}
+    if (data[2]==1) {$('#busy').show(); scheduleNextLoad(5000);} else if ($('#busy').is(':visible')) {$('#busy').hide(); scheduleNextLoad(3000);}
     <?if (_var($display,'resize')):?>
         function resizeTableColumns() { // Handle table header fixed positioning after resize
           $('#docker_containers thead,#docker_containers tbody').removeClass('fixed');
@@ -208,6 +221,9 @@ dockerload.on('message', function(msg){
     $('#cpu-'+id[0]).css('width',w1);
   }
 });
+// We don't need the event from the subscription. It's just to keep the worker running.
+var tailscaleStatus = new NchanSubscriber('/sub/tailscalestatus',{subscriber:'websocket'});
+tailscaleStatus.on('message', function(msg){});
 $(function() {
   $('.advancedview').switchButton({labels_placement:'left', on_label:"_(Advanced View)_", off_label:"_(Basic View)_", checked:$.cookie('docker_listview_mode')=='advanced'});
   $('.advancedview').change(function(){
@@ -219,6 +235,7 @@ $(function() {
   $.removeCookie('lockbutton');
   loadlist(true);
   dockerload.start();
+  tailscaleStatus.start();
 });
 
 </script>
--- a/emhttp/plugins/dynamix.docker.manager/DockerContainers.page
+++ b/emhttp/plugins/dynamix.docker.manager/DockerContainers.page
@@ -3,7 +3,7 @@ Title="Docker Containers"
 Tag="cubes"
 Cond="is_file('/var/run/dockerd.pid')"
 Markdown="false"
-Nchan="docker_load"
+Nchan="docker_load,tailscale_status"
 Tabs="false"
 ---
 <?PHP
--- a/emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
+++ b/emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
@@ -1,0 +1,123 @@
+#!/usr/bin/php -q
+<?PHP
+/* Copyright 2005-2025, Lime Technology
+ * Copyright 2012-2025, Bergware International.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 2,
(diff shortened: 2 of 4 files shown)


SpencerGraffunder:
If this really fixes the super slow web UI I've been experiencing, it might get me to pay for another year of updates.

elibosley:
Thank you for this PR! We really appreciate you taking the time to work through these performance issues, and I totally understand your sentiment about the Docker pages being very painful to work on (as well as slow 😢 )

We're making efforts to fix this with internal development as well, and I'm happy to give your PR a review and test and we can likely get this merged into 7.3.1 if all goes well.

emhttp/plugins/dynamix.docker.manager/DockerContainers.page
```diff
@@ -208,6 +221,9 @@ dockerload.on('message', function(msg){
     $('#cpu-'+id[0]).css('width',w1);
   }
 });
+// We don't need the event from the subscription. It's just to keep the worker running.
+var tailscaleStatus = new NchanSubscriber('/sub/tailscalestatus',{subscriber:'websocket'});
```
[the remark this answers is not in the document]
elibosley: @tspader - this seems like a legitimate finding to me, can you verify and look into adding this to the dockercontainers.page 


```
-Nchan="docker_load"
+Nchan="docker_load,tailscale_status"
```

ljm42:
I tested this PR on my production Unraid 7.3.0 server, including the `Nchan="docker_load,tailscale_status"` change in `DockerContainers.page` mentioned above.

The results are quite good. On my system, with multiple containers using Unraid’s Tailscale integration, the Docker page render endpoint went from about 1.7s down to about 0.3s. Wow (!!)

With the additional Nchan entry in place, the `tailscale_status` worker starts, `/var/lib/docker/unraid-tailscale-status.json` is created, and the Tailscale tooltips work. I also checked `/var/log/phplog` and `/var/log/syslog` and did not see any PR-related PHP/WebGUI errors.

One note: the current client-side subscriber keeps/starts the worker, but it does not update the tooltip DOM directly, so to see the latest tooltip data you have to refresh the page. That is similar to how it worked before, so it is not a regression.

I think this is an excellent change for 7.3

Squidly271:
One thing I personally don't like (but it's not a show stopper for me) is that the cache lives within the docker image.   Myself, I'd prefer that a reboot of the system would clear out the cache (along with an array stop / start or docker service stop/start), instead of the next 6 hour interval from when the cache was created.  

Other than that (and adding it to nchan= ), it's a great improvement in performance.

One note:  It actually will up to 6 hours for the script to be terminated if there's no listeners.   Not an issue since its sleeping, but termination happens only on a failed publish, so there's no wasted cycles.

emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
```diff
@@ -0,0 +1,123 @@
...
+  $state['containers'] = $fresh;
+  $state['ts']         = $now;
+
+  ts_save_state($state);
+
+  // Notify any watchers that the cache moved. We could push actual data here
+  // to get live tooltip updates, but that's overkill. A notification is all
+  // that we really want.
+  publish('tailscalestatus', json_encode(['ts' => $now]), 1, true);
+
```
SimonFair, line 121: For save and publish maybe worth only doing if it has changed.

elibosley (changes requested):
Hey! Please check out the comments on this, once addressed, we can get this into the codebase.

emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
elibosley: For clarification - we typically save an MD5 of the payload and then compare before publishing again, to preserve publish cycles. This is a simple optimization that may help out here.

emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
```diff
@@ -0,0 +1,123 @@
...
+    if ($status !== null) {
+      $fresh[$name] = ['data' => $status, 'ts' => $now];
+    } elseif (isset($state['containers'][$name]) && ($now - $state['containers'][$name]['ts']) < 60) {
+      $fresh[$name] = $state['containers'][$name];
+    }
+  }
+  $state['containers'] = $fresh;
+  $state['ts']         = $now;
+
+  ts_save_state($state);
```
elibosley, line 115: If this state hasn't changed, probably not worth saving, you can do the MD5 comparison and share the logic with the publish call as well.

Squidly271:
`publish_noDupe() ` handles not publishing if the payload hasn't changed, but will publish every time $abortTimeout is exceeded between publishing to allow exiting if there are no listeners.  (IE: with the 6 hour window and the 30 second abort timeout, the publish will always wind up happening.

publish has to be done one way or another - due to the extremely long time between publishing, and the rarity of changes in the payload, if the publish doesn't happen then the script won't exit if there are no listeners.

And additionally `Nchan="docker_load,tailscale_status"` not only starts up the script if its not running, but registers it with dynamix so that on array stop (or nginx reload etc) it will automatically kill the script off (the killing off of the script if there are no listeners also unregisters it with dynamix)

emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
Squidly271: Publishing has to be done @SimonFair  @elibosley @tspader 

See my full comment https://github.com/unraid/webgui/pull/2641#issuecomment-4491027386

emhttp/plugins/dynamix.docker.manager/nchan/tailscale_status
Squidly271: Instead of saving it at all, just store it in a static var.    Also has the advantage of not having anything cached that survives an array stop / start or reboot

emhttp/plugins/dynamix.docker.manager/DockerContainers.page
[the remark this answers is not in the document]
Squidly271: This is required for dynamix to be able to stop the scripts on array stop etc.

elibosley (approved):

limetech:
thank you!

SnowGodric:
So smooth now, thank you !

PR #2645, opened 2026-05-19
Fix file upload truncation on Safari

mgutt:
Fixes: https://product.unraid.net/p/7-3-0-file-manager-large-uploads-truncate-in-web-ui

**Problem**

On Safari, uploading files via the file manager resulted in only the first 20 MB being written. All subsequent chunks were silently dropped.

**Root cause**

When `xhr.send()` is called with a file-backed `Blob`, Safari sends the data correctly and receives the server's 200 response, but the XHR never transitions to `readyState 4` (DONE). As a result, `onload` never fires, the upload loop never advances to the next chunk, and the connection hangs indefinitely. This is a WebKit bug specific to file-backed Blobs as XHR request bodies.

**Fix**

Convert each file slice to an `ArrayBuffer` via `blob.arrayBuffer()` before passing it to `xhr.send()`. This uses a different WebKit code path that correctly reaches `readyState 4`.

To avoid a sequential disk-read bottleneck (read chunk N, then send chunk N, then read chunk N+1, ...), the next chunk's `arrayBuffer()` call is started immediately when the current chunk begins sending, so the disk read for chunk N+1 overlaps with the network transfer of chunk N.

**Testing**

- Chrome: 1 GB test file, ~70 MB/s, no regression
- Safari: upload completes correctly for files larger than 20 MB

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file upload speed through optimized chunk processing
  * Fixed upload stalls on Safari iOS browsers
  * User-cancelled uploads no longer display error dialogs

[Review Change Stack](https://app.coderabbit.ai/change-stack/unraid/webgui/pull/2645?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/emhttp/plugins/dynamix/Browse.page
+++ b/emhttp/plugins/dynamix/Browse.page
@@ -1247,6 +1247,7 @@ function stopUpload(file,error,errorType) {
     var message = "_(File is removed)_";
     if (errorType === 'timeout') message += "<br><br>_(Upload timed out. Please check your network connection and try again.)_";
     else if (errorType === 'network') message += "<br><br>_(Network error occurred. Please check your connection and try again.)_";
+    else if (errorType === 'read') message += "<br><br>_(Failed to read file. The file may have been removed or is inaccessible.)_";
     else if (errorType && errorType.indexOf('http') === 0) message += "<br><br>_(HTTP error: )_" + errorType.substring(5);
     setTimeout(function(){swal({title:"_(Upload Error)_",text:message,html:true,confirmButtonText:"_(Ok)_"});},200);
   }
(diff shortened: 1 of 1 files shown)


mgutt:
@coderabbitai status

mgutt:
@coderabbitai 
yes, check my last commit. it should have fixed the comments.

ljm42:
Note that this fix is for 7.3 only, it does not apply to 7.2

I was unable to repro the initial problem with Safari on macOS, and I don't have iOS to test with. But I can confirm the changes did not cause a regression... after installing the PR plugin on 7.3.0 I was still able to upload 300+MB files using Safari/Chrome on macOS as well as Chrome/Firefox on Windows. 

LGTM

PR #2649, opened 2026-05-22
fix: treat IPv4LL DHCP assignments as incomplete during boot

ljm42:
## Summary
- Treat IPv4LL/link-local DHCP assignments as incomplete during boot.
- When IPv4 DHCP is enabled and the interface has carrier, rc.inet1 now waits up to 60 seconds for a non-link-local IPv4 address instead of continuing as soon as dhcpcd assigns a 169.254.x.x address.
- Clarify DHCP fallback logs to distinguish an existing dhcpcd process from starting a new background process.

## Validation
- Ran: bash -n etc/rc.d/rc.inet1
- Tested on test4 with the updated rc.inet1 override. During boot, dhcpcd assigned 169.254.150.83 at 14:25:51, then rc.inet1 continued waiting until dhcpcd leased 10.11.1.191 at 14:25:59.
- Confirmed current test4 state: br0 has 10.11.1.191/24, default route via 10.11.1.1, and nameserver 10.11.1.1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DHCP handling to avoid unnecessary restarts when DHCP is already running and to better detect link-local-only IPv4 addresses.
  * Clarified log messages during interface bring-up/shutdown for more accurate diagnostics.

* **New Features**
  * Added IPv4-specific DHCP wait and address detection to reduce false negatives during interface initialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/unraid/webgui/pull/2649?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/etc/rc.d/rc.inet1
+++ b/etc/rc.d/rc.inet1
@@ -184,6 +184,24 @@ carrier(){
   return 1
 }
 
+# check for a non-IPv4LL address
+dhcp4_addr(){
+  ip -4 -o addr show dev $1 | awk '$4 !~ /^169[.]254[.]/ {found=1} END{exit !found}'
+}
+
+# wait for DHCP; return 2 when dhcpcd is running but only IPv4LL was assigned
+dhcp_wait(){
+  local n start=$SECONDS
+  run timeout 60 dhcpcd -w $DHCP_OPTIONS $1 || return 1
+  [[ $IP == ipv6 ]] && return 0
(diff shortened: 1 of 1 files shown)


ljm42:
Handled by https://github.com/unraid/webgui/commit/982b3c207599446989e521f13d707cf3809a4318 instead

PR #2654, opened 2026-05-28
Create status json for use by the panel control plugin backport

SimonFair:


elibosley:
Hey @SimonFair - no need to backport this, we can support in 7.3+

PR #2671, opened 2026-06-19
feat(power): configurable power button behavior + rename Power Mode → Power Options (OS-463)

elibosley:
## Summary

Implements [DST-139](https://linear.app/lime-technology/issue/DST-139/configurable-power-button-behavior-acpi-prevent-accidental-shutdowns): a native WebUI setting to control what the **physical power button** does, so an accidental bump on small-form-factor hardware (NUCs, etc.) no longer forces a shutdown.

The existing **Power Mode** settings page is renamed to **Power Options** and split into two tabs:

* **Power Mode** — the existing CPU governor controls (moved verbatim).
* **Power Button** — new action selector.

## Power Button actions

Core ships only the trivially-safe primitives:

| Action | Behavior |
| -- | -- |
| **Shut down** (default) | Clean shutdown (`/sbin/init 0`) — preserves today's behavior |
| **Reboot** | Clean reboot (`/sbin/init 6`) |
| **Do nothing** | Logs the press and takes no action — the core ask in [DST-139](https://linear.app/lime-technology/issue/DST-139/configurable-power-button-behavior-acpi-prevent-accidental-shutdowns) |

The choice is saved to `[powermode] powerbutton=` in `dynamix.cfg` (on the flash, so it persists) and read **live on each press**, so changes take effect with no reload or reboot. Unset/unexpected values fall back to **Shut down**.

## Two handlers, both neutralized

Hardware testing surfaced that a single power press fires **two** independent events:

1. **acpid** → `/etc/acpi/acpi_handler.sh`. We install our setting-aware handler over the stock Slackware one at boot via `rc.acpid` (`acpid_configure`), so it wins regardless of package install order — mirroring how `rc.cpufreq` applies the `[powermode]` governor at boot.
2. **elogind** → grabs the `Power Button` input device (PNP0C0C); its default `HandlePowerKey=poweroff` shut the box down regardless of acpid. A shipped `logind.conf.d` drop-in sets `HandlePowerKey=ignore`, making **acpid the sole authority**.

Both were verified on real hardware (elogind D-Bus `HandlePowerKey` now reports `ignore`).

## Plugin extension hook

Actions that need their own setup or are hardware-fragile (suspend, running a script) are **not** baked into core. Instead the power button is extensible so plugins add actions in their own repos:

* **UI:** `PowerButton.page` includes `plugins/*/include/powerbutton-option.php` (and `powerbutton-extra.php` for action-specific fields).
* **Action:** `unraid_power_handler.sh` dispatches any non-builtin action to `/etc/acpi/powerbutton.d/<action>`.
* Contract documented in `etc/acpi/powerbutton.d/README`.

Follow-up work to wire the first two consumers (tracked as sub-tasks of [DST-139](https://linear.app/lime-technology/issue/DST-139/configurable-power-button-behavior-acpi-prevent-accidental-shutdowns)):

* **S3 Sleep plugin** → adds a **Sleep** action.
* **User Scripts plugin** → adds a **Run script** action.

## Files

| File | Change |
| -- | -- |
| `emhttp/plugins/dynamix/PowerMode.page` | Now a tabbed container titled "Power Options" (basename kept, so `/Settings/PowerMode` URLs still work) |
| `emhttp/plugins/dynamix/PowerModeCpu.page` | New — tab 1, the CPU governor form |
| `emhttp/plugins/dynamix/PowerButton.page` | New — tab 2, action selector + plugin include hooks |
| `etc/acpi/unraid_power_handler.sh` | New — setting-aware ACPI handler with plugin dispatch |
| `etc/acpi/powerbutton.d/README` | New — plugin extension contract |
| `etc/rc.d/rc.acpid` | Installs the handler + creates `powerbutton.d/` at boot |
| `etc/elogind/logind.conf.d/20-unraid-powerbutton.conf` | New — elogind ignores the power key |
| `emhttp/languages/en_US/helptext.txt` | Help text for the new setting |

## Testing

* `bash -n` clean on the handler and `rc.acpid`; `php -l` clean on both page bodies.
* Handler routing verified for every action (shutdown / reboot / ignore / unset-default) plus plugin dispatch and missing-handler fallback.
* Boot-time install simulated and verified live on devgen.local; acpid + elogind both confirmed neutralized for "Do nothing".

## Notes for reviewers

* The new `etc/*` files (acpi, elogind, rc.acpid) reach a system via the package build, **not** the `deploy_to_unraid.sh` dev script (which only syncs `emhttp/`).
* Scope per [DST-139](https://linear.app/lime-technology/issue/DST-139/configurable-power-button-behavior-acpi-prevent-accidental-shutdowns) is **Shut down + Do nothing**; Reboot is a cheap safe addition. Suspend and Run-script are deliberately left to their plugins via the hook.

## Summary by CodeRabbit

* **New Features**
  * Added Power Options settings for configuring physical power-button behavior.
  * Choose clean shutdown, reboot, or no action when the button is pressed.
  * Added CPU power-mode selection with availability indicators and virtual-machine guidance.
  * Enabled plugin-provided power-button actions and additional configuration fields.
* **Documentation**
  * Added help text and integration guidance for power-button actions.


elibosley:
**Update — hardware testing on devgen.local revealed a second handler.**

The power button generates *two* events: an ACPI event (acpid → `acpi_handler.sh`) **and** an input-key event handled by **elogind**, whose default `HandlePowerKey=poweroff` shut the box down regardless of the acpid handler. There's a `Power Button` input device (PNP0C0C) that elogind grabs, so the acpid-only fix was not sufficient.

Added `etc/elogind/logind.conf.d/20-unraid-powerbutton.conf` (`HandlePowerKey=ignore`, `HandlePowerKeyLongPress=ignore`) so **acpid is the sole authority** over the power button. Verified live: elogind's D-Bus `HandlePowerKey` now reports `"ignore"` and the button no longer powers off in "Do nothing" mode.

Note for reviewers: the new `etc/acpi/*` and `etc/elogind/*` files land via the package build, not the `deploy_to_unraid.sh` dev script (which only syncs `emhttp/`).

elibosley:
**Update — added more actions + a plugin extension point** (commit `98fdab5ea`)

Power Button options are now: **Shut down** (default), **Reboot**, **Sleep (suspend)** (shown only when the kernel reports suspend-to-RAM support), and **Do nothing**.

Rather than baking plugin-specific actions into core, the power button is now extensible so plugins add their own actions in their own repos:

- **UI hook:** `PowerButton.page` includes `plugins/*/include/powerbutton-option.php` (and `powerbutton-extra.php` for action-specific fields).
- **Action hook:** `unraid_power_handler.sh` dispatches non-builtin actions to `/etc/acpi/powerbutton.d/<action>`. The builtin **Sleep** also defers to `powerbutton.d/sleep` if present, so the S3 Sleep plugin can do array-aware prep before suspend.
- Contract documented in `etc/acpi/powerbutton.d/README`.

This is the mechanism the **User Scripts** plugin would use to add a "Run script" action — intentionally left to that plugin rather than coupled into core.

Verified on devgen.local: all four handler branches (shutdown/reboot/sleep/ignore) dispatch correctly, the `sleep` plugin-override and generic `echo mem` fallback both work, and unknown/plugin actions route through `powerbutton.d/`.

elibosley:
**Update — Sleep is now plugin-provided, not a core option** (commit `b23289324`)

Dropped the built-in Sleep option and its generic `echo mem` suspend. Rationale: a core suspend bypasses the array/Docker/VM/wake prep the S3 Sleep plugin does and is hardware-fragile, and Unraid intentionally ships S3 sleep as a plugin rather than core.

Core now offers only the trivially-safe primitives: **Shut down**, **Reboot**, **Do nothing**. Sleep is added by the S3 Sleep plugin via the same extension hook as any other plugin action (`/etc/acpi/powerbutton.d/sleep` + `powerbutton-option.php`). The handler already routes `powerbutton="sleep"` through the plugin dir, so no special-casing remains in core.

PR #2674, opened 2026-06-19
feat(ci): PR test plugins ship a patch so they stack on shared files (OS-469)

elibosley:
## Summary
Implements [OS-469](https://linear.app/lime-technology/issue/OS-469/pr-test-plugin-ship-a-diffpatch-so-multiple-pr-plugins-can-stack-on). The per-PR test plugin used to package changed files as **whole-file copies**, overwrite them on install, and **abort** if another `webgui-pr-*` plugin already managed any of the same files — so two PRs touching one file (e.g. [#2672](https://github.com/unraid/webgui/pull/2672) + [#2673](https://github.com/unraid/webgui/pull/2673), both editing `CreateDocker.php`) couldn't be tested together.

Now it ships a **unified diff** and `patch`-applies it, so non-overlapping edits to the same file stack.

## How it works
- **Build** (`pr-plugin-build.yml`): for changed **text** files, stage base + head versions in system layout and emit `pr.patch` (`diff -ruN`, paths apply with `patch -p1` at `/`). Changed **binary** files are copied whole into `binary/` with a `binary_files.txt` list. Both go in the same tarball.
- **Plugin** (`generate-pr-plugin.sh`): on install, `patch -p1 --dry-run --forward` first — **abort with a clear message on a real reject** (overlapping change), otherwise apply and save `applied.patch`. Binaries are whole-file replaced with a per-binary conflict guard. On remove (and before update), reverse the patch (`patch -R`) and restore binary backups.
- **Upload/R2 plumbing unchanged** — still a single tarball with the same URL/SHA wiring.

## Validation (local simulation)
- Two patches editing the same file at different lines → **both apply, both changes coexist**.
- Reversing one → **only that change is removed, the other stays**.
- Overlapping edits → **dry-run fails → install aborts** cleanly.
- Generated `.plg` is well-formed XML (all placeholders substituted); `bash -n` + YAML parse clean.

## Notes / still to verify
- Real validation needs CI + an Unraid box installing two stacked PR plugins. Conveniently, **this PR touches `.github/**`, so it will itself build a PR test plugin with the new code** — a live smoke test of the generator.
- Binary deletions aren't handled (rare); text add/modify/delete are.
- `patch` is present in the Unraid base; uses `--forward`/`--batch` for non-interactive apply with fuzz.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Improved plugin install/update and uninstall to use a deterministic, patch-based manifest workflow for safer reversibility.
  * Enhanced plugin package generation to build a patch-oriented payload, distinguishing text changes from full binary replacements.
  * Updated the uninstall confirmation flow to reflect the new removal behavior.

### DIFF
--- a/.github/scripts/generate-pr-plugin.sh
+++ b/.github/scripts/generate-pr-plugin.sh
@@ -107,7 +107,13 @@ if [ -f "$PLUGIN_DIR/text_files.txt" ]; then
     for other in /boot/config/plugins/webgui-pr-*; do
         [ -d "$other" ] || continue
         [ "$other" == "$PLUGIN_DIR" ] && continue
-        [ -f "$other/applied.patch" ] && patch -p1 -d / --forward < "$other/applied.patch" >/dev/null 2>&1 || true
+        [ -f "$other/applied.patch" ] || continue
+        # Non-blocking: all patches passed dry-run at install time, so failure here
+        # is a rare edge case worth surfacing rather than swallowing.
+        if ! patch -p1 -d / --forward --batch < "$other/applied.patch" >/dev/null 2>&1; then
+            echo "⚠️  Warning: could not re-apply $(basename "$other")'s changes; reinstall it or reboot to restore them"
+            logger -t webgui-pr "re-apply of $(basename "$other") patch failed during $(basename "$PLUGIN_DIR") rollback"
+        fi
     done
 fi
 rm -f "$PATCH_APPLIED"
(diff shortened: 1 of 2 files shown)


elibosley:
**Update — removal is now deterministic (rebuild from originals)** (commit `c6b330eb6`)

Addressed a real weakness in the first cut: removal used `patch -R` (reverse-apply), which can **fail when another stacked PR plugin edited nearby lines** (the reverse context is disturbed), leaving a file half-reverted.

Now the payload also ships each text file's **original (base) version** (`orig/`). On removal (and before an update), the plugin:
1. restores the files it touched to their shipped originals, then
2. re-applies every **other still-installed** PR plugin's patch (`patch --forward`),

so it cleanly drops only its own changes regardless of overlap/order — no dependence on reverse-patch context.

Verified by simulation, including the hard case: two plugins editing **adjacent lines** of the same file. Old reverse-apply **fails**; the new rebuild removes one cleanly, keeps the other intact, and returns the file to stock when the last is removed.

(Files live in the RAM rootfs, so a reboot is always the ultimate fallback — but removal no longer needs it.)

elibosley:
Thanks @coderabbitai — both nitpicks addressed in `056c371b0`:

1. **`--batch` on forward applies** — added to the dry-run, the main install apply, and both re-apply loops (the reverse ops already had it), so no `patch --forward` can hang on input.
2. **Silent re-apply failures** — the re-apply loop now emits a warning to stdout and `logger -t webgui-pr` when another plugin's patch fails to re-apply during rollback, while keeping the non-blocking behavior.

PR #2676, opened 2026-06-19
feat(notify): persistent/keyed notifications; reboot notice -> bell (OS-471)

elibosley:
## Summary
Producer side of **persistent ("sticky until resolved") notifications** — the webgui half of replacing the floating yellow banner with the API notification **bell** ([OS-471](https://linear.app/lime-technology/issue/OS-471/persistent-sticky-until-resolved-notifications-to-replace-the-legacy)). Pairs with **unraid/api#2033** (which adds the `persistent`/`key` fields, idempotent keyed create, and `clearNotificationByKey`).

## Changes
- **`webGui/scripts/notify`** (canonical file; `webGui/scripts/notify` is a symlink to it):
  - `add` gains **`-k <key>`** — a stable key; keyed notices use the key as the filename so re-raising is **idempotent** (no dupes).
  - `add` gains **`-p`** — **persistent**: written as `persistent=true`, and *not* copied to the archive folder (it isn't an archival event).
  - new **`notify clear -k <key>`** — resolves a keyed notification (removes it from the bell).
  - `usage()` updated.
- **`PluginAPI.php`** — the **reboot-required** notice now also raises a persistent, keyed (`reboot-required`) **ALERT** notification, and **clears** it once no reboot reasons remain. It also clears naturally on reboot (RAM-backed `/tmp` is wiped and the bell repopulates from files).

## How it combines with the API PR
The API watches `/tmp/notifications/unread/*.notify`; these new fields flow straight through:
`notify add -p -k reboot-required …` → a sticky ALERT in the bell that the user can't dismiss and that auto-resolves. The API's `getLegacyScriptArgs` already passes `-k`/`-p`, so API-originated persistent notifications use this same script.

## Validation
- `php -l` clean on both files.
- End-to-end (reboot notice → bell, clear on resolve) is best verified on hardware with both PR test plugins installed (the diff-based stacking work makes that possible).

## Scope / follow-ups
This wires the clearest condition (reboot-required) end to end. Remaining banner call sites — **boot-device-corrupt** → persistent, and **transient** ones (popup blocked, etc.) → toaster — plus removing the `.upgrade_notice` element, are the documented next step.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added stable-key persistent and conditional notifications, plus `notify clear -k` to dismiss specific keyed alerts.
  * Updated reboot-required and PR test build messaging to use keyed notification-bell persistence.
  * Replaced the legacy upgrade banner with a unified banner/notification system (persistent bell warnings and transient toast notifications).
* **Bug Fixes**
  * Reboot notifications no longer stack; repeated notices update the same bell entry.
  * Reboot notices automatically clear once all reboot reasons are resolved.

### DIFF
--- a/.github/scripts/generate-pr-plugin.sh
+++ b/.github/scripts/generate-pr-plugin.sh
@@ -252,40 +252,33 @@ Link='nav-user'
 ---
 <script>
   $(function() {
-    // Check for updates (non-dismissible)
-    caPluginUpdateCheck("webgui-pr-PR_PLACEHOLDER.plg", {noDismiss: true},function(result){
+    // Surface the active PR test build as a persistent, keyed notification in the
+    // bell (not a banner): keyed so it dedupes across reloads, links to the
+    // Plugins page to remove it, and is cleared by the plugin's removal script.
+    function raisePrNotice() {
+      $.post('/webGui/include/Notify.php', {
+        cmd: 'add',
+        i: 'warning',
+        e: 'PR test build installed',
+        s: 'Modified GUI via webgui-pr-PR_PLACEHOLDER',
+        d: 'A pull-request test build is active. Remove it from Plugins before applying production updates.',
+        l: '/Plugins',
+        p: '1',
+        k: 'webgui-pr-PR_PLACEHOLDER'
+      });
+    }
+    // If the PR has been merged/removed upstream, clear the notice; otherwise raise it.
+    caPluginUpdateCheck("webgui-pr-PR_PLACEHOLDER.plg", {noDismiss: true}, function(result){
       try {
         let json = JSON.parse(result);
-        if ( ! json.version ) {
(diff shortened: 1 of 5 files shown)


elibosley:
**CodeRabbit fix-loop resolution:**

- **`notify` early `break` blocks keyed re-raise when an archive twin exists (`~270`)** — already addressed. The archive-exists check no longer short-circuits keyed notifications:

  ```php
  if (file_exists($archive)) {
    if ($key !== '') @unlink($archive); else break;
  }
  ```

  For a keyed (condition-style) notification it clears the stale archived copy and **proceeds** to write the fresh unread entry; only non-keyed legacy notifications `break` for dedup. So a keyed condition can re-raise even when a previous archived copy exists.

PR #2684, opened 2026-06-29
update.php fixes

elibosley:
Restrict `update.php` `#file` writes to the expected configuration roots (`/boot/config`, `/etc/wireguard`) so a write cannot land outside config storage via an absolute path or `..`.

- Resolve the target with `realpath()` (collapsing `..`/symlinks) and range-check it with a shared `in_safe_path()` helper.
- Promote `in_safe_path()` into `Wrappers.php` so `update.php` and `FileUpload.php` share one implementation (drops the duplicate copy in `FileUpload.php`).
- Add a regression test that drives the real `update.php` and asserts out-of-root targets are refused while legitimate `/boot/config` writes still succeed.

Relative `#file` paths are unaffected (still placed under `/boot/config/plugins/`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for command execution requests so only paths inside the allowed document root are accepted.
  * Requests with invalid or out-of-scope command paths are now rejected and logged, reducing the risk of unintended file access.

### DIFF
--- a/emhttp/update.php
+++ b/emhttp/update.php
@@ -114,7 +114,7 @@ if (isset($_POST['#command'])) {
   if (strpos($command, $docroot) !== 0)
     $command = "$docroot/$command";
   $command = realpath($command);
-  if ($command === false)
+  if ($command === false || strpos($command, rtrim(realpath($docroot), '/') . '/') !== 0)
     syslog(LOG_INFO, "Invalid #command: {$_POST['#command']}");
   else {
     $command = escapeshellcmd($command);


elibosley:
### Verifying this change (not committed)

One-line hardening of `update.php`'s `#command` handler: the `realpath`-resolved command must stay inside the document root before it runs, so a `../`-escape can't execute a binary outside the webgui.

Offline check from the repo root — should print `blocked`:

```bash
php -d short_open_tag=On -r '
  error_reporting(0);
  $_SERVER["DOCUMENT_ROOT"] = getcwd()."/emhttp";
  $_POST = ["#command" => "../../../../../../bin/echo", "#arg" => ["PWNED"]];
  chdir("emhttp"); include "update.php";
' | grep -q PWNED && echo "NOT contained" || echo "blocked"
```

On a running server: an authenticated POST with `#command=../../../../../../bin/touch` + `#arg[]=/tmp/x` must **not** create `/tmp/x`, and syslog logs `Invalid #command: ...`. Legit `#command` scripts under `/webGui/scripts` and `/plugins/*/scripts` still run.

PR #2686, opened 2026-07-03
fix(shares): allow setting Emptying disk(s) while array is Started

elibosley:
## Summary

The **Global Share Settings** page locked every control whenever the array was not `Stopped`, including **Emptying disk(s)** (`shareUserEmptying`) — even though `emhttpd` accepts the emptying command at runtime and does not require the array to be stopped.

This **splits "Emptying disk(s)" into its own form** at the bottom of the page (mirroring the two-form pattern already used in `NFS.page`), so it can be changed independently while the array is Started. The main settings form remains fully locked while the array is not Stopped.

## Why a separate form (vs. one form)

The page is a single markdown definition-list form and "Emptying disk(s)" sat in the middle of the stop-required settings. A dedicated form:

- posts **only** `shareUserEmptying` to `/update.htm` — no reliance on "disabled fields aren't submitted"
- has its own Apply button and explicit, self-documenting submit semantics
- stays editable while the array is Started, gated only on user shares being enabled

## Changes

- **`emhttp/plugins/dynamix/ShareSettings.page`**
  - New `share_emptying` form at the bottom with the "Emptying disk(s)" control + its own Apply (posts to `/update.htm` with `changeShare`, identical backend path to the existing form).
  - `prepareEmptying()` collapses the multi-select for submit; emptying collapse removed from `prepareShare()`.
  - `presetShare()` gates the emptying control only on user shares being enabled (not array state).
  - Main form footer note reverted to the original "Array must be **Stopped** to change".
- **`emhttp/languages/en_US/helptext.txt`**
  - Added the previously missing `:shares_emptying_disks_help:` entry.

## Testing

- Array **Started**: main settings locked; "Emptying disk(s)" editable; selecting a disk enables its Apply; applying writes only `shareUserEmptying`.
- Array **Stopped**: all settings editable, unchanged behavior.
- User shares = **No**: "Emptying disk(s)" disabled (nothing to empty).

## Linear

Resolves OS-495 — https://linear.app/lime-technology/issue/OS-495

### DIFF
--- a/emhttp/languages/en_US/helptext.txt
+++ b/emhttp/languages/en_US/helptext.txt
@@ -1553,6 +1553,13 @@ Uncheck all disks in order to not exclude any disks
 disks which represent a subset of the Included/Excluded disks defined here.
 :end
 
+:shares_emptying_disks_help:
+This setting designates array disk(s) that are being *emptied* of User Share data,
+typically as a prelude to removing the disk(s) from the array.
+
+Unlike the other Global Share Settings, this setting may be changed while the array is **Started**.
+:end
+
 :shares_exclusive_shares_help:
 If set to Yes, share directories under /mnt/user are actually symlinks to the share directory on a storage volume
 provided the following conditions are met:
--- a/emhttp/plugins/dynamix/ShareSettings.page
+++ b/emhttp/plugins/dynamix/ShareSettings.page
@@ -18,6 +18,9 @@ Tag="share-alt"
 ?>
 <?
 $disabled = _var($var,'fsState')!='Stopped' ? 'disabled' : '';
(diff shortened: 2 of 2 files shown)


elibosley:
Closing — we won't be doing this in 7.4.0

Per discussion with Tom: this is a larger change than it appears on the surface. Unlocking "Emptying disk(s)" at the UI level isn't sufficient — FUSE / the user file system needs to be made aware of the emptying change at runtime as well, otherwise the setting doesn't take effect safely while the array is Started. That backend awareness is the real scope here and is out of scope for 7.4.0

Leaving the branch in place for reference if we pick this up later.

PR #2698, opened 2026-07-16
fix(downgrade): block during protected array shrink

elibosley:
## Summary

An OS downgrade can strand a parity-protected array shrink on a release that cannot recover it. This adds a shared OS-level interlock plus a retained downgrade marker and blocks downgrade staging while protected recovery is active, incomplete, or uncertain.

## Resolution

- Acquire the same exclusive `/var/run/unraid-os-storage-interlock` file lock used by Core before either side checks or publishes its intent/marker.
- Create `/var/run/unraid-os-downgrade` exclusively while holding that lock, then inspect the two Core-owned production records:
  - `parity-protected-shrink.json`
  - `parity-protected-shrink.recovery.json`
- Run checked `/bin/sync -f /boot/config` (`syncfs(2)`) while still holding the lock, clear PHP stat caches, and reread both records before deciding.
- Permit downgrade only when both paths are real regular files below the trusted `/boot/config` directory chain and contain strict, identity-equivalent version-7 records whose literal stage is `completed`.
- Fail closed on missing siblings, I/O/stat ambiguity, non-regular files, final or ancestor symlinks, malformed JSON, active or legacy records, torn writes, and identity disagreement.
- Remove the transient marker and return HTTP 409/503 when protection remains active or durability cannot be proven; do not move boot artifacts.
- Retain the marker after successful downgrade staging so Core rejects new protected storage work until reboot.

## Reviewer considerations

- The shared `flock` makes downgrade-marker publication and protected-shrink publication one cross-process decision.
- The checked sync plus post-barrier reread closes the case where Core's v7 rename is visible but its own durability barrier returned an error before releasing the lock.
- Recursive directory enumeration distinguishes proven absence from an unclassifiable `lstat` failure without following symlinks.
- Rollout order is this WebGUI guard before [Core PR 290](https://github.com/unraid/core/pull/290). This PR does not add an unraidd dependency.

## Verification

- Current PR head: `d74e79c44c28c3e3c5b7648faf365a4d750e7ecd`.
- Destructive VM acceptance head: `31cb4ce25b08919b53ea89b7325967a3eba89935`; the current-head delta
  extracts the existing rejection boundary for direct testing and makes unknown
  decisions fail closed before boot mutation.
- PHP syntax checks pass for the guard, downgrade endpoint, and both tests.
- `parity_protected_shrink_guard_test.php` and
  `downgrade_endpoint_decision_test.php` pass.
- Route coverage proves exact 503/409 mappings, unknown-state fail-closed behavior,
  rejection-side boot-mutation suppression, and ok-only continuation.
- Guard coverage includes no state, active pairs, recovery-only state, matching v7 completion, post-barrier reread, sync failure, identity mismatch, numeric-looking strict identity, torn JSON, legacy additive completion, directory paths, direct and ancestor symlinks, injected stat ambiguity, marker cleanup, and lock handoff.
- Core flow 460 passed in 8.7 minutes and the full three-checkpoint flow-465 hard-power matrix passed in 44.6 minutes against the destructive acceptance WebGUI head. The accepted environment was then restored and independently verified clean.

## Risk

Low code volume but safety-critical ordering. The shared lock, checked durability barrier, strict two-copy validation, retained downgrade marker, and fail-closed path classification keep rollback decisions conservative.

## Summary by CodeRabbit

* **New Features**
  * Added parity-protected downgrade fencing with pre-checks for active/completed/recovery proof identity consistency before proceeding.
  * Downgrades are now blocked with translated, user-facing HTTP status/error guidance when safety/durability synchronization fails or proofs are missing/diverge.
  * Added coordinated downgrade marker handling and an exclusive interlock to prevent unsafe concurrent actions.
* **Tests**
  * Added end-to-end and decision-mapping PHP tests covering active/completed/recovery scenarios, invalid/torn data, durability barrier handling, and interlock locking behavior.


elibosley:
Closing as obsolete: the new Core storage surface replaces the legacy WebGUI path, so this PHP downgrade guard is not part of the rollout and must not be shipped. Core owns storage and OS replacement coordination directly.

PR #2705, opened 2026-07-27
fix(mover): use a widely supported arrow for the Mover direction label

### LINKED ISSUE #2545: Mover Action Direction in Share Settings: illegible character to indicate the direction of Mover's action
https://github.com/unraid/webgui/blob/8b8d1d42444107646b4489b22abf8cabef8413a8/emhttp/plugins/dynamix/ShareEdit.page#L753
https://github.com/unraid/webgui/blob/8b8d1d42444107646b4489b22abf8cabef8413a8/emhttp/plugins/dynamix/ShareEdit.page#L754

Hi All,

 The refenced lines above defines the **select** element's **option** tags used to select the direction of Mover's action on the **Share settings** page.
 The character '⮕' in the lines is supposed to indicate the direction of Mover's action. It doesn't really do that :-)
Maybe it would be better to use HTML entity/codes, like `&#8592;` (&#8592;) or `&#8594;` (&#8594;)? These are all supported by modern browsers, I guess.

junkerderprovinz:
Fixes #2545.

The Mover direction options in Share Settings use `⮕` (U+2B95), which is missing from many system fonts and renders as a tofu box. Replacing it with `→` (U+2192), which is universally available, keeps the same meaning and renders everywhere.

The label is set via jQuery `.text()`, so a literal character is used rather than an HTML entity.

## Summary by CodeRabbit

* **Style**
  * Updated the “Mover action” direction indicator to use a simpler arrow symbol for improved visual consistency.

### DIFF
--- a/emhttp/plugins/dynamix/ShareEdit.page
+++ b/emhttp/plugins/dynamix/ShareEdit.page
@@ -750,8 +750,8 @@ function updateScreen(cache, slow) {
 			}
 
 			if ((secondaryValue !== "0") && (secondaryValue !== primaryValue)) {
-				$('#moverDirection1 option:eq(0)').text($('#primary option:eq(' + z(0) + ')').text() + ' ⮕ ' + $('#secondary option:eq(' + z(1) + ')').text());
-				$('#moverDirection1 option:eq(1)').text($('#secondary option:eq(' + z(1) + ')').text() + ' ⮕ ' + $('#primary option:eq(' + z(0) + ')').text());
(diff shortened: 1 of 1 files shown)


junkerderprovinz:
Small ping on this one. It swaps U+2B95 for U+2192 in the Mover direction labels in ShareEdit.page, which is the tofu box in #2545, open since February. Six lines in one file and CodeRabbit found nothing; happy to rebase, or close it if you'd rather fix this on the font side.

PR #2707, opened 2026-07-27
fix(docker): do not delete the shared fallback icon when removing an icon-less container

### LINKED ISSUE #2691: Docker manager deletes its shared fallback icon (question.png) when an icon-less container is removed/updated
### Environment
Unraid 7.3.2. The relevant code is unchanged on current `master` (line refs below are master).

### Summary
Containers whose template has no `Icon` get the **shared fallback path** recorded as their personal icon, and `removeContainer()` later unlinks that recorded path — deleting `question.png` from the docroot for the whole system. The file only comes back on reboot (bzroot re-extract).

### Chain
1. `include/DockerClient.php:346` — `if (empty($tmp['icon'])) $tmp['icon'] = '/plugins/dynamix.docker.manager/images/question.png';` → persisted per-container into the webui-info registry (docker.json).
2. `include/DockerClient.php:894` (`removeContainer()`) — `$iconRAM = $docroot.$info[$name]['icon']; if ($cache>=1 && is_file($iconRAM)) unlink($iconRAM);` — for an icon-less container, `$info[$name]['icon']` *is* the shared fallback, so the static file `/usr/local/emhttp/plugins/dynamix.docker.manager/images/question.png` gets unlinked.
3. Anything that removes/recreates such a container with cache flags triggers it. Observed in the wild via CA Docker Auto-Update updating an icon-less container: syslog `Installing Updates ...` 22:03:08–22:06:30, images dir mtime 22:06:02, `question.png` gone.

### Impact / aggravator
Once the file is missing, `include/DockerContainers.php:99-100` renders icon-less containers with `src` = question.png **and** `onerror=this.src='(the same path)'`. Since the fallback itself now 404s, the unguarded `onerror` reassigns the same src and the browser re-requests in a tight loop — measured **~737 requests in 5 seconds** from a single open Docker page with 15 such ``s (≈150 req/s against nginx, plus error-log churn on the `/var/log` tmpfs). Third-party plugins that reuse the same fallback path in their own `onerror` handlers (e.g. folder-view style plugins) inherit the storm.

### Reproduction
1. Create a container with no `Icon` template value — its docker.json entry records `icon: /plugins/dynamix.docker.manager/images/question.png`.
2. Update or remove it through a path that passes cache flags (CA Docker Auto-Update does).
3. `ls /usr/local/emhttp/plugins/dynamix.docker.manager/images/` → `question.png` is gone; the Docker page now 404-loops for every icon-less container until the next reboot.

### Suggested fix
In `removeContainer()`, only unlink per-container cached icons (`<name>-icon.png` under images-ram/images) and never the shared default — e.g. skip the unlink when the recorded icon equals the question.png constant. Independently, guarding the JS fallback in `DockerContainers.php` (`onerror="this.onerror=null;this.src=..."`) would eliminate the 404-loop failure class entirely.

junkerderprovinz:
Fixes #2691.

`removeContainer()` unlinks `$docroot.$info[$name]['icon']`. For a container whose template has no icon, that recorded path is the shared fallback `/plugins/dynamix.docker.manager/images/question.png` (set when the icon is empty), so removing or updating such a container deletes `question.png` for the whole system until the next reboot re-extracts bzroot.

Per-container cached icons live under `images-ram` (`$dockerManPaths['images-ram']`); the shared fallback does not. This guards the cleanup so it only unlinks icons inside the `images-ram` cache and never the shared static file. Containers with a real icon are cleaned up exactly as before.

### DIFF
--- a/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
+++ b/emhttp/plugins/dynamix.docker.manager/include/DockerClient.php
@@ -900,9 +900,14 @@ class DockerClient {
 		if (isset($info[$name])) {
 			if (isset($info[$name]['icon'])) {
 				$iconRAM = $docroot.$info[$name]['icon'];
-				$iconUSB = str_replace($dockerManPaths['images-ram'], $dockerManPaths['images'], $iconRAM);
-				if ($cache>=1 && is_file($iconRAM)) unlink($iconRAM);
-				if ($cache==2 && $code===true && is_file($iconUSB)) unlink($iconUSB);
+				// Only remove per-container cached icons (they live under images-ram).
+				// Icon-less containers record the shared fallback (e.g. question.png) as
+				// their icon; unlinking that would delete it system-wide until reboot.
+				if (strpos($iconRAM, $dockerManPaths['images-ram']) === 0) {
+					$iconUSB = str_replace($dockerManPaths['images-ram'], $dockerManPaths['images'], $iconRAM);
+					if ($cache>=1 && is_file($iconRAM)) unlink($iconRAM);
+					if ($cache==2 && $code===true && is_file($iconUSB)) unlink($iconUSB);
+				}
 			}
 			unset($info[$name]);
 			DockerUtil::saveJSON($dockerManPaths['webui-info'], $info);


junkerderprovinz:
Closing this, #2693 already fixed the underlying bug and #2691 is closed. Since I'm here anyway, one thing worth knowing for later: that fix checks equality against the one literal path /plugins/dynamix.docker.manager/images/question.png, this PR instead checked whether the icon path falls under the shared images-ram folder at all (strpos prefix match), which would also cover any future shared fallback icon, not just this one filename. Not worth a new PR for on its own, just flagging it in case it's ever useful.

PR #2720, opened 2026-08-11
fix(os): guard incompatible Unraid installations

SimonFair:
## What changed

The Unraid release installer now checks whether the selected target version is compatible with the current license and boot-device layout.

- Block targets below `7.3.0-beta.0.1` when the license uses a TPM GUID.
- Warn, but continue, when `bootEligible` is nonempty and is not exactly `yes`.
- Block targets at or above `7.3.0-beta.0.1` when `/boot` is mounted from a whole device without a partition table.
- Preserve the existing ReiserFS blocker and its precedence.
- Add a Bash 3.2-compatible integration harness covering 22 compatibility cases.

## Why

Installing an incompatible release can leave a server unable to use its license or boot device. The checks run before release artifacts are staged or written to `/boot`.

## Validation

- `/bin/bash emhttp/plugins/unRAIDServer/tests/unRAIDServer_preflight_test.sh`
- `/bin/bash -n emhttp/plugins/unRAIDServer/tests/unRAIDServer_preflight_test.sh`
- `shellcheck emhttp/plugins/unRAIDServer/tests/unRAIDServer_preflight_test.sh`
- Extracted embedded preflight: `bash -n` and ShellCheck
- `git diff --check`

Related to OS-618.

The direct legacy `Downgrade.php` restore path is tracked separately in OS-647.

### DIFF
--- a/emhttp/plugins/unRAIDServer/tests/unRAIDServer_preflight_test.sh
+++ b/emhttp/plugins/unRAIDServer/tests/unRAIDServer_preflight_test.sh
@@ -1,0 +1,309 @@
+#!/bin/bash
+
+set -u
+
+TEST_DIR=$(cd "$(dirname "$0")" && pwd)
+PLUGIN_FILE="$TEST_DIR/../unRAIDServer.plg"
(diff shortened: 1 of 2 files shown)


SimonFair:
Closing this draft because OS-618 will be reviewed through the master PR only.

PR #2722, opened 2026-08-11
Theme-aware plugin icons via an optional -dark or -light variant of the icon file

### LINKED ISSUE #2704: Feature request: theme-aware plugin icons (light/dark logo that switches with the theme, like glyph icons)
### Problem

A plugin's `.page` `Icon` can be a font glyph or a static image. Glyph icons adapt to the active theme automatically: they render as `<b class="fa fa-... system">` and inherit the theme text color. Image icons do not. A `.png` renders as a static ``, so a full-color logo can look wrong on some themes (a dark logo disappears on the black theme, a light logo washes out on the white theme).

Today the only theme-adaptive option for an image is a single-color inline SVG using `currentColor`, which forces the logo to be monochrome and loses its real colors.

### Request

Let a plugin provide a light and a dark logo variant that the webgui switches based on the active theme, the same way glyph icons already adapt. This lets plugin authors ship proper full-color logos that look right on every theme.

### Where it stands today

The icon is rendered per type, for example in `plugins/dynamix/include/DefaultPageLayout/Navigation/Main.php`:

```php
$icon = $button['Icon'];
if (substr($icon, -4) == '.png') {
  $icon = "<img src='/{$button['root']}/icons/$icon' class='system'>";   // static image
} elseif (substr($icon, 0, 5) == 'icon-') {
  $icon = "<b class='$icon system'></b>";                                // Unraid glyph, theme-aware
} else {
  $icon = "<b class='fa fa-$icon system'></b>";                          // FontAwesome, theme-aware
}
```

The same `.png -> ` pattern also appears in the dashboard tiles, the sidebar-theme icon CSS, the settings/tools tiles, and the plugin page header.

### Proposed mechanism (backward compatible)

Two options, whichever you prefer:

- A) Naming convention: alongside `icon.png`, an optional `icon-dark.png`. If the dark variant exists, render both and let CSS pick the right one per theme. No manifest change.
- B) A new optional `.page` field, for example `IconDark=icon-dark.png`, for an explicit second variant.

In both cases: dark themes show the light logo, light themes show the dark logo. No variant present means the single image renders exactly as today, and glyph icons are unchanged.

The switch can reuse the existing theme classes, for example:

```css
.icon-theme-dark { display: none; }
.Theme--black .icon-theme-light { display: none; }
.Theme--black .icon-theme-dark  { display: inline; }
```

### Open questions for maintainers

- Preferred mechanism: naming convention (A) or a new field (B)?
- Which themes count as dark (black, azure)?
- Should this also cover SVG icons and the sidebar-theme icon path?

Happy to open a PR once the mechanism and the dark-theme set are confirmed.

### Environment

- Unraid 7.x webgui (dynamix)

junkerderprovinz:
A plugin can now ship a per-theme variant of its `.png` icon. Next to `icon.png` it may place `icon-dark.png` and/or `icon-light.png`, and the webgui serves the variant matching the active theme.

Plugins that ship a single icon are untouched: with no matching variant file present the icon renders exactly as it does today, and glyph icons (`icon-*` and FontAwesome) are not affected at all.

Closes #2704.

### Why

Glyph icons already follow the theme, because they render as `<b class="fa fa-... system">` and inherit the theme text color. Image icons do not. A `.png` renders as a static ``, so a full-color logo can be unreadable on some themes: a dark logo disappears on the black theme, a light one washes out on the white theme.

The only theme-adaptive option for an image icon today is a single-color inline SVG using `currentColor`, which forces the logo to be monochrome and loses its real colors. In practice plugin authors work around this in the artwork instead, for example by drawing concentric dark and light rings so that one of them always contrasts with whatever background the icon lands on. That constrains the logo rather than fixing the problem.

### How

One helper in `Helpers.php`:

```php
function theme_icon($icon)
```

It takes the docroot-relative path of an already resolved icon and returns either the theme variant or the same path unchanged. It returns early for anything that is not a `.png`, so glyph icons never reach the file check.

Which themes count as dark is not a new decision. The helper asks the existing `ThemeHelper::isDarkTheme()`, so `black` and `gray` get `-dark`, `white` and `azure` get `-light`, and any theme added later is classified in one place. The suffix is resolved once per request and cached in a static, so a page with many icons costs at most one extra `file_exists` per icon.

The theme comes from `$display['theme']` on a normal page load. `ShowPlugins.php` is fetched standalone over AJAX and has no `$display`, so the helper falls back to `parse_plugin_cfg('dynamix',true)` there. That keeps the helper self-contained instead of changing the variable scope of that endpoint.

The helper is applied at the four places that render a plugin-supplied `.png` icon:

- the nav and sidebar utility buttons (`.page` `Icon=`, from `<root>/icons/`)
- the Settings and Tools tiles (`.page` `Icon=`, from `<root>/images/` or `<root>/`)
- the page and tab title (from `<root>/icons/`)
- the Plugins page list (`.plg` icon, from `plugins/<name>/images/` or `plugins/<name>/`)

In each case the helper runs after the existing lookup has decided which file to use, so the existing resolution order and the existing fallbacks are unchanged.

Deliberately not covered: Docker container and VM icons, which come from user-editable templates and remote URLs rather than from files a plugin ships. Also SVG, because a `.svg` value in `Icon=` is not handled as an image today (it falls through to the FontAwesome branch), so supporting it would be a separate change.

This implements option A from the issue (naming convention, no manifest change), because it needs no new field in either the `.page` or the `.plg` format and works the same for both.

### Testing

`php -l` is clean on all changed files under PHP 8.4.23 with `short_open_tag=On`.

`theme_icon()` was also exercised directly against the real `ThemeHelper` using a sandbox docroot:

- `black` and `gray` pick `icon-dark.png` when it exists
- `white` and `azure` pick `icon-light.png` when it exists
- an icon with no variant, an icon whose file is missing, and the glyph values `rocket` and `icon-app` all come back unchanged
- with `$display` unset, the config fallback resolves the theme correctly and raises nothing at `E_ALL`

### Files

- `emhttp/plugins/dynamix/include/Helpers.php`
- `emhttp/plugins/dynamix/include/PageBuilder.php`
- `emhttp/plugins/dynamix/include/DefaultPageLayout/MainContent.php`
- `emhttp/plugins/dynamix/include/DefaultPageLayout/Navigation/Main.php`
- `emhttp/plugins/dynamix.plugin.manager/include/ShowPlugins.php`

## Summary by CodeRabbit

- **New Features**
  - Added theme-aware icons across plugins, navigation, pages, tabs, and virtual machine displays.
  - Icons automatically use light or dark variants based on the active theme when available.
  - Standard icons remain available as a fallback when no themed version exists.
  - Virtual machine icon selectors now hide theme-specific variant files to simplify selection.
  - Consistent icon styling is now maintained throughout the interface.


junkerderprovinz:
CodeRabbit's `/boot/config/...` finding is a false positive, so I have left the code as is.

`emhttp/boot` is a committed symlink to `/boot`, which makes `$docroot/boot/config/...` the correct filesystem path for a `/boot/config/...` URL. That is already the convention in this same code: `libvirt.php` line 1461 tests `is_file("$docroot/boot/config/plugins/dynamix.vm.manager/templates/images/...")` before returning the `/boot/config/...` URL, and the icon chooser in `VMedit.php` globs `$docroot/boot/config/...` while emitting `/boot/config/...` as the URL base. `theme_icon()` follows the same convention, so user-installed VM icons on the flash drive do resolve their `-dark` and `-light` variants.

The only branch of `domain_get_icon_url()` that can return a path outside the docroot is the raw `is_file($strIcon)` case. Such a path is not servable as an `` today either, and `theme_icon()` returns it unchanged, so nothing regresses there.

PR #2754, opened 2026-09-16
fix(rc.S): reinstate a boot mirror member that drops after import (DST-252)

elibosley:
> **Boot code — please review carefully.** A mistake here makes systems unbootable. The behaviour change is bounded and additive, and the common path is untouched, but it deserves a second pair of eyes.

## What happens today

A mirrored ZFS boot pool can import healthy and then lose a member moments later. Captured on 7.4.0-beta.1 while a mirror member rejoined after being absent for one boot:

```
+0.092s  pool_import                       both members
+0.244s  resilver_start
+0.537s  resilver_finish                   errors=0
+0.614s  ereport.fs.zfs.vdev.unknown       /dev/sdc3
+0.614s  statechange                       UNAVAIL
```

The member resilvers cleanly, then ~77ms later ZFS closes the vdev, fails to reopen it, and parks it as UNAVAIL. The evidence says the device is fine:

- all four ZFS labels present, correct pool GUID, correct topology, 4 txg behind
- `vdev_read_errors`, `vdev_write_errors`, `vdev_cksum_errors`, `vdev_delays` all **0**
- `prev_state = CLOSED` — an **open** failure, not a media failure
- partition table, ESP hash and data hash byte-identical before and after
- a full scrub immediately afterwards also reports `errors=0`

ZFS's status text, *"label is missing or invalid"*, is `VDEV_AUX_BAD_LABEL` — what a **failed reopen** sets. It is not a statement about the on-disk label, which is intact.

## Why nothing recovers it

`rc.S` checks health once, sees a healthy pool, and returns — **before** the member drops — so the degraded branch below never runs. And this system starts no ZED (no matches in the boot log, nothing in `etc/rc.d/`), so no daemon notices afterwards.

The pool stays degraded until somebody looks. For a boot mirror that means its redundancy is silently gone — and the boot mirror is the one pool whose redundancy is supposed to survive a disk being pulled and put back.

For contrast, Unraid OS 8 handles the same rejoin correctly (`zfs-import-scan.service` with the cache masked): member returns, resilvers, pool goes back to `ONLINE`.

## The change

Settle briefly, then ask ZFS to bring each member back. `zpool online` on a member that is already ONLINE does nothing, so this needs no state parsing — which matters, because `awk` and `sed` are not available this early in boot.

| case | behaviour |
| --- | --- |
| healthy pool (incl. every single-disk flash) | returns on the first check — **no sleep, no online, no log line, no added boot time** |
| degraded, recovers | ~3s settle, online each member, reports ONLINE |
| degraded, unrecoverable | ~13s max, then boots on the surviving member exactly as before |

Verified against a stubbed `zpool` for all three paths. `bash -n` clean; `shellcheck -S error` reports only the pre-existing SC2145 on line 13.

## What this does not do

**It does not fix the reopen failure**, and I could not determine the cause from the artifacts. It fixes the separate, clearer problem: that a valid member, once dropped, is never reinstated. If the underlying reopen bug is later fixed, this becomes a harmless no-op.

## Risk

The one real behaviour change is that a member ZFS faulted for a genuine reason now gets one bounded `online` attempt. ZFS re-faults it if it is truly bad, and the current alternative is booting degraded indefinitely — but it is worth a reviewer's judgement.

Not yet exercised on real hardware or in the migration suite; the failing case that motivated it is `QA VM · mirror-update` in unraid/os, which has never passed since it was written on 2026-09-10.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boot-pool recovery by restoring eligible offline pool members after import.
  * Added stability checks requiring three consecutive healthy readings within six seconds before confirming recovery.
  * Retained bounded recovery polling when stability cannot be confirmed.
  * Boot pools that remain degraded now report their final health status after recovery attempts.
  * Single-device pools are not subject to member-reinstatement checks.


elibosley:
Linear: **OS-884** — *Mirrored ZFS boot pool silently loses a member after it rejoins* (project `Unraid OS (7.4.0)`).

The ticket carries the full evidence and, importantly, records the two problems separately: this PR fixes **(1) nothing reinstates a valid dropped member**, and leaves **(2) why the reopen fails** open and unresolved.

elibosley:
Correction to the reference above: the issue moved to the **Distro** team, since this is a Slackware-level change. It is now **DST-252** (was OS-884 — same issue, Linear reassigns the identifier on a team move).

https://linear.app/lime-technology/issue/DST-252/mirrored-zfs-boot-pool-silently-loses-a-member-after-it-rejoins

Project `Unraid OS (7.4.0)` is unchanged.

elibosley:
## Fixed in 46c4f75 — the finding was right, and the bug was worse than described

The settle loop returned on the **first** ONLINE sample. That repeats the exact mistake this function exists to catch: the failure is a pool that reads healthy and loses a member ~77ms later, and this runs immediately after the import check — **well inside that window**.

So in the real failing case it would have sampled ONLINE, returned, and reinstated nothing. **The patch would not have fixed its own motivating scenario.**

### Change

Require the pool to hold ONLINE across consecutive samples, resetting the counter on any non-ONLINE sample. The recovery loops for a pool that stays degraded are unchanged.

A single-device pool now returns *before* sampling at all — it has no mirror member that can drop, so there is nothing to wait for. That keeps the ordinary single flash boot exactly as fast as it was and confines the stability wait to mirrored pools, where it is the point.

### Verified against a stubbed `zpool`

| health sequence | old | new |
| --- | --- | --- |
| ONLINE → DEGRADED… (the real failure) | **returned silently, reinstated nothing** | reinstates, recovers to ONLINE |
| ONLINE throughout | returns | returns, no `online` calls |
| single-disk pool | returns | returns immediately, zero samples |
| DEGRADED throughout | attempts recovery | attempts recovery, reports DEGRADED |

The old-vs-new comparison on row 1 was run explicitly rather than assumed: the previous version produced no output at all.

`bash -n` clean; `shellcheck -S error` still reports only the pre-existing SC2145 on line 13.

`coderabbit` CLI is not installed here, so I could not run the local agent review.

elibosley:
Closing without merging. The branch is kept.

The failure this targeted did not reproduce. On unraid/os run 35051846466 all four mirror rejoin phases came back ONLINE, including both `source-rejoined-*` phases, which had never passed before. So the UNAVAIL failure is intermittent, and the probe from unraid/os#138 never fired, which leaves this patch unvalidated.

There is also counter-evidence against it. In the original failing boot, rc.S already ran `zpool online -e` on both members, and ZFS answered "onlined, but remains in faulted state". This patch runs a plain `zpool online` and waits, which is a different operation, but the only data we have suggests that onlining the member did not bring it back.

Because the failure is intermittent, a fix is hard to confirm and easy to credit falsely. If this were merged and `mirror-update` then passed, we could not tell whether the patch helped or whether the race simply did not occur.

The gap it describes is still real: rc.S checks health once and returns, no ZED runs, and Unraid OS 8 handles the same rejoin correctly. That stays tracked in DST-252. unraid/os#138 is merged, so the probe runs automatically on the next recurrence. That verdict is the evidence that should decide whether to reopen this.

